PR #3, opened 2024-04-19
Changed repo to mgs-workflow in the definition of three paths in index.config to be consistent with the name the repo has.

simonleandergrimm:
My index run choked on these file names. If there is a cleaner fix let me know. I'd assume it would be:
```
repo_dir = "mgs-workflow"
[...]

contaminants = "${projectDir}/../${repo_dir}/ref/contaminants.fasta.gz" // FASTA file containing additional "contaminant" sequences to screen out during viral identification.
```

### DIFF
--- a/configs/index.config
+++ b/configs/index.config
@@ -23,12 +23,12 @@ params {
     taxonomy_url = "https://ftp.ncbi.nlm.nih.gov/pub/taxonomy/taxdump_archive/taxdmp_2022-12-01.zip"
 
     // Other reference files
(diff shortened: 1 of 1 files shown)


willbradshaw:
"The name the repo has" is purely up to the person cloning the repo, this isn't something we can control within it. A `repo_dir` argument would mean this only has to be set once within the config file but wouldn't fully resolve this.

In general, the reason things are in the config file is because it's the user's responsibility to specify them.

jeffkaufman:
> "The name the repo has" is purely up to the person cloning the repo

Why not just say "you need to clone this repo as `mgs-workflow`"?  This is what most folks will do by default.

PR #41, opened 2024-07-12
Slight doc additions to early cleaning steps

evanfields:
Suggesting a few notes I would have found helpful when reading the pipeline's early cleaning steps. I'm not strongly attached to any of them, so I welcome feedback on what kinds of inline documentation we want in the pipeline. (i.e. the right inline comments might not be the same as the set of comments a new reader would want to see.)


test/nextflow.config
```diff
@@ -8,7 +8,7 @@ params {
     // Directories
     base_dir = "${launchDir}" // Parent for working and output directories (can be S3)
     raw_dir = "${launchDir}/raw" // Raw data directory
-    ref_dir = "s3://nao-mgs-wb/index-20240618/output/results" // Reference/index directory (generated by index workflow)
+    ref_dir = "s3://nao-mgs-wb/index-20240618/output" // Reference/index directory (generated by index workflow)
```
evanfields, line 11: This is the only non-comment/readme change in the PR. I had to make this change to get the test data to run with the existing index.

willbradshaw:
This all looks fine, but can you merge to `dev` instead of `master`? I'd like to control updates to master a bit more carefully going forward.

jeffkaufman:
> can you merge to dev instead of master? I'd like to control updates to master a bit more carefully going forward.

When you get a chance, could you document the workflow you'd like to see here?

(Also, note that if you put off README changes until the next dev-to-master merge then people will who go to the https://github.com/naobservatory/mgs-workflow will see obsolete docs)

willbradshaw:
> When you get a chance, could you document the workflow you'd like to see here?

My suggested workflow would be:

1. People work on pipeline changes on their own branches
2. When ready, they submit pull requests to merge into `dev`
3. Periodically (could be immediately for important fixes or on some schedule otherwise) I update the changelog and version number and merge into `master`.

> (Also, note that if you put off README changes until the next dev-to-master merge then people will who go to the https://github.com/naobservatory/mgs-workflow will see obsolete docs)

There's one suggested change in this PR that's fixing an error in the documentation; the rest are minor QOL improvements. It seems potentially justifiable to merge the former directly into master, but much less so for the latter.

jeffkaufman:
> 
> My suggested workflow would be: ...

This looks good, and would be good to add to the README

> > (Also, note that if you put off README changes until the next dev-to-master merge then people will who go to the https://github.com/naobservatory/mgs-workflow will see obsolete docs)
> 
> There's one suggested change in this PR that's fixing an error in the documentation; the rest are minor QOL improvements. It seems potentially justifiable to merge the former directly into master, but much less so for the latter.

We should probably say something in the docs about how if you have a documentation change that's describing how to use the code that's already on master, that should be a PR to master (and that PR shouldn't contain anything else)?

evanfields:
@willbradshaw thanks for the look; I changed the base branch.

Will and also @jeffkaufman 
* thanks for eyes and discussion
* by "work on pipeline changes in their own branches", do you mean in this repo or a forked version?
* I'm not sure if I should read Jeff's last comment as instruction to break the `README` change into a separate two-line PR or just brainstorming about potential processes.

jeffkaufman:
> I'm not sure if I should read Jeff's last comment as instruction to break the `README` change into a separate two-line PR or just brainstorming about potential processes.

I don't think you need to change this PR; this was a suggestion (not yet settled) for what the process should be for future PRs

PR #60, opened 2024-09-30
Remove concat gzipped

harmonbhasin:
Removes concat gzipped and migrates the input format to something more similar to that used in nf-core.


willbradshaw:
Thanks @harmonbhasin, this is looking good! Some comments/questions/requests:

1. Could you tell me what testing you've done on this branch? E.g. did you test that the new test samplesheet runs the test workflow correctly?
2. In both the README and the samplesheet generation script, we need to change things so they can handle arbitrary read1/read2 suffixes. (E.g. for the latest Broad data the suffixes were "_L001_R1_001"/"_L001_R2_001", not "_1"/"_2".) In the script it's probably fine for now to just have arguments to configure these manually if they're not _1/_2.
4. Let's use standard filename conventions from nf-core here, so call the samplesheet `samplesheet.csv` by default and refer to it as `params.sample_sheet` or similar in the workflow (rather than `params.library_tab`).
5. Could we give the samplesheet generation script a shorter name?

harmonbhasin:
> Thanks @harmonbhasin, this is looking good! Some comments/questions/requests:
> 
> 1. Could you tell me what testing you've done on this branch? E.g. did you test that the new test samplesheet runs the test workflow correctly?
> 2. In both the README and the samplesheet generation script, we need to change things so they can handle arbitrary read1/read2 suffixes. (E.g. for the latest Broad data the suffixes were "_L001_R1_001"/"_L001_R2_001", not "_1"/"_2".) In the script it's probably fine for now to just have arguments to configure these manually if they're not _1/_2.
> 3. Let's use standard filename conventions from nf-core here, so call the samplesheet `samplesheet.csv` by default and refer to it as `params.sample_sheet` or similar in the workflow (rather than `params.library_tab`).
> 4. Could we give the samplesheet generation script a shorter name?

1. I've tested this both on the test branch and on the Cris-Cristoph reads.
2. Adjusted the script to allow for different suffixes for forward and reverse read and even allow for characters in the middle that can be mapped by regex.
3. Done
4. Done, now it's called generate_samplesheet.sh

willbradshaw:
[Writing up TODOs from in-person conversation with @harmonbhasin]

- `generate_samplesheet.sh` suffix parameters should take regex, then we can eliminate the middle parameter
- `generate_samplesheet.sh` should accept parameters from the command line
- Everything else looks good, once those are implemented I will merge

harmonbhasin:
@willbradshaw done!

willbradshaw (approved):

PR #63, opened 2024-10-04
Amended RAW workflow to take in single read files.

simonleandergrimm:
@harmonbhasin Can you look this over and let me know if we should continue with this style?

Specifically, what I'm doing is:

- Create alternative versions of processes where needed and then name process `{PROCESS}_PAIRED` and `{PROCESS}_SINGLE`
- I keep workflow names the same throughout. Within workflows, the processes are picked based on the run being single-end or paired-end.
- The workflows themselves include if/then clauses (e.g., see what I did in `raw/main.nf`)

The `run.nf` currently only executes RAW. I will progressively work forward through the entire run.nf, so the final `simon-single-read-dev` branch has the full `run.nf`.


harmonbhasin (changes requested):
@simonleandergrimm this looks good to me, good work! I just have a question and a request:

Request: Could you make sure to write clear commit messages for each file. You did this for your later commits, but your first commit only had one commit message for a bunch of files. In general, this makes the review process much easier, and makes analyzing the repo in the future easier as well.

Question: Have you checked that the output in summarize-multiqc-single.R is correct? I remember that changing this script had lead me to some issues, specifically, it would output a file, but some of the columns had NA where there should've been output. Also, I think that summarize-multiqc.R should be one file, you can pass in an argument specifying whether it is single or paired end, then update the code from there.

PR #67, opened 2024-10-09
Deduplication on paired reads using clumpify and BBtools deduplication and fragment length analysis

harmonbhasin:
#60 should be merged first before attempting to merge this branch as it is based on that branch.

Deduplication and fragment length analysis from bbtools has been added to the `hv_hits_putative_collapsed.tsv.gz` table.


jeffkaufman:
> #60 should be merged first before attempting to merge this branch as it is based on that branch.

The way to do this is to set `remove_concat_gzipped` as the merge target for this PR.  Then this will show up as a delta on top of #60

modules/local/bbmerge/main.nf
```diff
@@ -5,6 +5,7 @@ process BBMERGE {
     input:
         tuple val(sample), path(reads)
     output:
+        tuple val(sample), path("${sample}_bbmerge_merged.fastq.gz"), emit: merged
```
willbradshaw, line 8: @harmonbhasin what's the thinking behind outputting the merged read file twice (once as `merged` and once in `reads`)? If you need to access this file specifically in another process, you can just access `reads[1][0]`.

modules/local/collapseHV/main.nf
```diff
@@ -5,6 +5,8 @@ process COLLAPSE_HV {
     memory "16.GB"
     input:
         path(hv_hits_filtered)
+        path(merged_bbmerge_paired_results)
```
willbradshaw, line 8: This module is really only solving one specific problem, namely the presence of discordant read pairs in the HV TSV. I'd suggest moving the new parts to a different module.

modules/local/summarizeBBMerge/main.nf
```diff
@@ -0,0 +1,17 @@
+process SUMMARIZE_BBMERGE {
+    label "base"
+    label "single"
+    input:
+        tuple val(sample), path(merged)
+    output:
+        path("${sample}_bbmerge_merged_summary.txt")
+    script:
+        """
+        zcat ${merged} | awk '
```
willbradshaw, line 10: Can we also get the post-merge sequence? This isn't currently included in the HV table and would be good to have.

(Ditto the post-concat sequence for unmerged read pairs, though this is somewhat easier to derive from the paired sequences.)

willbradshaw: Also, what's your plan for the merge status?

modules/local/summarizeDedup/main.nf
```diff
@@ -0,0 +1,25 @@
+process SUMMARIZE_DEDUP {
```
willbradshaw, line 1: Let's talk this one over in person.

subworkflows/local/taxonomy/main.nf
```diff
@@ -32,15 +35,26 @@ workflow TAXONOMY {
         } else {
             subset_ch = SUBSET_READS_PAIRED(reads_ch, params.read_fraction)
         }
+
+         // Deduplicate reads (if applicable)
```
willbradshaw, line 39: What's the thinking behind making this optional? Do you have a case in mind where we wouldn't want to do this?

willbradshaw:
Independently from my other comments, it looks like this PR is trying to merge changes that are actually made in #60, which suggests something might have gone awry somewhere?

modules/local/bbmerge/main.nf
harmonbhasin: @willbradshaw I ended up fixing this. The solution you suggested works, I just had to adjust the syntax.

modules/local/collapseHV/main.nf
harmonbhasin: @willbradshaw done.

modules/local/summarizeBBMerge/main.nf
harmonbhasin: @willbradshaw chatted about this during out meeting, but this was in collapseHV, which has now been moved to it's own module.

modules/local/summarizeDedup/main.nf
harmonbhasin: @willbradshaw done.

willbradshaw (approved):

PR #70, opened 2024-10-15
Count duplicate reads using alignment from Bowtie2

harmonbhasin:
Implemented a script in rust that counts duplicate reads using the alignment from bowtie2 once it has been converted to a tsv.


jeffkaufman:
Is the base branch set correctly here?  This is showing up as a PR to merge 42 commits, with a ton of modified lines, which is usually a sign that something is set wrong,

willbradshaw:
It looks like this includes changes from #60 and #67.

willbradshaw (approved):

PR #71, opened 2024-10-15
Bug fix in subset reads

harmonbhasin:
Updated subset reads so that it uses all reads if the number of reads is below the subset threshold set in the nextflow.config file. Solution was copied from @willbradshaw.


willbradshaw (approved):

willbradshaw:
@harmonbhasin any change to master should have a CHANGELOG update and minor version increment; can you do this? Then I'll merge.

harmonbhasin:
@willbradshaw done!

PR #79, opened 2024-10-22
Reintroduce concatenating samples

harmonbhasin:
Reintroducing concatenating samples within the same group during hv workflow. I still need to update the README.md and configs/run.config. However, I want to first see if you think this implementation is sufficient. I've tested this on the test dataset with and without groups.


harmonbhasin:
Results can be found here for subsetting in hv workflow and taxonomic profile:
- Results with no subsetting (s3://nao-harmon/test-now/)
- Results with subsetting (s3://nao-harmon/test-subset/)

harmonbhasin:
After Will pointed out the bug, that I just fixed, I reran the pipeline, and results can be found in the same location, they now look fixed.

README.md
```diff
@@ -287,7 +287,7 @@ To run the workflow on another dataset, you need:
...
+> The easiest way to get this file is by using the `generate_samplesheet.sh` script. As input, this script takes a path to raw FASTQ files (`dir_path`), and forward (`forward_suffix`) and reverse (`reverse_suffix`) read suffixes, both of which support regex, and an optional output path (`output_path`). Those using data from s3 should make sure to pass the `s3` parameter. Those who would like to group samples by some metadata should pass a path to a CSV file containing a header column named `sample,group`, where each row gives the sample name and the group to group by (`group_file`). As output, the script generates a CSV file named (`samplesheet.csv` by default), which can be used as input for the pipeline.
```
willbradshaw, line 290: I'd prefer we say something like "can pass a CSV file [...] or edit the samplesheet manually after generation" (since in cases where the groups CSV isn't readily available manual editing will often be easier).

bin/generate_samplesheet.sh
```diff
@@ -43,7 +49,15 @@ done
...
-    echo "Usage: $0 --dir_path <path> --forward_suffix <suffix> --reverse_suffix <suffix> [--s3] [--output_path <path>]"
+    echo -e "\nUsage: $0 [options]"
+    echo -e "\nRequired arguments:"
+    echo -e "  --dir_path <path>         Directory containing FASTQ files"
+    echo -e "  --forward_suffix <suffix>  Suffix identifying forward reads, supports regex (e.g., '_R1_001' or '_1')"
+    echo -e "  --reverse_suffix <suffix>  Suffix identifying reverse reads, supports regex (e.g., '_R2_001' or '_2')"
+    echo -e "\nOptional arguments:"
+    echo -e "  --s3                      Use if files are stored in S3 bucket"
+    echo -e "  --output_path <path>      Output path for samplesheet [default: samplesheet.csv]"
+    echo -e "  --group_file <path>       Path to group file for sample grouping [header column must be named 'sample,group']"
```
willbradshaw, line 60: Can the group file have other columns as well? Or columns in a different order? Or must it be sample,group and nothing else?

willbradshaw (commented):
A couple minor comments about documentation. Nearly there!

README.md
harmonbhasin: Agreed, updated.

bin/generate_samplesheet.sh
harmonbhasin: It can have other columns, but these won't be added to the samplesheet.csv, I'll add this to the README and comments here.

harmonbhasin: Only updated this in script, let me know if you want me to do it in the README as well.

PR #83, opened 2024-11-01
Will debug multiqc

willbradshaw:
Fix bugs arising in QC workflow for inputs with nonstandard read naming.


harmonbhasin (approved):
Looks good to me!

PR #85, opened 2024-11-12
Expanding HV identification workflow to cover arbitrary host taxa

willbradshaw:
I've rewritten the index workflow to generate reference files for viruses that infect arbitrary host taxa, and the run workflow to work with the results. Results for the latest Sabeti data are shown [here](https://data.securebio.org/wills-private-notebook/notebooks/2024-11-11_sabeti_expanded.html); everything seems to work in terms of the outputs, but the increased size of the viral reference DB seems to lead to substantially longer runtimes of the viral-read-identification workflow, which is something we should work on down the line.

I haven't updated the README yet; once this is merged I'll work on updating documentation before merging into master.


willbradshaw:
I realized I hadn't pulled in the latest changes from Harmon's work on dev. This is now done and I'm testing the output; will update this shortly.

harmonbhasin (approved):
Looks good thus far.

PR #94, opened 2024-11-14
Extracting BBDuk-subset raw reads for chimera detection

willbradshaw:
This modification to the pipeline adds an additional intermediate output, containing raw reads corresponding to those cleaned reads that make it through BBDuk screening for host-viral reads.

Specifically, the code takes the output of BBDuk, extracts the read IDs, and then extracts the raw reads with corresponding IDs from the raw FASTQ files.


harmonbhasin (approved):
Looks good to me, the description you provided accurately describes what you did.

jeffkaufman (approved):
I'm not very familiar with Nextflow, but this seems plausible!

PR #103, opened 2024-11-20
generate_samplesheet: add --group_across_illumina_lanes argument

jeffkaufman:
We are often working with Illumina data, and a very common way to run a full flowcell is to prepare a single library that you run across all the lanes.  For example, this is how all the data the NAO is receiving from Marc Johnson's lab looks.  With this data the _Lnnn suffix on each sample tells you what lane the data came from, which isn't meaningful.  Add an option to group across lanes, so samples 'foo_L001' and 'foo_L002' are the same group, 'foo'.

Also add -e and -u so the script will exit if a command fails or tries to read an undefined variable. This is best practice for bash scripts.

Separately, this script is too complex to be in bash, and should be python.


willbradshaw:
This looks fine to me, but @harmonbhasin should review since he wrote the script.

> Separately, this script is too complex to be in bash, and should be python.

I agree. I'll open an issue.

harmonbhasin (approved):
This looks good to me. I agree that this bash script should be converted to python.

willbradshaw:
@jeffkaufman it looks like this now has conflicts with dev that need to be resolved before I can merge.
While you're at it, could you add a line to CHANGELOG about this?

willbradshaw:
Once that's done I'll merge into dev and then merge 5.2 into master.

jeffkaufman:
> it looks like this now has conflicts with dev that need to be resolved before I can merge. While you're at it, could you add a line to CHANGELOG about this?

Done and done!

PR #112, opened 2024-11-26
Grouping

harmonbhasin:
Added grouping functionality, fixed bug with multiqc container, and successfully tested on Crits-Christoph dataset. Once this merges into dev, it should be read to merge into master.

### DIFF
--- a/workflows/run.nf
+++ b/workflows/run.nf
@@ -30,6 +30,16 @@ workflow RUN {
     kraken_db_path = "${params.ref_dir}/results/kraken_db"
     blast_db_path = "${params.ref_dir}/results/${params.blast_db_prefix}"
 
+    // Check if grouping column exists in samplesheet
+    check_grouping = new File(params.sample_sheet).text.readLines()[0].contains('group') ? true : false
+    if (params.grouping != check_grouping) {
+        if (params.grouping && !check_grouping) {
+            throw new Exception("Grouping column not found in samplesheet")
+        } else if (!params.grouping && check_grouping) {
+            throw new Exception("Grouping column found in samplesheet but not requested")
+        }
+    }
+
     // Prepare samplesheet
     if ( params.grouping ) {
         samplesheet = Channel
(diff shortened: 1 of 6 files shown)


workflows/run.nf
```diff
@@ -30,6 +30,16 @@ workflow RUN {
     kraken_db_path = "${params.ref_dir}/results/kraken_db"
     blast_db_path = "${params.ref_dir}/results/${params.blast_db_prefix}"
 
+    // Check if grouping column exists in samplesheet
+    check_grouping = new File(params.sample_sheet).text.readLines()[0].contains('group') ? true : false
+    if (params.grouping != check_grouping) {
+        if (params.grouping && !check_grouping) {
+            throw new Exception("Grouping column not found in samplesheet")
```
willbradshaw, line 37: I think this could be more informative, e.g. "Grouping enabled in config file, but group column absent from samplesheet."

workflows/run.nf
```diff
@@ -30,6 +30,16 @@ workflow RUN {
...
+        } else if (!params.grouping && check_grouping) {
+            throw new Exception("Grouping column found in samplesheet but not requested")
```
willbradshaw, line 39: What do you think about making this a warning? In principle people might genuinely want to run an ungrouped analysis off a grouped samplesheet.

willbradshaw (changes requested):

willbradshaw:
Regarding running the test locally, if we’re expecting folks to do this before they make pull requests it should be included in the documentation somewhere

workflows/run.nf
harmonbhasin: Ah, good point, I'll change that.

workflows/run.nf
harmonbhasin: I think it's better to have an exception as, in my experience, people tend to ignore warnings or just may miss them. The user can make another samplesheet without the grouping column if they wish.

harmonbhasin:
> Regarding running the test locally, if we’re expecting folks to do this before they make pull requests it should be included in the documentation somewhere

This makes sense, I'll add something to the README.

willbradshaw (approved):

PR #113, opened 2024-11-26
Adding infrastructure for S3 auto-cleaning

willbradshaw:
- Added tags to published files to facilitate S3 auto-cleanup
- Added S3 lifecycle configuration file to `ref`, along with a script in `bin` to add it to an S3 bucket


harmonbhasin (approved):
Given the test passes, this looks good to me.

PR #116, opened 2024-11-27
Fixing bug that's causing a fatal error in master pipeline.

harmonbhasin:
This allows the pipeline to run, however I'm now running into other issues with the Crits-Christoph dataset not being able to run, but this is most likely due to files being deleted in the index directory. I expect this test to fail.


harmonbhasin:
Oops, actually this test won't fail because it's using my index (one I made for testing) and not the one we use for running actual datasets

willbradshaw (approved):

PR #118, opened 2024-12-03
Changed version of nextflow, hopefully this works

harmonbhasin:
Simon bought up how github actions fails now, this is most likely due to using "latest-edge", which is [not always stable](https://github.com/marketplace/actions/setup-nextflow)

### DIFF
--- a/.github/workflows/end-to-end.yml
+++ b/.github/workflows/end-to-end.yml
@@ -17,10 +17,10 @@ jobs:
           java-version: '11'
           distribution: 'adopt'
 
-      - name: Setup Nextflow latest-edge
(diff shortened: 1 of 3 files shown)


harmonbhasin:
Updated github actions to now work, and added changes from master, that were not merged into dev.

willbradshaw:
@harmonbhasin do you want to add anything to CHANGELOG for this?

harmonbhasin:
@willbradshaw I don't think I'll add anything to the CHANGELOG for this commit, we'll be pushing the output end-to-end test to dev soon, I'll cover the splitting of these tests in that CHANGELOG commit.

PR #132, opened 2024-12-17
Implementing an end-to-end test that verifies the run workflow output

harmonbhasin:


willbradshaw:
I ran this on my own EC2 instance and the tests passed and failed as expected!

@harmonbhasin I know you have some more stuff you want to do here so let me know when you want me to review again. Remember to modify the CHANGELOG with your changes.

PR #138, opened 2025-01-10
Added directory for docs, and added our version schema for future use

harmonbhasin:
I still need to update the changelog, but we should first change the versions in the changelog to use the new format "Major.Schema.Results.Point". @willbradshaw how do we want to go about updating the prior versions?


willbradshaw:
@harmonbhasin I decided to copy these changes over to a new branch that I could merge directly into master without pulling in the Atria changes. Closing this now.

PR #141, opened 2025-01-10
Replace Trimmomatic with Atria

harmonbhasin:
Found a replacement for trimmomatic that does approximately as well, but takes significantly less time.

As per Will's request, I will describe the changes that were made to the gold output:

- bracken_reports_merged: No changes seen and  expected
- kraken_reports_merged: No changes seen and expected
- qc_adapter_stats: No changes seen and expected
- qc_basic_stats: No changes seen and expected
- qc_quality_base_stats: No changes seen and expected
- qc_quality_sequence_stats: No changes seen and expected
- **virus_hits_1**: This is a list of reads for BLAST. We see the same reads show up, the difference is that some of these reads are slightly longer or shorter, which is expected given that the adapter trimming softwares, Trimmomatic and Atria, are doing something slightly different.
- **virus_hits_2**: Same as above but for reverse read.
- **virus_hits_db**: All of the seq_id's and taxid's are the same, however the mapping of 1 read's taxid is a bit different (1/8) as the reads gained/loss a few base pairs on both the forward and reverse read. The other reads have this as well, but not to the extent that it changes it's taxonomic assignment.
- **virus_clade_counts**: The number of read counts for viral clades changed. Read counts either went from a genus to species, or species to genus, which makes sense given that a few base pairs are changing on each read.
- **blast_hits_paired**: A few BLAST scores changed slightly, but all hit's remained the same, this is again expected given that the reads are slightly different.


configs/containers.config
```diff
@@ -73,4 +73,7 @@ process {
     withLabel: fastp {
         container = "staphb/fastp:0.23.4"
     }
+    withLabel: atria {
+        container = "harmonb/atria:latest"
```
willbradshaw, line 77: Can you add details about the container composition here in case we need to rebuild it? Package versions, that sort of thing.

modules/local/atria/main.nf
```diff
@@ -0,0 +1,16 @@
...
+    label "large"
+    input:
+        // reads is a list of two files: forward/reverse reads
+        tuple val(sample), path(reads)
+        val(adapters_ch)
+    output:
+        tuple val(sample), path("*{1,2}.atria.fastq.gz"), emit: reads
+    shell:
+        '''
+        par="--adapter1 !{adapters_ch.join(' ')} --adapter2 !{adapters_ch.join(' ')} --kmer-tolerance 2 --kmer-n-match 9 --trim-score-pe 10.0 --quality-score 10 --quality-kmer 4 --quality-score 15 --length-range 20:999999 --no-consensus"
```
willbradshaw, line 12: We have two different --quality-score arguments here, 10 and 15. Should there only be one?

subworkflows/local/extractViralReads/main.nf
```diff
@@ -63,11 +64,24 @@ workflow EXTRACT_VIRAL_READS {
         bbm_human_index_path = "${ref_dir}/results/bbm-human-index"
         bbm_other_index_path = "${ref_dir}/results/bbm-other-index"
         virus_db_path = "${ref_dir}/results/total-virus-db-annotated.tsv.gz"
+
+        // Load in adapter channel
+        adapters_ch = Channel
+            .fromPath(adapter_path)
```
willbradshaw, line 70: Consider using [splitFasta](https://www.nextflow.io/docs/latest/reference/operator.html#splitfasta) here. (Not worth more than 20 minutes of your time to change.)

modules/local/atria/main.nf
```diff
@@ -0,0 +1,16 @@
+process ATRIA {
```
willbradshaw, line 1: @harmonbhasin can you also delete the Trimmomatic module? We're not using it anywhere else and I don't think we'll want it back.

willbradshaw (changes requested):

configs/containers.config
harmonbhasin: I added the Dockerfile.

modules/local/atria/main.nf
harmonbhasin: Good catch, I went through the code and they select for the second argument if there are two, so we'll stick with that.

subworkflows/local/extractViralReads/main.nf
harmonbhasin: Great suggestion, this worked well!

modules/local/atria/main.nf
harmonbhasin: It is now deleted.

willbradshaw (approved):

modules/local/atria/main.nf
willbradshaw: OK cool, potentially something to look into if we want to optimize Atria's performance later.

PR #144, opened 2025-01-13
Bbduk rearrangement

harmonbhasin:
Still need to update the CHANGELOG, however I'm not sure what this change would entail.

Changes:

- Remove raw and clean subworkflow
- Move fastp into profile and extract_viral_reads subworkflow
- Only do FASTQC in profile subworkflow, once before fastp, and once after fastp
- Fixed LOAD_SAMPLESHEET subworkflow to be correct.
- Updated output tests
- Added COUNT_TOTAL_READS subworkflow to count the total reads for each sample


workflows/run.nf
```diff
@@ -53,9 +52,9 @@ workflow RUN {
         blast_paired_ch = Channel.empty()
     }
     // Taxonomic profiling
-    PROFILE(CLEAN.out.reads, group_ch, kraken_db_path, params.n_reads_profile, params.ref_dir, "0.4", "27", "ribo", params.grouping, params.single_end)
+    PROFILE(samplesheet_ch, group_ch, kraken_db_path, params.n_reads_profile, params.ref_dir, "0.4", "27", "ribo", params.grouping, params.adapters, "2", "4 GB", params.single_end)
```
willbradshaw, line 55: Now that we're also using the subsetting for QC as well as for profiling, the current setup is a bit confusing to users IMO. I think it might make sense to split up PROFILE into 2-3 subworkflows instead of loading everything into one. E.g.:

1. Subset and FASTP, return pre- and post-FASTP reads in different channels
2. Run FASTQC and MultiQC, process and output results (replaces current QC and PROCESS_OUTPUT workflows)
3. Rest of profiling workflow

@harmonbhasin wdyt?

willbradshaw: (The changes here look fine other than that)

subworkflows/local/qc/main.nf
```diff
@@ -34,10 +34,10 @@ workflow QC {
         multiqc_qbase_ch = process_ch.map{ it[2] }.collect().ifEmpty([])
         multiqc_qseqs_ch = process_ch.map{ it[3] }.collect().ifEmpty([])
         // 5. Merge MultiQC outputs
-        basic_out_ch = MERGE_MULTIQC_BASIC(multiqc_basic_ch, "${stage_label}_qc_basic_stats")
-        adapt_out_ch = MERGE_MULTIQC_ADAPT(multiqc_adapt_ch, "${stage_label}_qc_adapter_stats")
-        qbase_out_ch = MERGE_MULTIQC_QBASE(multiqc_qbase_ch, "${stage_label}_qc_quality_base_stats")
-        qseqs_out_ch = MERGE_MULTIQC_QSEQS(multiqc_qseqs_ch, "${stage_label}_qc_quality_sequence_stats")
+        basic_out_ch = MERGE_MULTIQC_BASIC(multiqc_basic_ch, "${stage_label}_subsetted_qc_basic_stats")
```
willbradshaw, line 37: Abject nitpick but it's going to bother me forever if I don't address it now: can we use "subset" instead of "subsetted" here?

tests/subworkflows/local/countTotalReads/main.nf.test
```diff
@@ -0,0 +1,37 @@
+nextflow_workflow {
```
willbradshaw, line 1: @harmonbhasin thanks for adding some new tests here!

I'd also like to see unit tests for COUNT_READS and COMBINE_READ_COUNTS, ideally ones that go beyond just success/failure. You can use the [csv plugin](https://github.com/lukfor/nft-csv) to query the outputs against some toy data file and confirm that it gives the expected results.

subworkflows/local/profile/main.nf
```diff
@@ -74,9 +81,20 @@ workflow PROFILE {
         } else {
             grouped_ch = subset_ch
         }
+
```
willbradshaw, line 84: See my comments on run.nf about whether we should load all of this into PROFILE or split the subsetting and QC out into different subworkflows.

tests/main.nf.test.snap
```diff
@@ -3,10 +3,10 @@
         "content": [
             "bracken_reports_merged.tsv.gz:md5,d116fc06ea955409463686df223ec413",
             "kraken_reports_merged.tsv.gz:md5,5d04e8270bfddc7f435bcc0a76bafbcb",
-            "qc_adapter_stats.tsv.gz:md5,a6441f841e6e1743067f6eb1bb68a3ef",
-            "qc_basic_stats.tsv.gz:md5,fd1a20ad7bb06eecaaa9cc51b81efa8d",
-            "qc_quality_base_stats.tsv.gz:md5,1e75fe721479280f83d7e2f8c9fa975b",
-            "qc_quality_sequence_stats.tsv.gz:md5,0d835c1d9417df961b4aadd7ac42137d",
+            "qc_adapter_stats.tsv.gz:md5,5c76643e879a3ad34f18833445c98c1b",
```
willbradshaw, line 6: @harmonbhasin as with your last PR, if we're going to override a failure of test-run-output I'd like a file-by-file summary of what changes occurred in the test output files.

willbradshaw (changes requested):
Great first pass, thanks @harmonbhasin!

modules/local/countReads/main.nf
```diff
@@ -0,0 +1,14 @@
...
+    label "countReads"
+    label "single_cpu_32GB_memory"
+    input:
+        tuple val(sample), path(reads)
+    output:
+        path("${sample}_read_count.txt")
+    shell:
+        '''
+        # Count reads in file
+        COUNT=$(bioawk -c fastx 'END{print NR}' !{reads[0]})
```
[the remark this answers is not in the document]
harmonbhasin: Yup, this sounds good, I've ported this over!

modules/local/combineReadCounts/main.nf
```diff
@@ -0,0 +1,17 @@
...
+    label "coreutils_gzip_gawk"
+    label "single"
+    input:
+        path(read_counts)
+    output:
+        path("read_counts.tsv.gz")
+    shell:
+        '''
+        # Add header
+        echo "sample,total_read_count" > read_counts.tsv
```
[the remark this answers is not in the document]
harmonbhasin: Good catch, I've switched the delimiter in countReads, this script has also been replaced with MERGE_TSV, I just had to add the header in countReads.

modules/local/combineReadCounts/main.nf
```diff
@@ -0,0 +1,17 @@
...
+        # Add read counts
+        cat !{read_counts.join(" ")} >> read_counts.tsv
```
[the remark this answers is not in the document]
harmonbhasin: Yup this is possible and I did this, thanks for recommending this! The next push I make will remove this process.

subworkflows/local/qc/main.nf
harmonbhasin: Yup, changed!

tests/subworkflows/local/countTotalReads/main.nf.test
harmonbhasin: Yes this sounds good! I'm assuming that I can still use the plugin and just specify other delimiters, but I'll be sure to do this.

workflows/run.nf
harmonbhasin: I did 2. slightly different than I think you stated up there (I kept the QC module), let me know if this bothers you and I can just call the RUN_QC module twice).

tests/subworkflows/local/countTotalReads/main.nf.test
harmonbhasin: I'm assuming you'll have a test implemented for MERGE_TSV so I did not make one, but I did make one for COUNT_READS. Let me know if you want one for MERGE_TSV.

workflows/run.nf
```diff
@@ -52,11 +54,16 @@ workflow RUN {
...
         blast_paired_ch = Channel.empty()
     }
-    // Taxonomic profiling
-    PROFILE(CLEAN.out.reads, group_ch, kraken_db_path, params.n_reads_profile, params.ref_dir, "0.4", "27", "ribo", params.grouping, params.single_end)
-    // Process output
-    qc_ch = RAW.out.qc.concat(CLEAN.out.qc)
-    PROCESS_OUTPUT(qc_ch)
+
+    // Subset reads to target number, and trim adapters
+    SUBSET_AND_TRIM_READS(samplesheet_ch, group_ch, params.n_reads_profile, params.grouping, params.adapters, params.single_end)
```
willbradshaw, line 59: Another nitpick, but I'd prefer something shorter. Maybe just SUBSET_TRIM?

tests/subworkflows/local/countTotalReads/main.nf.test
willbradshaw: Yeah I have tests for MERGE_TSV ready to merge soon. 

Can you add anything to this test (for COUNT_TOTAL_READS) for anything other than success/failure?

willbradshaw (changes requested):

workflows/run.nf
willbradshaw: OK, this is fine

willbradshaw:
@harmonbhasin at this point I think it would be a good idea to merge in `dev`, resolve any conflicts, and add a CHANGELOG entry. I think this is still a Results change since it doesn't change or delete any output files (just adds a new one)?

subworkflows/local/runQc/main.nf
willbradshaw, line 1: @harmonbhasin this split could be a good opportunity to write some tests here

subworkflows/local/subsetAndTrimReads/main.nf
willbradshaw, line 1: @harmonbhasin this split could be a good opportunity to write some tests here

harmonbhasin:
> @harmonbhasin at this point I think it would be a good idea to merge in dev, resolve any conflicts, and add a CHANGELOG entry. I think this is still a Results change since it doesn't change or delete any output files (just adds a new one)?

Sounds good, I actually already merged into dev and resolved the conflicts when I sent out this last push. I haven't updated the CHANGELOG entry though so I'll do that in my next push.

subworkflows/local/runQc/main.nf
willbradshaw: Actually, I want to get this merged soon so I can work from it. I expect I'll end up adding tests here later anyway.

subworkflows/local/subsetAndTrimReads/main.nf
willbradshaw: Actually, I want to get this merged soon so I can work from it. I expect I'll end up adding tests here later anyway.

tests/subworkflows/local/runQc/main.nf.test
willbradshaw, line 1: @harmonbhasin does this actually do anything right now, since you're not specifying any inputs? If not I'd just delete it, we can add some actually working tests later.

tests/subworkflows/local/subsetTrim/main.nf.test
```diff
@@ -0,0 +1,45 @@
...
+                input[4] = params.adapters // adapter_path
+                input[5] = params.single_end // single_end
+                """
+            }
+        }
+
+        then {
+            assert workflow.success
+            def countGzipLines = { file -> path(file).linesGzip.size() }
+            assert countGzipLines(workflow.out.subset_reads[0][1][0]) == 165 * 4
```
willbradshaw, line 39: @harmonbhasin can you explain these 165 and 135 numbers?

willbradshaw (approved):

PR #149, opened 2025-01-20
Remove Atria and BBMAP

harmonbhasin:
After discussion with Will and Jeff, we've decided to remove ATRIA and BBMAP.

I updated this CHANGELOG under the assumption that we upload current dev to master, then merge this into dev. 

Changes:

- Remove ATRIA and BBMAP from the pipeline.

Changes in gold standard output:

- virus_hits_db.tsv: 24 of the reads changed; with new we gained a read, and 3 out of the 24 had changes in read id. This can be attributed to removing Atria as it won’t be trimming some adapters.
- virus_clade_counts.tsv: Total count of viruses went up one in the new output, which follows the understanding above. We see reads either going up one taxa or down one taxa.
- virus_hits_1.fasta: There is one additional read in the new output which follows the above, and only a few of the other reads change in the front or end of their read, due to atria being removed (and hence adapters not being trimmed).
- virus_hits_2.fast: Same trend as above.
- blast_hits_paired.tsv: Two new reads are identified in the new output, due to the lack of adapter trimming, and for the rest of the reads, their taxid remains the same, but the BLAST metrics slightly change, which can be attributed to the addition of base pairs that are most likely from adapters.


CHANGELOG.md
```diff
@@ -1,4 +1,7 @@
-# v2.6.1.0 (in progress)
+# v2.6.1.1
```
willbradshaw, line 1: I'm planning on merging this before we merge to master, so this can go under 2.6.1.0.

subworkflows/local/extractViralReads/main.nf
```diff
@@ -108,10 +101,8 @@ workflow EXTRACT_VIRAL_READS {
         // Filter contaminants
         human_bt2_ch = BOWTIE2_HUMAN(bowtie2_reads_combined_ch, bt2_human_index_path, "", "human")
         other_bt2_ch = BOWTIE2_OTHER(human_bt2_ch.reads_unconc, bt2_other_index_path, "", "other")
-        human_bbm_ch = BBMAP_HUMAN(other_bt2_ch.reads_unconc, bbm_human_index_path, "human")
-        other_bbm_ch = BBMAP_OTHER(human_bbm_ch.reads_unmapped, bbm_other_index_path, "other")
         // Run Kraken on filtered viral candidates
-        tax_ch = TAXONOMY(other_bbm_ch.reads_unmapped, kraken_db_ch, true, "F", single_end)
+        tax_ch = TAXONOMY(human_bt2_ch.reads_unconc, kraken_db_ch, true, "F", single_end)
```
willbradshaw, line 105: This should be other_bt2_ch

willbradshaw (changes requested):

PR #154, opened 2025-01-22
Fix small bug when extracting read type from bowtie2

harmonbhasin:
Updated the regex for a grep such that we don't pull out the wrong column from SAM files.


willbradshaw:
@harmonbhasin have you confirmed the samples that were previously failing now run to completion?

harmonbhasin:
> @harmonbhasin have you confirmed the samples that were previously failing now run to completion?

@willbradshaw the pipelline is still running, but yes, the reads don't fail on that step anymore.

willbradshaw (approved):

PR #155, opened 2025-01-22
Replace generate_samplesheet.sh with generate_samplesheet.py

harmonbhasin:
Ported the script over to python, tested it on the following combinations:

- local: single end
- local: paired end
- s3: single end
- s3: paired end
- s3: paired end group_file
- s3: paired end group_across_illumina_lanes

I also added some new exceptions, such as throwing an error if the user has s3 in their dir_path and hasn't passed the s3 flag, and thorwing an error if the user specifies single-end but has forward and reverse suffix.


harmonbhasin:
Going to push the change for Jeff's error message.

harmonbhasin:
In the latest version of the documentation, we're going to say that the column headers in the samplesheet are required, however for this happen, I would have to change the LOAD_SAMPLESHEET to require this, not the generate_samplesheet.py script, is this something I should do now?

willbradshaw:
Yes, I would make any changes like that in this branch too

harmonbhasin:
I need to write tests for this!

tests/subworkflows/local/loadSampleSheet/main.nf.test
```diff
@@ -0,0 +1,74 @@
...
+        then {
+            assert workflow.success
+            assert workflow.out.samplesheet.contains(
+                ['gold_standard', ["/nao-testing/gold-standard-test/raw/gold_standard_R1.fastq.gz", "/nao-testing/gold-standard-test/raw/gold_standard_R2.fastq.gz"]]
+            )
+            assert workflow.out.samplesheet.size() == 1
+            assert workflow.out.group.size() == 0
+        }
+    }
+    test("Test normal paired-end samplesheet without grouping") {
```
willbradshaw, line 29: With grouping?

tests/subworkflows/local/loadSampleSheet/main.nf.test
willbradshaw, line 1: Could you add 1-2 tests for single-end?

willbradshaw (changes requested):

tests/subworkflows/local/loadSampleSheet/main.nf.test
harmonbhasin: Yes!

willbradshaw (approved):

PR #165, opened 2025-01-28
Merge v2.7.0.0 to master

willbradshaw:


configs/containers.config
```diff
@@ -67,10 +67,13 @@ process {
...
-    withLabel: trimmomatic {
-        container = "staphb/trimmomatic:0.39"
-    }
     withLabel: fastp {
         container = "staphb/fastp:0.23.4"
     }
+    withLabel: filtlong {
+        container = "staphb/filtlong:0.2.1"
+    }
+    withLabel: atria {
```
willbradshaw, line 76: @harmonbhasin looks like this snuck through. Let's delete and re-merge to dev before we merge to master.

subworkflows/local/extractViralReads/main.nf
```diff
@@ -49,29 +48,38 @@ workflow EXTRACT_VIRAL_READS {
...
         bt2_virus_index_path = "${ref_dir}/results/bt2-virus-index"
         bt2_human_index_path = "${ref_dir}/results/bt2-human-index"
         bt2_other_index_path = "${ref_dir}/results/bt2-other-index"
-        bbm_human_index_path = "${ref_dir}/results/bbm-human-index"
-        bbm_other_index_path = "${ref_dir}/results/bbm-other-index"
         virus_db_path = "${ref_dir}/results/total-virus-db-annotated.tsv.gz"
-        // Run initial screen against viral genomes with BBDuk
+
+        // Read adapter file and create a channel containing just the adapter sequences
+        adapters_ch = Channel
```
willbradshaw, line 64: @harmonbhasin I think this now isn't doing anything since we're no longer using Atria? Can you confirm?

harmonbhasin: Yes, good catch! We don't need this anymore, I'll remove it.

configs/containers.config
harmonbhasin: Yes, that's quite strange as we removed it here, https://github.com/naobservatory/mgs-workflow/pull/149/files. I'll remove this though and make a new branch.

harmonbhasin (approved):
LGTM!

PR #168, opened 2025-01-29
Run nf-test tests in parallel.

harmonbhasin:
Added new script, `bin/run_parallel_test.sh`, that allows users to run nf-test tests locally in parallel


bin/run_parallel_test.sh
```diff
@@ -0,0 +1,64 @@
...
+    print_usage
+fi
+
+# Store exit statuses
+declare -a pids
+declare -a statuses
+
+# Download any plugins as running in parallel doesn't guarantee that they install properly
+nf-test update-plugins
+sleep 10
```
willbradshaw, line 33: 10 seconds seems like a long time to wait

bin/run_parallel_test.sh
```diff
@@ -0,0 +1,64 @@
...
+    wait "$pid"
+    statuses+=($?)
+done
+
+# Check exit statuses
+all_passed=true
+for status in "${statuses[@]}"; do
+    if [ "$status" -ne 0 ]; then
+        all_passed=false
+        echo "One or more test shards failed with exit status: $status"
```
willbradshaw, line 55: Can we do any better than this? E.g. specify which specific tests failed? I'm worried this will lead to a lot of useful information getting lost and make debugging slow.

willbradshaw (changes requested):

bin/run_parallel_test.sh
harmonbhasin: fair, i'll reduce this to 2 seconds

bin/run_parallel_test.sh
harmonbhasin: yeah we can probably do better, i'll look into this.

bin/run_parallel_test.sh
```diff
@@ -0,0 +1,77 @@
...
+done
+
+# Check exit statuses
+if [ "$failed_shards" == "true" ]; then
+    echo "Some test shards failed. Check the test log at: $error_log"
+    grep "FAILED" "$error_log"
+    exit 1
+else
+    echo "All test shards completed successfully"
+    rm -rf "$temp_dir"
```
willbradshaw, line 73: @harmonbhasin do you mean to only remove the tempdir if all the tests succeed? Why do we need to keep it around if a test fails?

bin/run_parallel_test.sh
```diff
@@ -0,0 +1,77 @@
...
+    exit 0
+fi
+
+rm -rf ${temp_dir}
```
willbradshaw, line 77: Does this do anything? Both branches of the if statement above exit before we get here.

bin/run_parallel_test.sh
harmonbhasin: Oops, this was a mistake, thanks for catching this, it should always remove the tempdir.

willbradshaw (approved):

PR #173, opened 2025-01-30
Pipeline doesn't work with new index

harmonbhasin:
New index, s3://nao-mgs-wb/index/20250128/, uses name "virus-genomes-masked.fasta.gz" over "virus-genomes-filtered.fasta.gz".


harmonbhasin:
Ah, this will cause the tests to fail.

PR #174, opened 2025-01-30
Fixing fatal bug

harmonbhasin:
New index uses virus-genomes-masked.fasta.gz instead of virus-genome-filtered.fasta.gz.

Creating a new index as we speak.


harmonbhasin:
Snapshot needs to be updated for run output.

harmonbhasin:
Changes to gold standard output:
- blast_hits_paired.tsv.gz: All that changed were the blast scores, which makes sense given that we applied the masking that katherine made to the final BLAST DB.

willbradshaw (approved):

PR #179, opened 2025-02-03
Fixing misused labels

harmonbhasin:
[Nextflow](https://github.com/nextflow-io/nextflow/issues/4730) doesn't check if labels exist, which allows people to misspell labels. In that situation, nextflow will use whatever is available on the machine. This allowed some tests to pass using nf-test, but not using AWS Batch, since someone using nf-test probably has some pre-installed packages on their local machine. Additionally, two processes were mislabeled.


harmonbhasin:
Tests will fail until master is merged into dev.

willbradshaw (approved):

PR #182, opened 2025-02-05
Removed defaults from config files

harmonbhasin:
Encourages users to read the manuals and understand how the pipeline operates.


configs/run_validation.config
```diff
@@ -6,12 +6,12 @@ params {
...
     // Directories
-    base_dir = "s3://nao-mgs-wb/test-remote" // Parent for working and output directories (can be S3)
-    ref_dir = "s3://nao-mgs-wb/index/20241209/output" // Reference/index directory (generated by index workflow)
+    base_dir = <PATH TO YOUR DIRECTORY> // Parent for working and output directories (can be S3)
+    ref_dir = <PATH TO YOUR ORGS REFERENCE DIRECTORY> // Reference/index directory (generated by index workflow)
 
     // Files
-    viral_tsv = "${projectDir}/test-data/gold-standard-results/virus_hits_filtered.tsv.gz"
-    viral_fastq = ""
+    viral_tsv = <PATH TO 'virus_hits_db.tsv.gz` FROM OUTPUT OF RUN WORKFLOW>
```
willbradshaw, line 13: @harmonbhasin we should clarify that one or other of these should be blank (see logic in pipeline for which is the default)

configs/run.config
```diff
@@ -6,14 +6,14 @@ params {
...
 
     // Directories
-    base_dir = "s3://nao-mgs-wb/test-batch" // Parent for working and output directories (can be S3)
-    ref_dir = "s3://nao-mgs-wb/index/20241209/output" // Reference/index directory (generated by index workflow)
+    base_dir = <PATH TO YOUR DIRECTORY> // Parent for working and output directories (can be S3)
+    ref_dir = <PATH TO YOUR ORGS REFERENCE DIRECTORY> // Reference/index directory (generated by index workflow)
 
     // Files
-    sample_sheet = "${launchDir}/samplesheet.csv" // Path to library TSV
+    sample_sheet = <PATH TO YOUR SAMPLING SHEET> // Path to library TSV
```
willbradshaw, line 16: I think this one can stay as-is

willbradshaw (changes requested):

configs/run_validation.config
harmonbhasin: Sounds good

PR #194, opened 2025-02-11
Fix validation bug

harmonbhasin:
Old parameter was kept from previous version of pipeline, so this has been updated in run_validation workflow.


willbradshaw (approved):

willbradshaw:
@harmonbhasin, looks good, you can go ahead and merge

PR #199, opened 2025-02-12
Making PROFILE take in ONT data

simonleandergrimm:
This PR makes PROFILE take in ONT data. To achieve this, I added a second samtools process, and ribo identification using minimap2. Again, @harmonbhasin I'm using my own ribo reference for minimap2. Let me know if I should add the creation of said reference to the index workflow.


harmonbhasin (changes requested):
LGTM. I would add the creation of the reference to the index workflow. Once you do that I can approve this.

tests/modules/local/minimap2/main.nf.test
```diff
@@ -42,4 +42,28 @@ nextflow_process {
...
+
+    test("When run against ribo index, should run without failures and return properly formatted SAM") {
+        tag "expect_success"
+        tag "single_end"
+        when {
+            params {}
+            process {
+                '''
+                input[0] = LOAD_SAMPLESHEET.out.samplesheet
+                input[1] = "s3://nao-mgs-simon/ont-indices/2024-12-14/minimap2-ribo-index/ribo-ref-concat-unique.mmi"
```
jeffkaufman, line 54: ditto

subworkflows/local/profile/main.nf
```diff
@@ -27,13 +32,19 @@ workflow PROFILE {
...
+        if (params.ont) {
+            ribo_path = "s3://nao-mgs-simon/ont-indices/2024-12-14/minimap2-ribo-index/ribo-ref-concat-unique.mmi"
+            mapped_ch = MINIMAP2_RIBO(reads_ch, ribo_path, ribo_suffix)
+            ribo_ch = SAMTOOLS_SEPARATE(mapped_ch, ribo_suffix)
+        } else {
+            ribo_path = "${ref_dir}/results/ribo-ref-concat.fasta.gz"
+            ribo_ch = BBDUK(reads_ch, ribo_path, min_kmer_fraction, k, ribo_suffix, !single_end)
+        }
         // Run taxonomic profiling separately on ribo and non-ribo reads
         tax_ribo_ch = TAXONOMY_RIBO(ribo_ch.match, kraken_db_ch, "D", bracken_threshold, single_end)
```
jeffkaufman, line 49: FYI I expect Kraken to do a bad job on generating a taxonomic profile here

But I still think this PR is in the right direction: if we want to do better taxonomic profiling for ONT data that should be a separately-prioritized followup

jeffkaufman (changes requested):

modules/local/samtools/main.nf
```diff
@@ -18,4 +18,29 @@ process SAMTOOLS_FILTER {
         # Link input to output for testing
         ln -s !{sam} !{sample}_in.sam
         '''
+}
+
+// Return aligned and unaligned reads separately as FASTQs
+process SAMTOOLS_SEPARATE {
```
willbradshaw, line 24: This process is a superset of SAMTOOLS_FILTER. You don't need both.

(I would implement this as two calls to a SAMTOOLS_FASTQ subworkflow with different input parameters, followed by a channel merge.)

subworkflows/local/profile/main.nf
```diff
@@ -27,13 +32,19 @@ workflow PROFILE {
...
+        ribo_suffix
         bracken_threshold
         single_end
     main:
         // Separate ribosomal reads
-        ribo_path = "${ref_dir}/results/ribo-ref-concat.fasta.gz"
-        ribo_ch = BBDUK(reads_ch, ribo_path, min_kmer_fraction, k, bbduk_suffix, !single_end)
+        if (params.ont) {
+            ribo_path = "s3://nao-mgs-simon/ont-indices/2024-12-14/minimap2-ribo-index/ribo-ref-concat-unique.mmi"
+            mapped_ch = MINIMAP2_RIBO(reads_ch, ribo_path, ribo_suffix)
```
willbradshaw, line 42: Speaking from ignorance here, but why should BBDuk not work on ONT data? It's calculating a fraction of matching k-mers so read length per se shouldn't be fatal.

tests/modules/local/minimap2/main.nf.test
```diff
@@ -42,4 +42,28 @@ nextflow_process {
...
+                input[2] = "ribo"
+                '''
+            }
+        }
+        then {
```
willbradshaw, line 59: As commented in another PR, I'd prefer a test of the actual content, not just that there is content.

willbradshaw (changes requested):

tests/modules/local/minimap2/main.nf.test
simonleandergrimm: This is now fixed.

simonleandergrimm:
@willbradshaw Before creating the new index, tests here will fail as there is currently no ribo index for minimap2. It's still be good to get your input here.

subworkflows/local/profile/main.nf
```diff
@@ -27,16 +29,35 @@ workflow PROFILE {
...
+            ribo_ch = mapped_ch
+        } else {
+            ribo_path = "${ref_dir}/results/ribo-ref-concat.fasta.gz"
+            ribo_ch = BBDUK(reads_ch, ribo_path, min_kmer_fraction, k, ribo_suffix, !single_end)
+        }
         // Run taxonomic profiling separately on ribo and non-ribo reads
-        tax_ribo_ch = TAXONOMY_RIBO(ribo_ch.match, kraken_db_ch, "D", bracken_threshold, single_end)
-        tax_noribo_ch = TAXONOMY_NORIBO(ribo_ch.nomatch, kraken_db_ch, "D", bracken_threshold, single_end)
+        tax_ribo_ch = TAXONOMY_RIBO(
+            ont ? ribo_ch.reads_mapped : ribo_ch.match,
```
simonleandergrimm, line 48: not particularly happy with this, but I don't want to call the output of minimap2 "match" and "no_match"

willbradshaw: I don't think this is terrible, but given that the two branches of the if statement produce quite different things (k-mer matches vs alignments) it might be better to give them different names in the if statement, then define some new variables (maybe ribo_in and noribo_in) to hold them. Then the calls to TAXONOMY can take ribo_in and noribo_in as their first input.

Also, minor, but I'd prefer if we went back to having multiple arguments per line, I'm not a fan of these very long one-arg-per-line function calls.

tests/modules/local/minimap2/main.nf.test
```diff
@@ -0,0 +1,71 @@
...
+            def input_reads = path(process.out.input[0][1]).fastq.readNames.toSet()
+
+            // Verify read partitioning is correct
+            assert sam_read_ids_mapped == fastq_read_ids_mapped
+
+            // Verify no overlapping reads between mapped and unmapped sets
+            assert fastq_read_ids_mapped.intersect(fastq_read_ids_unmapped).size() == 0
+
+            // Verify input FASTQ contains all reads
+            assert input_reads == fastq_read_ids_mapped + fastq_read_ids_unmapped
```
willbradshaw, line 67: Does this pass? I'd expect you to need some sorting.

modules/local/minimap2/main.nf
```diff
@@ -0,0 +1,41 @@
...
+        al="!{sample}_!{suffix}_minimap2_mapped.fastq.gz"
+        un="!{sample}_!{suffix}_minimap2_unmapped.fastq.gz"
+
+        # Run pipeline
+        # Outputs a SAM file for all reads, which is then partitioned based on alignment status
+        #   - First branch (samtools view -u -f 4 -) filters SAM to unaligned reads and saves FASTQ
+        #   - Second branch (samtools view -u -F 4 -) filters SAM to aligned reads and saves FASTQ
+        #   - Third branch (samtools view -h -F 4 -) also filters SAM to aligned reads and saves SAM
+        zcat !{reads} \
+            | minimap2 -a ${idx} /dev/fd/0 \
```
willbradshaw, line 30: Do you expect to be able to use all default parameters every time you call minimap2? I'd be surprised if so. If not, you might want to add some configurability.

subworkflows/local/profile/main.nf
willbradshaw: @simonleandergrimm I'd still like this to be addressed before we move forward here, I remain uneasy about switching to an alignment-based approach without having a good understanding of how this might affect the results.

simonleandergrimm: On priors I'd assume minimap2 to be better than bbduk (I'd assume the latter needs a bunch of playing around to find the fraction of matching k-mers that gives equivalent performance to minimap2).

simonleandergrimm: Finally, I'd be pretty keen to get the ONT version of the pipeline up and running fast and do optimization later (especially because this part of the pipeline shouldn't affect the Illumina runs).

tests/modules/local/minimap2/main.nf.test
simonleandergrimm: It passes. I think sets normally do not need parsing?

modules/local/minimap2/main.nf
simonleandergrimm: The settings are given during the index creation, so there's no need to pass parameters here for now.

PR #206, opened 2025-02-21
Updated annotateVirusInfection to better handle viruses missing from Virus-Host DB

willbradshaw:


harmonbhasin (dismissed):
Overall, LGTM

modules/local/annotateVirusInfection/resources/usr/bin/annotate-viral-hosts.py
```diff
@@ -178,32 +178,43 @@ def expand_taxids(taxids):
     return taxids_new
 
 def mark_descendant_infections(virus_tree: Dict[str, Set[str]],
```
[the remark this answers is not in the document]
harmonbhasin: Is it possible that we switch from the numbers to something more interpretable like labels?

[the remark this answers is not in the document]
willbradshaw: > Is it possible that we switch from the numbers to something more interpretable like labels?

Interesting idea! I don't hate it, but some of the meanings seem specific enough that finding good concise labels seems tricky.

We could have "0 (Inconsistent)" "1 (Match)" "2 (Unclear)" and "3 (Consistent)"? Not sure about 2...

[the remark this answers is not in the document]
harmonbhasin: I would heavily lean towards finding labels that accurately describe the numbers even if they're a bit verbose. Given that it is so unclear, I think it makes it more important that we have these labels.

docs/annotation.md
```diff
@@ -0,0 +1,47 @@
...
+
+- An infection status of **1** designates taxa that are affirmatively marked as infecting that host group. It includes taxa that are directly marked in Virus-Host DB as infecting that host group; taxa that are descended from such taxa; and taxa whose descendants are all marked 1.
+- An infection status of **0** designates taxa that are marked as not infecting that host group. It includes taxa that: are included in Virus-Host DB; are not marked in Virus-Host DB as infecting that host group; and have no 1-marked descendants. It also includes taxa that are descended from such taxa, and taxa whose descendants are all marked 0.
+- An infection status of **3** designates taxa that are likely to infect that host group, but cannot be affirmatively marked as such. It includes taxa that are not included in Virus-Host DB, have at least one 1-marked descendant, and no 0-marked descendants. It also includes all taxa that are descended from such taxa and cannot be marked 1 according to the rules above, as well as taxa whose descendants are all marked 3.
```
harmonbhasin, line 7: Could you provide an example for the status of *3* (e.g, use some viral family that follows this example).

docs/annotation.md
```diff
@@ -0,0 +1,47 @@
+# Viral infection status annotation
+
+As part of the [index workflow](./index.md), the pipeline generates a database of all viral taxa from the [NCBI Taxonomy](https://www.ncbi.nlm.nih.gov/taxonomy), annotating each with its taxonomic ID (taxid), parent taxid in the taxonomy tree, and infection status for a range of host taxa of interest (e.g. humans, vertebrates). This host infection status is determined from the infection information provided by [Virus-Host DB](https://www.genome.jp/virushostdb/), and is stored as a four-state integer series for each host taxon of interest:
+
+- An infection status of **1** designates taxa that are affirmatively marked as infecting that host group. It includes taxa that are directly marked in Virus-Host DB as infecting that host group; taxa that are descended from such taxa; and taxa whose descendants are all marked 1.
```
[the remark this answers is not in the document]
harmonbhasin: I still think we should have the label in the code such that a dev can read through that file and not have to jump anywhere to find the interpretation.

docs/annotation.md
```diff
@@ -0,0 +1,47 @@
...
+- An infection status of **3** designates taxa that are likely to infect that host group, but cannot be affirmatively marked as such. It includes taxa that are not included in Virus-Host DB, have at least one 1-marked descendant, and no 0-marked descendants. It also includes all taxa that are descended from such taxa and cannot be marked 1 according to the rules above, as well as taxa whose descendants are all marked 3.
+- Finally, an infection status of **2** designates taxa of uncertain infection status. It includes all other taxa: those that are not included in Virus-Host DB and have both 0- and 1-marked descendants; those descended from a 2-marked taxon that are not given another status by the rules above; and those whose descendants are all marked 2.
+
+Some taxa are *hard-excluded* by the user via the index workflow's config file; any such taxon, and all its descendants, are automatically marked with an infection status of 0. This is useful for cases where we believe Virus-Host DB has misannotated certain taxa (such as human-gut-colonizing phages) in ways that give rise to misleading results in our pipeline.
+
+```mermaid
```
harmonbhasin, line 12: This diagram is a life saver, thank you.

modules/local/annotateVirusInfection/resources/usr/bin/annotate-viral-hosts.py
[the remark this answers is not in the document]
harmonbhasin: an alternative is defining some CONSTANTS at the top of the file that are more descriptive of the number, but just point to it, e.g. INCONSISTENT=0, MATCH=1.

modules/local/filterVirusReads/resources/usr/bin/filter_virus_reads.py
harmonbhasin, line 1: Line 47 says the following:
`kraken_assigned_host_virus = int(fields[idx["kraken_assigned_host_virus"]]) # 3-state: 0, 1, or 2`

should the comment be updated to say "# 4-state: 0, 1, 2, or 3".

modules/local/filterVirusReads/resources/usr/bin/filter_virus_reads.py
```diff
@@ -57,7 +57,7 @@ def filter_virus_reads(input_path, score_threshold, out_path):
                 msg = msg+"\tKEEP"
                 print_log(msg)
                 outf.write("\t".join(fields) + "\n")
-            elif adj_score >= score_threshold and kraken_assigned_host_virus == 2:
+            elif adj_score >= score_threshold and kraken_assigned_host_virus > 1:
```
harmonbhasin, line 60: Have a comment above this linking to the viral annotation doc `kraken_assigned_host_virus > 1:`

ref/host-taxa.tsv
```diff
@@ -1,6 +1,6 @@
 name	taxid
```
harmonbhasin, line 1: why were these changed?

harmonbhasin (changes requested):
Mostly looks good, just have some comments/suggestions.

configs/index.config
```diff
@@ -35,7 +35,8 @@ params {
 
     // Other input values
     virus_taxid = "10239"
-    viral_taxids_exclude = "2731619 2732413 2732411" // Exclude Caudoviricetes, Malgrantaviricetes, Faserviricetes
+    viral_taxids_exclude_hard = "2731619 2732413 2732411" // Exclude Caudoviricetes, Malgrantaviricetes, Faserviricetes
+    viral_taxids_exclude_soft = virus_taxid + " 12429"
```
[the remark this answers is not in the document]
willbradshaw: I ended up dropping soft exclusion based on talking to Jeff. (though we might end up reintroducing it at some point depending on how some other experiments pan out)

docs/annotation.md
willbradshaw: A good example of this would be sapovirus, which is what originally prompted this revision in the first place. There are clades of sapovirus that are definitely human (e.g. isolated from some Chinese patient) but aren't in Virus-Host DB, so can't be given an annotation of 1, even though they have siblings that are in the DB and are marked 1.

In the old system, because the parent clade didn't have _all_ its children marked 1, it would be marked 2, and then the unmarked children would be marked 0. Now, both the parent and unmarked children are marked 3.

modules/local/filterVirusReads/resources/usr/bin/filter_virus_reads.py
willbradshaw: Yes!

ref/host-taxa.tsv
willbradshaw: Based on discussion with Jeff about what infection status would actually want us to take further action. If we see a chimp virus in US wastewater, there's a good chance it's actually infecting people.

docs/annotation.md
harmonbhasin: This is good to know, thanks! What I meant to say was that could you provide the example in the docs itself, given that the description for 3 is a bit long, I think putting the example you listed above in the docs, could clear things up.

ref/host-taxa.tsv
harmonbhasin: I think it would be good to have the rationale for each of these hosts somewhere in the docs (even if it's just a sentence), would you be fine with adding this?

harmonbhasin (approved):
This looks good to me, the only thing I request is that you list the rationale for each of the host taxa somewhere in the docs if that doesn't already exist.

docs/annotation.md
willbradshaw: I added a footnote (footnote 2)

ref/host-taxa.tsv
willbradshaw: I added a footnote to docs/annotation.md (footnote 1)

harmonbhasin (approved):
LGTM, thank you for making the changes Will!

PR #209, opened 2025-02-28
change error rate

katherine-stansifer:
cutadapt allowed mismatch rate with adapters 33%->12.5%
note that CUTADAPT_MASK had a pre-existing error rate of 20%; I left this as this process isn't currently used anywhere in pipeline.

### DIFF
--- a/modules/local/cutadapt/main.nf
+++ b/modules/local/cutadapt/main.nf
@@ -13,14 +13,14 @@ process CUTADAPT {
         -b (-B for R2 read) to trim any adapter in the adapters file from either end of either read
         -j number of cpu cores
         -m to drop a read pair where either element of the pair is <20 bp after trimming
-        -e 0.33 To allow up to a 33% error rate in the matching region between an adapter
+        -e 0.125 To allow up to a 12.5% error rate in the matching region between an adapter
             and the read
         --action=trim to trim adapters and up/downstream sequence
         */
         '''
         output="!{sample}_cutadapt.fastq.gz"
         log="!{sample}_cutadapt_log.txt"
-        par="-b file:!{adapters} -B file:!{adapters} -j !{task.cpus} -m 20 -e 0.33 --action=trim --interleaved"
+        par="-b file:!{adapters} -B file:!{adapters} -j !{task.cpus} -m 20 -e 0.125 --action=trim --interleaved"
         zcat !{reads_interleaved} | cutadapt ${par} - 2> ${log} | gzip -c > ${output}
         ln -s !{reads_interleaved} !{sample}_cutadapt_in.fastq.gz
         '''


willbradshaw:
@katherine-stansifer I can't currently merge this as one of the tests is failing; this isn't surprising as this change will change the output of the viral identification workflow, but needs to be resolved before we merge to dev.

To resolve this, can you:
1. Run the run output test locally
2. Verify that the changes match what you'd expect (most of the files should be unchanged, those that are changed should be restricted to the viral workflow, and those should be changed in sensible ways), and write up the changes in a comment here
4. Generate a new nf-test snapshot
5. Copy the new output to `test-data/gold-standard-results` (this lets us view the changes in a diff)
6. Make a new commit for review

For step 2, you can verify that most files are unchanged by comparing their md5 sums, then look into those that are changed by comparing the test output with the files in `test-data/gold-standard-results`.

willbradshaw:
Oh, also run the CUTADAPT process test locally (`nf-test test tests/modules/local/cutadapt`) and confirm that that still passes.

katherine-stansifer:
Checking for relevant tests:
the string CUTADAPT in found in 
./modules/local/cutadapt/main.nf (obviously)
./modules/local/fastp/main.nf (but this is just in a comment; FASTP process doesn't run CUTADAPT)
./subworkflows/local/extractViralReads/main.nf
./tests/modules/local/cutadapt/main.nf.test

So, it looks like we need to run the cutadapt test and also any tests for extractViralReads

CUTADAPT test passes
`%nf-test test tests/modules/local/cutadapt
Test process CUTADAPT

Test [eccfab17] 'Should run without failures and conserve reads'     
PASSED (22.97s)
`
Similarly, EXTRACT_VIRAL_READS test passes (`nf-test test tests/subworkflows/local/extractViralReads/`)

I don't see any local tests for the entire RUN workflow (which of course sources EXTRACT_VIRAL_READS), but I'm still getting my head around tests so let me know if there are others I should run!

willbradshaw:
Thanks @katherine-stansifer! The tests for the RUN workflow are in `tests/workflows/run.nf.test`. The main thing there is to check and update the snapshot on the run_output test so it doesn't fail and block the merge; this is a bit fiddly so I think it's probably easiest if @harmonbhasin or I walk you through it tomorrow.

katherine-stansifer:
cutadapt and extractviralreads local tests still pass! ready for re-review.

willbradshaw (dismissed):

willbradshaw:
@katherine-stansifer two last things on this:

1. You need to merge in other changes from dev before we can merge this
2. Can you add an update to the CHANGELOG about this change? This shouldn't affect any results, so it can be a [point release](https://github.com/naobservatory/mgs-workflow/blob/master/docs/versioning.md), something like 

```
# v2.8.1.1-dev
- Made Cutadapt mismatch rate parameter configurable
```

willbradshaw (approved):

PR #211, opened 2025-03-04
Updating index.

harmonbhasin:
Corrected changes to index file, and reran index for running the run test, then updated the snapshot.


harmonbhasin:
Flagging that this change has broken two tests:

- `tests/modules/local/joinRiboRef/main.nf.test`
- `tests/modules/local/minimap2/align.nf.test`

These will be addressed before merging dev to master.

willbradshaw (approved):

PR #214, opened 2025-03-06
Fixing some ont tests

harmonbhasin:
This was partially broken due to our earlier pull request to get index compatibility working, but it turns out one of the containers for the ont minimap workflow didn't exist so I created it with seqera.


willbradshaw:
Actually, these test failures are interfering with my own work today, so I'm going to merge this. @simonleandergrimm, have a look when you can and we can always change it later.

willbradshaw (approved):

PR #215, opened 2025-03-06
Fixing some tests that fail

harmonbhasin:
Ran all the tests and these were the only ones that failed.


willbradshaw (approved):

tests/subworkflows/local/loadSampleSheet/main.nf.test
```diff
@@ -69,7 +69,7 @@ nextflow_workflow {
         then {
             assert workflow.failed
             assert workflow.exitStatus == 1
-            assert workflow.stdout[0].contains("Invalid samplesheet header")
```
willbradshaw, line 72: On reflection, I'm nervous about this change. Are you sure this won't happen if the workflow fails for a different, unrelated reason? This would happen if Nextflow prints out the whole text of the workflow on a failure, for example.

willbradshaw: Also, the current version passes on my machine, which is weird.

willbradshaw (changes requested):

tests/subworkflows/local/loadSampleSheet/main.nf.test
harmonbhasin: Summary from chatting in person: I'll do the above, and it is weird that the verisons differ in passing, maybe due to nf-test versions.

harmonbhasin: I just discovered why these tests fail for me, but not for you. I usually run tests with `--verbose` at the end so that I can see what's going on. This ends up adding whatever output nextflow has to stdout. However, I printed out workflow.stdout and it only prints out the error, so this current implementation works! The current implementation will also work with verbose.

PR #227, opened 2025-03-11
Modify Kraken2 DB handling in index workflow to avoid staging

willbradshaw:


willbradshaw:
I wasn't able to come up with a good way to test the GET_TARBALL process in a useful way. I'm open to suggestions.

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# v2.8.1.1
+- Modified Kraken2 DB handling in index workflow to avoid staging
```
harmonbhasin, line 2: I would also add that you're updating some of the versioning for the other references as well (SILVA and taxonomy).

harmonbhasin (approved):
Left a small comment, but it's up to you whether you want to implement it or not.

CHANGELOG.md
willbradshaw: 👍

PR #229, opened 2025-03-12
Fix FILTER_BLAST behavior

willbradshaw:
- Changed FILTER_BLAST to filter conjunctively on bitscore rank and bitscore fraction, rather than disjunctively.
- Fixed Bitscore ranking in FILTER_BLAST to correctly give tied bitscores tied ranks
- Updated default thresholds

CHANGELONG updated, local FILTER_BLAST test passes.


harmonbhasin:
The run output test is failing, I'll take a look once that passes.

willbradshaw:
Updated snapshot; changes to BLAST output match expectations

configs/run.config
```diff
@@ -29,7 +29,7 @@ params {
     blast_db_prefix = "core_nt"
     blast_viral_fraction = 0 // Fraction of putative host-infecting virus reads to BLAST vs nt (0 = don't run BLAST)
     blast_min_frac = 0.9 // Keep BLAST hits whose bitscore is at least this fraction of the best bitscore for that query
-    blast_max_rank = 5 // Keep BLAST hits whose dense bitscore rank for that query is at most this value
+    blast_max_rank = 10 // Keep BLAST hits whose dense bitscore rank for that query is at most this value
```
harmonbhasin, line 32: For my understanding (and anyone else in the future who looks back at this commit), why are we increasing the score from 5 to 10?

modules/local/filterBlast/resources/usr/bin/filter_blast.py
```diff
@@ -70,14 +70,20 @@ def filter_blast(input_path, output_path, max_rank, min_frac, query_index, bitsc
                 bitscore_max = bitscore_new
                 query_last = query_new
             else:
-                # Existing query: compare bitscore
-                bitscore_frac = bitscore_new / bitscore_max
-                if bitscore_frac > min_frac:
+                # Existing query: compare bitscore and filter on rank & frac
```
harmonbhasin, line 73: Briefly explain why we're changing this logic/why it was incorrect before (if it was incorrect).

harmonbhasin (approved):

PR #232, opened 2025-03-12
Automatically testing a changed process/subworkflow and its dependencies

simonleandergrimm:
0th draft. See thread for context: https://twist.com/a/197793/inbox/t/6808534/


harmonbhasin (approved):
LGTM!

harmonbhasin (changes requested):
Changes this to 'Request changes' for the CHANGELOG update.

simonleandergrimm:
@harmonbhasin Can you assign to Will once you're happy with this?

simonleandergrimm:
I just noticed that we don't check the workflow runs anymore as part of the Github-side tests. Will add those back in.

simonleandergrimm:
Just noting that this is in my backlog, will get to it.

simonleandergrimm:
Just make sure to assign it to me when you're ready for me to look at it again.
Yeah, will do!

simonleandergrimm:
@harmonbhasin Updated the script and reassigned it to you. Also, do you think the script should go under bin/?

harmonbhasin:
@simonleandergrimm yes i think putting it under `bin/` would be good.

harmonbhasin (dismissed):
Given that you've confirmed that this script works as intended locally, this looks good to me.

simonleandergrimm:
@harmonbhasin Can you run it locally too?

harmonbhasin:
@simonleandergrimm yes let me try this now

harmonbhasin:
@simonleandergrimm this worked well for me, just move the script into the bin folder, and you're good to merge.

willbradshaw:
@harmonbhasin @simonleandergrimm noting that I want to review this (and all PRs) before merging into dev. Let me know when it's ready for me to do that.

harmonbhasin:
@willbradshaw sounds good, this is ready for you to look at.

bin/test_component_dependencies.py
```diff
@@ -0,0 +1,160 @@
...
+
+def main():
+    component, verbose = parse_args()
+
+    tests_to_execute = set()
+
+    direct_tests = find_dependency("tests/", component)
+    tests_to_execute.update(direct_tests)
+
+    print("=" * 80)
```
willbradshaw, line 82: This should probably be 72: https://peps.python.org/pep-0008/#maximum-line-length

bin/test_component_dependencies.py
```diff
@@ -0,0 +1,160 @@
...
+    print(f"Found {len(direct_tests)} direct tests:")
+    print("=" * 80)
+    for test in direct_tests:
+        print(f"   • {test}")
+    print()
+    # print("\n")
```
willbradshaw, line 88: Delete or uncomment

bin/test_component_dependencies.py
```diff
@@ -0,0 +1,160 @@
...
+        default=False,
+    )
+
+    args = parser.parse_args()
+    component = args.component
+    verbose = args.verbose
+    return component, verbose
+
+
+def main():
```
willbradshaw, line 74: This function needs comments

bin/test_component_dependencies.py
```diff
@@ -0,0 +1,160 @@
...
+
+    tests_to_execute.update(subworkflow_tests)
+    tests_to_execute.update(workflow_tests)
+
+    print("=" * 80)
+    print(f"Identified {len(tests_to_execute)} tests to execute. Running...")
+
+    test_results = {}
+    for test in sorted(tests_to_execute):
+        args = ["nf-test", "test", test]
```
willbradshaw, line 126: I'm worried that running a separate nf-test command for every test will be substantially slower than a single command (or a small number of commands) passed multiple tests. I'm not confident about that but I think it's worth checking before we commit to this approach.

OTOH this format allows us to add in parallelisation later, which is great.

willbradshaw: > OTOH this format allows us to add in parallelisation later, which is great.

Speaking of which, @harmonbhasin we should integrate this with the functionality from your https://github.com/naobservatory/mgs-workflow/blob/master/bin/run_parallel_test.sh. No point having two competing test handlers, it will just confuse people.

bin/test_component_dependencies.py
```diff
@@ -0,0 +1,160 @@
...
+        if verbose:
+            args.append("--verbose")
+        print(f"\nRunning \"{' '.join(args)}\"")
+        print("-" * 80)
+        result = subprocess.run(args)
```
willbradshaw, line 131: @harmonbhasin and I went back and forth a fair bit about how test handlers should deal with output from tests (pass to stderr? write to a file? etc) for run_parallel_test.sh, I recommend taking the same approach here.

willbradshaw: Also relevant: `execute_single` method from https://github.com/naobservatory/mgs-orchestrator/blob/main/orchestrator/pipeline_executor.py

willbradshaw: (Though in this case I probably recommend writing to a file rather than passing through directly, or maybe capturing and then writing the output for failed tests at the end)

bin/test_component_dependencies.py
```diff
@@ -0,0 +1,160 @@
...
+        subworkflow_tests.update(tests)
+
+    print("=" * 80)
+    print(f"Found {len(subworkflow_tests)} dependent subworkflow tests:")
+    print("=" * 80)
+    for test in subworkflow_tests:
+        print(f"   • {test}")
+    print()
+
+    workflow_tests = set()
```
willbradshaw, line 105: Note that workflows can take modules as dependencies, not just subworkflows, so need to check those as well

bin/test_component_dependencies.py
```diff
@@ -0,0 +1,160 @@
...
+def parse_args():
+    """Parse arguments and run the test finder."""
+    parser = argparse.ArgumentParser(
+        description="Find and run tests for specified components"
+    )
+    parser.add_argument(
+        "-c",
+        "--component",
+        required=True,
+        help="Name of component (process or subworkflow) to test",
```
willbradshaw, line 57: Is this intended to be a name or a filepath? It says name, but in some functions above it gets treated like a filepath.

I'd recommend a name, e.g. `EXTRACT_VIRAL_READS`, rather than a path, e.g. `subworkflows/local/extractViralReads/main.nf`

If you go with a name you can enforce it, e.g. by requiring it to only include alphanumerics and underscores.

bin/test_component_dependencies.py
```diff
@@ -0,0 +1,160 @@
...
+import os
+
+
+def find_dependency(directory, component):
+
+    directory = directory.rstrip("/")
+    component = component.replace("/main.nf", "")
+
+    grep_results = subprocess.run(
+        ["grep", "-r", component, directory], capture_output=True, text=True
```
willbradshaw, line 13: Just grepping like this will fail in cases where one component is a substring of another, e.g. RUN vs RUN_VALIDATION.

bin/test_component_dependencies.py
```diff
@@ -0,0 +1,160 @@
...
+
+def workflow_uses_subworkflow(workflow, subworkflow_path):
+
+    workflow_path = f"workflows/{workflow}"
+    subworkflow_dir = subworkflow_path.replace("/main.nf", "")
+
+    with open(workflow_path, "r") as f:
+        workflow_content = f.read()
+
+    if subworkflow_dir in workflow_content:
```
willbradshaw, line 33: Here also you're vulnerable to substring matches

bin/test_component_dependencies.py
willbradshaw: I'd suggest only looking in specific lines, e.g. `include` statements and setup-block `run` statements.

Oh, and nf-test `workflow` / `process` declarations.

willbradshaw (changes requested):

bin/test_component_dependencies.py
simonleandergrimm: Good point yeah, it should be something like EXTRACT_VIRAL_READS.

bin/test_component_dependencies.py
simonleandergrimm: @harmonbhasin can you take that on or shall I.

bin/test_component_dependencies.py
simonleandergrimm: I fixed this with word boundaries, let me know if this still addresses your concern.

bin/test_component_dependencies.py
simonleandergrimm: See my longer comment below.

simonleandergrimm:
@willbradshaw for your concerns regarding substring matching, instead of what you suggested I implement grepping/regex that uses word boundaries.

E.g.: with the faulty subworkflow `subworkflows/local/profi/main.nf`, the original code would have returned True:

```{test-script.py}
#! /usr/bin/env python3
import re

workflow = "run.nf"
subworkflow_path = "subworkflows/local/profi/main.nf"
workflow_path = f"workflows/{workflow}"
subworkflow_dir = subworkflow_path.replace("/main.nf", "")

with open(workflow_path, "r") as f:
    workflow_content = f.read()

# Use word boundaries to preempt substring matches
if subworkflow_dir in workflow_content:
    print("True")
else:
    print("False")
```
```
[ec2-user@ip-172-31-40-150 mgs-workflow]$ ./test-script.py 
True
```

However, with word boundaries, this is fixed:

```{test-script-boundary.py}
#! /usr/bin/env python3
import re

workflow = "run.nf"
subworkflow_path = "subworkflows/local/profi/main.nf"
workflow_path = f"workflows/{workflow}"
subworkflow_dir = subworkflow_path.replace("/main.nf", "")

with open(workflow_path, "r") as f:
    workflow_content = f.read()

# Use word boundaries to preempt substring matches
if re.search(r'\b' + re.escape(subworkflow_dir) + r'\b', workflow_content):
    print("True")
else:
    print("False")
```

```
[ec2-user@ip-172-31-40-150 mgs-workflow]$ ./test-script-boundary.py 
False
```


I did the same for grepping the component names.

Now when using the subcomponent name "FAST" I get no tests returned:

```
[ec2-user@ip-172-31-40-150 mgs-workflow]$ bin/test_component_dependencies.py -c FAST
{.....}

----------------------------------------
Total:  0 tests
Passed: 0 tests
Failed: 0 tests
----------------------------------------
```

Whereas previously we did:

```
[ec2-user@ip-172-31-40-150 mgs-workflow]$ bin/test_component_dependencies.py -c FAST
{....}
Identified 38 tests to execute. Running...
{....}
```

bin/test_component_dependencies.py
simonleandergrimm: Added something to this effect:
```
[ec2-user@ip-172-31-40-150 mgs-workflow]$ bin/test_component_dependencies.py -c FASTP/
Error: Input component 'FASTP/' must be a process or subworkflow name, not the file path that contains said process/subworkflow. E.g., 'FASTP' not 'modules/local/fastp/main.nf'
```

bin/test_component_dependencies.py
simonleandergrimm: Added

bin/test_component_dependencies.py
simonleandergrimm: Having looked into this, I'm unsure if nf-test --shard is the right approach here? We could just use Python's multiprocessing

simonleandergrimm: Given that we are passings tests individually, not as dirs.

bin/test_component_dependencies.py
simonleandergrimm: If we implement parallelisation, putting the stdout in log, with stderr printed to terminal upon completion (and keeping the summary I give with which tests worked vs which didn't).

bin/test_component_dependencies.py
simonleandergrimm: @harmonbhasin Might be most efficient to briefly chat about this on Tuesday.

bin/test_component_dependencies.py
simonleandergrimm: Done

bin/test_component_dependencies.py
```diff
@@ -0,0 +1,177 @@
...
+    verbose = args.verbose
+    return component, verbose
+
+
+def main():
+    component, verbose = parse_args()
+
+    # Validate component name - only allow alphanumerics and underscores
+    if not re.match(r'^[a-zA-Z0-9_]+$', component):
+        print(f"Error: Input component '{component}' must be a process or subworkflow name, not the file path that contains said process/subworkflow. E.g., 'FASTP' not 'modules/local/fastp/main.nf'")
```
willbradshaw, line 81: I think this is good to have but the message is a bit too specific for the condition that triggers it. E.g. if someone specified "ABC+" they'd be told "you can't use a path here", but they didn't.

Also, you should either use an `assert` or a `raise` here, not a print + exit.

bin/test_component_dependencies.py
willbradshaw: I agree word boundaries solve the narrow problem I was pointing to here, but there are other ways this check is overbroad, e.g. what if a process is mentioned in a comment in another process? That would be enough to trigger this even if the process isn't actually called anywhere in the file.

bin/test_component_dependencies.py
willbradshaw: Yeah, if you're calling each test file (I believe you're passing test files, not individual tests?) separately, I don't see much advantage of nf-test --shard over another parallelisation approach.

bin/test_component_dependencies.py
willbradshaw: Even without parallelisation, I think you need to think about how best to capture, aggregate and return the results of all these subprocess calls.

willbradshaw (changes requested):

willbradshaw:
My main concern about not implementing parallelisation now is this earlier comment:

> I'm worried that running a separate nf-test command for every test will be substantially slower than a single command (or a small number of commands) passed multiple tests. I'm not confident about that but I think it's worth checking before we commit to this approach.

bin/test_component_dependencies.py
simonleandergrimm: FWIW, I already do that (lmk if you don't like how it's currently looking).

bin/test_component_dependencies.py
simonleandergrimm: Made both fixes.

bin/test_component_dependencies.py
simonleandergrimm: Good point. I added further logic that should stop that from happening.

simonleandergrimm: Though I didn't know what you meant by `run` statements.

simonleandergrimm:
> My main concern about not implementing parallelisation now is this earlier comment:
> 
> I'm worried that running a separate nf-test command for every test will be substantially slower than a single command (or a small number of commands) passed multiple tests. I'm not confident about that but I think it's worth checking before we commit to this approach.
> 
> If that turns out not to be an issue I agree we don't lose much by just merging this in serial mode and then adding parallel processing later.

Running 6 tests with launching nf-test individually added 23 seconds, so around 4 seconds per test. Seems fine?
```
Approach 1 (All at once):
  - Total time: 0:02:05.100760
  - Status: SUCCESS

Approach 2 (Individual runs):
  - Total time: 0:02:28.822050
  - Successful tests: 6/6

Time difference:
  - Running all tests at once was 0:00:23.721289 faster
  - Improvement: 18.96%
```

willbradshaw:
That doesn't seem fine to me! A ~20% increase for just six tests, with the expectation that it will get worse with more tests run, is pretty bad. We care quite a bit about performance for testing.

The alternative, which is less amenable to parallel processing but better for serial processing, is to just collate a bunch of paths in a list and then pass them all to nf-test in a single call. That seems like it should be pretty easy to implement.

simonleandergrimm:
Alright, I've now done the thing of collating paths into a list and passing that to nf-test. The performance for that is fine:
```
================================================================================
SUMMARY
================================================================================
Approach 1 (All at once with directory):
  - Total time: 0:01:49.218480
  - Status: SUCCESS

Approach 2 (All at once with explicit test files):
  - Total time: 0:01:48.840106
  - Status: SUCCESS

Time difference:
  - Running with explicit files was 0:00:00.378374 faster
  - Improvement: 0.35%
 ```

**Comment by simonleandergrimm** [comment]
@willbradshaw This is ready for re-review. Additionally, it'd be good if you ran the script on your side to spot any final improvements you'd like to be made.

File: bin/test_component_dependencies.py (hunk 8 above)
Comment (willbradshaw) [reply bin/test_component_dependencies.py]: By `run` statements, I mean statements like the following within `setup` blocks in tests:

```
setup {
    run("LOAD_DOWNSTREAM_DATA") {
        // ...
    }
}
```

File: bin/test_component_dependencies.py [hunk 11]
```diff
@@ -0,0 +1,166 @@
...
+        print(f"   • {test}")
+    print()
+
+    tests_to_execute.update(subworkflow_tests)
+    tests_to_execute.update(workflow_tests)
+
+    print("=" * 72)
+    print(f"Identified {len(tests_to_execute)} tests to execute. Running...")
+
+    test_results = {}
```
willbradshaw, line 149: I think this now isn't doing anything?

bin/test_component_dependencies.py
willbradshaw: This is now moot since we're just making one big call to nf-test, so I'll resolve this for now, but we'll need to come back to this when we introduce parallelisation.

willbradshaw:
OK, I tried running this on my own instance:

```
ec2-user@ip-172-31-75-190:~/mgs-workflow/repo$ ./bin/test_component_dependencies.py -c BBMERGE
========================================================================
Found 1 direct tests:
========================================================================
   • tests/modules/local/bbmerge/main.nf.test

========================================================================
Found 2 dependent subworkflow tests:
========================================================================
   • tests/subworkflows/local/taxonomy/main.nf.test
   • tests/modules/local/processKrakenViral/main.nf.test

========================================================================
Found 0 dependent workflow tests:
========================================================================

========================================================================
Identified 3 tests to execute. Running...
// Etc
```

This fundamentally works, but I have some notes:

1. I don't know why it's running `tests/modules/local/processKrakenViral/main.nf.test`; is it because it calls the taxonomy subworkflow in the setup method? And why is it calling it a subworkflow test when it's a module? Probably some clearer logging would be helpful here.
2. I think this should run the workflow that depends on the dependent subworkflows (in this case, RUN) to check that it still works in context.

simonleandergrimm:
> I don't know why it's running tests/modules/local/processKrakenViral/main.nf.test; is it because it calls the taxonomy subworkflow in the setup method? And why is it calling it a subworkflow test when it's a module? Probably some clearer logging would be helpful here.

Thanks for testing! This is because the logging comment was wrong: We are executing tests that _use_ that subworkflow, which can include tests of modules.

> I think this should run the workflow that depends on the dependent subworkflows (in this case, RUN) to check that it still works in context.

Yup, I agree, this misses that some subworkflows that are affected are used by another susbworkflow, which gets missed by `workflow_uses_subworkflow()`.

I added code that fixes this. Most important logic:

```
...
    # Find subworkflows that directly use the component
    dependent_subworkflows = set(find_dependency("subworkflows/", component))

    # Find transitive dependencies - subworkflows that use other dependent subworkflows
    transitive_dependencies = set()
    for subworkflow in os.listdir("subworkflows/local/"):
        for dependent_subworkflow in dependent_subworkflows:
            if subworkflow_uses_subworkflow(subworkflow, dependent_subworkflow):
                transitive_dependencies.add(subworkflow)

    # Combine all dependent subworkflows
    dependent_subworkflows.update(transitive_dependencies)

    # Identifying tests that use the affected subworkflows
    subworkflow_tests = set()
    for dependent_subworkflow in dependent_subworkflows:
        tests = find_dependency("tests/subworkflows/", dependent_subworkflow)
        subworkflow_tests.update(tests)
```

bin/test_component_dependencies.py
willbradshaw: @simonleandergrimm ping on this comment

willbradshaw:
Hmm, now I'm getting a weird error when I try to run the script. It works at [3fe9878](https://github.com/naobservatory/mgs-workflow/pull/232/commits/3fe98784c04fd77e8cf7632a6e36bd9afe6bddb4), but breaks at [daddb29](https://github.com/naobservatory/mgs-workflow/pull/232/commits/daddb29f3a3e488d6ef696d75e91f430d2c6baff) or later.

willbradshaw:
OK I figured out the problem on my end and fixed it, now it works, here's another example output:

```
ec2-user@ip-172-31-75-190:~/mgs-workflow/repo$ ./bin/test_component_dependencies.py -c SUBSET_FASTQ
========================================================================
Found 2 direct tests:
========================================================================
   • tests/modules/local/.nf-test.log
   • tests/modules/local/subsetFastq/main.nf.test

========================================================================
Found 1 tests for affected subworkflows:
========================================================================
   • tests/subworkflows/local/blastViral/main.nf.test

========================================================================
Found 2 tests for affected workflows:
========================================================================
   • tests/workflows/run_validation.nf.test
   • tests/workflows/run.nf.test

========================================================================
Identified 5 test files to execute. Running...

Running test files:
   • tests/modules/local/.nf-test.log
   • tests/modules/local/subsetFastq/main.nf.test
   • tests/subworkflows/local/blastViral/main.nf.test
   • tests/workflows/run.nf.test
   • tests/workflows/run_validation.nf.test
------------------------------------------------------------------------
...
```

Looking very nice! It shouldn't be pulling in `.nf-test` files, though.

simonleandergrimm:
Added logic to drop those files:

```
    file_paths = set()
    for line in grep_results.stdout.splitlines():
        if any(trigger in line for trigger in relevant_triggers):
                file = line.split(":", 1)[0]
                if "nf-test" in file:
                    continue
                else:
                    file_paths.add(file)
```

willbradshaw:
OK great, this looks good to me now! @simonleandergrimm can you update the CHANGELOG? Then we should be good to merge.

simonleandergrimm:
CHANGELOG is updated.

willbradshaw:
@simonleandergrimm now you need to merge in dev and fix the conflicts.

willbradshaw (approved):

simonleandergrimm:
Done!

willbradshaw (approved):

willbradshaw:
@simonleandergrimm assuming the tests all pass, you can merge this!

willbradshaw:
Nice work, I'm excited to have this functionality.

PR #235, opened 2025-03-14
Added code that makes generate_samplesheet.py take in ONT barcodes.

simonleandergrimm:
ONT demultiplexing sometimes matches reads to barcodes that do not actually exist in the library. We don't want to analyze those, as the files are often very small and can make the pipeline trip up. This will also make sure we do not process unclassified reads, which are often just noise.

### DIFF
--- a/bin/generate_samplesheet.py
+++ b/bin/generate_samplesheet.py
@@ -28,14 +28,18 @@ def main():
                         help="Optional CSV with columns 'sample,group' to merge grouping info")
     parser.add_argument("--group_across_illumina_lanes", action="store_true",
                         help="Group samples across Illumina lanes by removing '_Lnnn' from sample name")
-    
-    args = parser.parse_args()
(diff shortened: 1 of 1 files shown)


jeffkaufman:
This is ok, though I think the best way to do it would be to merge them into 'unclassified' at basecalling time.

harmonbhasin (changes requested):
LGTM, I would make sure to consider Jeff's comment. Before we can merge, you should update the changelog, thanks!

jeffkaufman:
> This is ok, though I think the best way to do it would be to merge them into 'unclassified' at basecalling time.

Thinking more, I'm really not that excited about the approach of doing this when creating samplesheets.  These are reads that couldn't be demultiplexed to a known barcode, and should be in the "unclassified" sample with all the other such reads.

simonleandergrimm:
Sounds good, will implement it during basecalling time.

PR #236, opened 2025-03-14
Implement two-way version compatibility checking between pipeline and index

willbradshaw:


willbradshaw:
All workflow tests pass locally, as do the new tests for CHECK_VERSION_COMPATIBILITY subworkflow.

willbradshaw:
Test failures seem to be permission issues, which is weird since the new test index is in the same bucket as the old one. I'll rerun them when the index test is done, then if they fail again will look into it in more detail.

harmonbhasin (dismissed):
LGTM!

harmonbhasin:
> Test failures seem to be permission issues, which is weird since the new test index is in the same bucket as the old one. I'll rerun them when the index test is done, then if they fail again will look into it in more detail.

I discussed this with Will, but for future reference (this will go in our developer docs), anytime someone generates a new test index, they must make it public in s3.

harmonbhasin (approved):
LGTM!

PR #244, opened 2025-03-20
Making BLAST parameters qcov_hsp_perc and perc_identity configurable

simonleandergrimm:
**Changes:**
- Added ability to set BLAST parameters qcov_hsp_perc and perc_identity

**Tests run**
 - tests/workflows
 - tests/subworkflows/local/blastViral/main.nf.test
 - tests/subworkflows/local/extractViralReadsONT/main.nf.test
 - tests/modules/local/blast/main.nf.test


subworkflows/local/extractViralReadsONT/main.nf
```diff
@@ -10,7 +10,6 @@ include { ADD_SAMPLE_COLUMN as LABEL_HV_TSVS } from "../../../modules/local/addS
 include { FILTLONG } from "../../../modules/local/filtlong"
 include { MASK_FASTQ_READS } from "../../../modules/local/maskRead"
 include { PROCESS_VIRAL_MINIMAP2_SAM } from "../../../modules/local/processViralMinimap2Sam"
-include { BLAST_VIRAL } from "../../../subworkflows/local/blastViral"
```
simonleandergrimm, line 13: This was an unused import I now deleted.

tests/subworkflows/local/blastViral/main.nf.test
```diff
@@ -140,6 +146,106 @@ nextflow_workflow {
...
+                    input[1] = "input.fastq.gz"
+                    '''
+                }
+            }
+        }
+        when {
+            params {
+                max_rank = 5
+                min_frac = 0.9
+                blast_frac = 0.004 // ONT test file is large (241k reads).
```
willbradshaw, line 234: We shouldn't be using files that large for local testing. Can you make a new test file with fewer reads? (as few as possible, ideally 250 or less)

configs/run.config
```diff
@@ -30,6 +30,8 @@ params {
     blast_viral_fraction = 0 // Fraction of putative host-infecting virus reads to BLAST vs nt (0 = don't run BLAST)
     blast_min_frac = 0.9 // Keep BLAST hits whose bitscore is at least this fraction of the best bitscore for that query
     blast_max_rank = 10 // Keep BLAST hits whose dense bitscore rank for that query is at most this value
+    blast_perc_id = 60 // Percent identity threshold for BLAST hits
+    blast_qcov_hsp_perc = 30 // Query coverage threshold for BLAST hits
```
willbradshaw, line 34: This doesn't apply until the main RUN workflow can handle ONT data, but: To play nicely with mgs-orchestrator, we need to make these defaults responsive to the value of `ont` (otherwise it'll run all runs with the Illumina defaults).

willbradshaw (changes requested):

willbradshaw:
Overall, assuming all the local tests you ran pass, this looks good!

tests/subworkflows/local/blastViral/main.nf.test
simonleandergrimm: Created a new file with 250 reads. Now re-running all tests that use ONT data.

willbradshaw: The only thing to be careful about is that the subset needs to have some HV reads in it to make it through the filters.

simonleandergrimm:
After editing ont-samplesheet.csv I reran the following tests, which all passed.

 - tests/workflows/run_dev.nf.test --tag run_dev_ont
 - tests/modules/local/minimap2/align.nf.test
 - tests/modules/local/minimap2/align_non_streamed.nf.test
 - tests/modules/local/processViralMinimap2Sam/main.nf.test
 - tests/subworkflows/local/blastViral/main.nf.test 
 - tests/subworkflows/local/extractViralReadsONT/main.nf.test
 - tests/configs/run_dev_ont.config

simonleandergrimm:
test-index timed out, rerunning.

configs/run.config
simonleandergrimm: Just to make sure, is this something that needs fixing now or no?

willbradshaw (approved):

PR #246, opened 2025-03-25
Changing Github Actions to use secrets so that we can make our testing directories pay-to-use.

harmonbhasin:
Currently have nao-testing set to `requester pays` so if these test pass then this is working.

### DIFF
--- a/.github/workflows/end-to-end.yml
+++ b/.github/workflows/end-to-end.yml
@@ -12,6 +12,12 @@ jobs:
     timeout-minutes: 10
 
     steps:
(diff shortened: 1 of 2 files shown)


willbradshaw (approved):

willbradshaw:
This looks good and can be merged. We should add some documentation about how someone else can set up the tests to work, but that can be in another PR.

harmonbhasin:
> This looks good and can be merged. We should add some documentation about how someone else can set up the tests to work, but that can be in another PR.

Sounds good!

PR #247, opened 2025-03-25
Fixing bug in BLAST validation

harmonbhasin:


harmonbhasin:
Need to write a test to catch this locally.

willbradshaw:
What was going wrong before?

harmonbhasin:
@willbradshaw Running blast validation with just the fastq file wasn't working before. Pretty sure the reason was because we only have tests for passing in the viral hits table.

willbradshaw:
OK, yeah, good to get a test then. Make sure to merge in dev first.

tests/configs/run_validation.config
```diff
@@ -9,10 +9,6 @@ params {
     base_dir = "./" // Parent for working and output directories (can be S3)
     ref_dir = "s3://nao-testing/index/20250314/output/" // Reference/index directory (generated by index workflow)
 
-    // Files
```
willbradshaw, line 12: Why cut this?

willbradshaw:
@harmonbhasin can you confirm which local tests you ran and that they all pass?

tests/configs/run_validation.config
harmonbhasin: As we're only using that config for the validation test, so I figured I'd just only include those parameters there not to lead to any confusion. I can reintroduce it if you'd like.

harmonbhasin:
@willbradshaw I ran all local tests that used `run_validation.config`, which was only `tests/workflow/validation.nf.test`, and both of those tests passed.

willbradshaw (changes requested):

willbradshaw:
Other than the one outstanding comment thread, this looks good. I'll merge it once that's resolved, then work on merging to master.

tests/configs/run_validation.config
harmonbhasin: Sounds good, this makes sense, I've reintroduced them, but have left them with empty strings.

harmonbhasin:
@willbradshaw sounds good! I ran the relevant local tests again and they've passed.

PR #250, opened 2025-03-26
Merge v2.8.2.0 to master

willbradshaw:
All local tests passing, small composite dataset ran through without errors.

I haven't made a new index, since nothing in the CHANGELOG touches on that and the index from the previous version continues to work fine.


harmonbhasin (approved):
LGTM!

PR #252, opened 2025-03-27
Lowered default bracken_threshold to avoid run failures

willbradshaw:


willbradshaw:
Annoyingly this is a results bump since the new default will affect taxonomic profiling results (though I expect only slightly).

harmonbhasin (approved):
LGTM!

Issue #248, opened 2025-03-25
Setting memory param in FASTQC (so it doesn't trip over long ONT reads)


simonleandergrimm:
Some raw ONT reads can be very long (100k+); they are most often junk. Nevertheless, when running QC on them, they can trip up FASTQC, giving OOM errors. 

FASTQC has hardcoded memory per thread (250 MB per processing unit), so even though we allocate 4GB to FASTQC, that's not actually used. Memory can be increased by setting the --memory flag (in MB):

```
%  ../FastQC/fastqc -h
...
    --memory        Sets the base amount of memory, in Megabytes, used to process 
                    each file.  Defaults to 512MB.  You may need to increase this if
                    you have a file with very long sequences in it.
...
```

This should be a quick fix, given that we do allocate enough memory to the overall process.

willbradshaw:
@katherine-stansifer did you mean to reopen this?

katherine-stansifer:
(but it is fully addressed in dev)

PR #264, opened 2025-04-03
Katherine fastqc memory

katherine-stansifer:
Main goal of this PR is to address https://github.com/naobservatory/mgs-workflow/issues/248 (OOM errors for ONT on fastqc)

At Simon's suggestion, I fixed this by adding the --memory parameter to fastqc, so it uses all 4GB allocated to it rather than using the usual baked-in 512MB limit.

I made an artificial ONT input file (pulled a random PHC ONT file and appended a randomly generated 2 million bp read to the end of it; I verified that I get an OOM error with the old code and that my tweak has fixed this).

I also did some basic cleanup: removed fastqc resource requirements, added a "does it run" test for fastqc, and removed unused processes. 

Searched for tests that touch FASTQC and ran them:
*tests/modules/local/fastqc
*tests/modules/local/summarizeMultiqc
*tests/subworkflows/local/QC
*tests/subworkflows/local/runQc

All run fine!


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# v2.8.3.2
```
willbradshaw, line 1: Could you make this 2.8.3.2-dev? (here and in pipeline-version.txt)
I like to be able to distinguish between pipeline outputs from a dev version vs a stable version on master.

pipeline-version.txt
```diff
@@ -1 +1 @@
-2.8.3.1
+2.8.3.2
```
willbradshaw, line 1: As above, can you append "-dev" until we're ready to merge into master?

willbradshaw (changes requested):

willbradshaw:
Code looks good, requested a couple of very small book-keeping changes.

willbradshaw (approved):

PR #270, opened 2025-04-04
Add rank-raised taxids to viral taxonomy DB

willbradshaw:
This PR modifies the MAKE_VIRUS_TAXONOMY_DB subworkflow to add rank-raised taxids to each entry in the viral taxonomy db (with column names `taxid_species`, `taxid_genus` etc).

These are currently not used directly in the pipeline itself, so index/pipeline backwards compatibility is preserved in both directions (though I intend to use them in the DOWNSTREAM workflow in an upcoming PR).

I ran tests for the following locally, all passing:
- INDEX
- MAKE_VIRUS_TAXONOMY_DB
- ANNOTATE_VIRUS_INFECTION
- RAISE_TAXONOMY_RANKS


harmonbhasin (approved):
LGTM!

PR #278, opened 2025-04-08
Prepare samples for post-hoc validation in DOWNSTREAM workflow

willbradshaw:
Part 1 of a series of PRs implementing post-hoc validation of putative viral reads in the DOWNSTREAM workflow.

The code in this PR:

1. Joins the viral taxonomy DB in the index to the viral hit TSV (to get higher-level taxids)
2. Partitions the hits TSV by species (into a list of single-species TSVs)
3. Rearranges the channel into a list of (label, TSV) 2-tuples (needed for downstream processes)
4. Extracts sequences from each TSV into interleaved FASTQ
5. Merges read pairs, then joins unmerged read pairs, to get a FASTQ with a single sequence per input pair (required for downstream clustering)

The next steps (not yet implemented) are:

1. Cluster sequences within each species to get a reduced list of representative sequences
2. Align these representatives against core_nt to determine true taxonomic status
3. Propagate the alignment results back to the hits TSV

The partial subworkflow implemented by this PR works and passes local tests, and has been incorporated into the DOWNSTREAM workflow, which also passes tests. However, no new files are yet published. I decided to make this PR, even though the new subworkflow isn't finished, because it's already fairly substantial and I felt it was better to get review in chunks rather than drop a giant PR on someone.


subworkflows/local/validateViralAssignments/main.nf
```diff
@@ -0,0 +1,74 @@
...
+    main:
+        // 1. Join to viral taxonomy DB
+        rehead_ch = REHEAD_TSV(groups, "bowtie2_taxid_best", "taxid").output
+        rehead_sorted_ch = SORT_GROUP_TAXID(rehead_ch, "taxid").sorted
+        db_sorted_ch = SORT_DB_TAXID(
+            Channel.of("db").combine(db),
+            "taxid"
+        ).sorted.map{ sample, db -> db }
+        join_prep_ch = rehead_sorted_ch.combine(db_sorted_ch)
+        join_ch = JOIN_TSVS(join_prep_ch, "taxid", "inner", "taxonomy").output
```
katherine-stansifer, line 35: double checking that we're sure an inner join is what we want (I think it's right because I don't think we would ever have taxids in our output that didn't appear in the taxonomy db, but I'm always cautious)

subworkflows/local/validateViralAssignments/main.nf
```diff
@@ -0,0 +1,74 @@
...
+        // 2. Partition on species taxid
+        join_sorted_ch = SORT_JOINED_SPECIES(join_ch, "taxid_species").sorted
```
katherine-stansifer, line 37: how does this "taxid_species" column get in the tsv?

subworkflows/local/validateViralAssignments/main.nf
```diff
@@ -0,0 +1,74 @@
...
+                    if (!matcher) {
+                        def msg = "Filename doesn't match required pattern: ${sample}, ${path}, ${path.last()}"
+                        throw new IllegalArgumentException(msg)
+                    }
+                    // Convert to 2-tuples with key combining sample and species
+                    ["${sample}_${matcher[0][1]}", path]
+                }
+            }
+        // Then flatten
+        partitioned_flattened_ch = partitioned_tuple_ch.flatMap{it -> it}
```
katherine-stansifer, line 57: I asked Claude to help me with this bit as I was struggling to parse/understand, it helped me make sense of it and suggested:
* can you just go straight to flatMap rather than first making a map and then flattening it?
* the def matcher line is a beast and would be much easier to parse separated into multiple lines. Claude suggests
def filename = path.last()
def pattern = ...
def matcher = (filename =~ pattern)

However, I wonder if it would be better to make a version of PARTITION_TSVS that returns labeled partitions, rather than going through this regex process.

subworkflows/local/validateViralAssignments/main.nf
```diff
@@ -0,0 +1,74 @@
...
+        // 4. Extract into interleaved FASTQ format
+        fastq_ch = EXTRACT_FASTQ(partitioned_flattened_ch, false).output
+        // 5. Merge and concatenate pairs into single sequences for clustering
+        bbmerge_ch = BBMERGE(fastq_ch).reads
+        concat_ch = JOIN_FASTQ(bbmerge_ch, false).reads
+        // 6. Cluster merged reads
+        // TODO
+    emit:
```
katherine-stansifer, line 65: should probably have a TODO comment here about what the eventual output will be/is any of this test output expected to stick around.

katherine-stansifer (approved):
Looks good! Left a few comments but as it's a work in progress feel free to punt any until later.

modules/local/extractViralHitsToFastqNoref/main.nf
```diff
@@ -14,3 +14,20 @@ process EXTRACT_VIRAL_HITS_TO_FASTQ_NOREF {
         ln -s !{tsv} hits_in.tsv.gz
         '''
 }
+
```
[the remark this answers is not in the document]
willbradshaw: In this case we actually use both the unlabeled and labeled versions (in different workflows). We could easily enough convert the unlabeled use to use the labeled one, though.

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# v2.8.3.3-dev
+- Began development of post-doc validation of putative viral hits in the DOWNSTREAM workflow
```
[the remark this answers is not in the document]
willbradshaw: I personally feel it's important to validate one's postdocs

tests/modules/local/extractViralHitsToFastqNoref/labeled.nf.test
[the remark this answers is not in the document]
willbradshaw: It's only possible to test one process/workflow in a given .nf.test file, so if we want to test multiple processes in the same module we need different files.

(This is possibly just an argument for keeping each module to a single process.)

subworkflows/local/validateViralAssignments/main.nf
```diff
@@ -0,0 +1,74 @@
+/*
```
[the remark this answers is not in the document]
willbradshaw: Oh, yeah, that was an accident, will fix

tests/subworkflows/local/validateViralAssignments/main.nf.test
```diff
@@ -0,0 +1,101 @@
...
+        }
+        run("PREPARE_GROUP_TSVS") {
+            script "subworkflows/local/prepareGroupTsvs/main.nf"
+            process {
+                '''
+                input[0] = LOAD_DOWNSTREAM_DATA.out.input
+                '''
+            }
+        }
+        run("COPY_FILE_BARE") {
```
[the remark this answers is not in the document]
willbradshaw: For a very silly reason: nf-test can't handle S3 paths in tests, so in order to get the file into the local file tree we need to pass it through a nextflow process.

subworkflows/local/validateViralAssignments/main.nf
willbradshaw: This was added to the viral taxonomy DB in #270

subworkflows/local/validateViralAssignments/main.nf
willbradshaw: 👍

modules/local/extractViralHitsToFastqNoref/main.nf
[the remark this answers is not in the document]
katherine-stansifer: I'd be a fan of converting all to use the unlabeled version (though it doesn't need to be part of this PR)

tests/subworkflows/local/validateViralAssignments/main.nf.test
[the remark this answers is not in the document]
katherine-stansifer: thanks!

subworkflows/local/validateViralAssignments/main.nf
katherine-stansifer: Literally copy-pasting from an LLM and have not tried to run but looks reasonable, so take with that grain of salt

// Restructure channel to separate species
partitioned_flattened_ch = part_ch.flatMap{ sample, filepaths -> 
    // Ensure paths are a list
    def pathlist = (filepaths instanceof List) ? filepaths : [filepaths]
    
    // Process each path
    pathlist.collect { path ->
        // Extract species taxid from filename
        def filename = path.last()
        def pattern = /^partition_(.*?)_sorted_taxid_species_${sample}_taxonomy_inner_joined_taxid.*\.tsv.*\.gz$/
        def matcher = (filename =~ pattern)
        
        if (!matcher) {
            throw new IllegalArgumentException("Invalid filename format: ${filename} for sample ${sample}")
        }
        
        // Create tuple with combined sample_species key
        ["${sample}_${matcher[0][1]}", path]
    }
}

katherine-stansifer: And for the second, this might just push the horribleness to PARTITION_TSV (and might also break other processes), but I'm wondering if rather than
tuple val(sample), path("partition_*_${tsv}")
PARTITION_TSV could output
tuple val(sample), val(partition), path("partition_${partition}_${tsv}")

subworkflows/local/validateViralAssignments/main.nf
katherine-stansifer: yup I like a left join!

subworkflows/local/validateViralAssignments/main.nf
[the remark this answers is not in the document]
willbradshaw: Done

katherine-stansifer (approved):
Looks great now!

PR #286, opened 2025-04-22
Make the rest of RUN robust to empty files for Illumina data

willbradshaw:


tests/subworkflows/local/subsetTrim/main.nf.test
```diff
@@ -140,5 +140,57 @@ nextflow_workflow {
...
+        when {
+            params {
+                n_reads = 100
+            }
+            workflow {
+                """
+                input[0] = GZIP_FILE.out
+                input[1] = params.n_reads
+                input[2] = params.adapters
+                input[3] = Channel.of(true)
```
katherine-stansifer, line 167: Double checking that you meant to test on single-end data

tests/subworkflows/local/runQc/main.nf.test
```diff
@@ -121,5 +121,52 @@ nextflow_workflow {
...
+                }
+            }
+        }
+        when {
+            params {}
+            workflow {
+                """
+                input[0] = GZIP_FILE.out
+                input[1] = GZIP_FILE.out
+                input[2] = Channel.of(true)
```
katherine-stansifer, line 145: checking you meant to test on single-end data

tests/subworkflows/local/runQc/main.nf.test
```diff
@@ -121,5 +121,52 @@ nextflow_workflow {
...
+            // Basic stats TSV should have expected structure and content
+            def tsv_basic = path(workflow.out.qc_basic[0]).csv(sep: "\t", decompress: true)
+            assert tsv_basic.rowCount == 2
+            assert tsv_basic.columnCount == 15
+            // Lengths TSV should have expected structure and content
+            def tsv_lengths = path(workflow.out.qc_lengths[0]).csv(sep: "\t", decompress: true)
+            assert tsv_lengths.rowCount == 2
+            assert tsv_lengths.columnCount == 5
+            // Other TSVs should have expected structure and content
+            def tsv_adapt = path(workflow.out.qc_adapt[0]).linesGzip
```
katherine-stansifer, line 161: why do these use `linesGzip` to read rather than `.csv(sep: ..` ?

tests/subworkflows/local/countTotalReads/main.nf.test
```diff
@@ -40,7 +40,29 @@ nextflow_workflow {
...
+        tag "expect_success"
+        tag "empty_input"
+        config "tests/configs/run.config"
+
+        when {
+            params {}
+            workflow {
+                """
+                input[0] = Channel.of("empty_sample").combine(Channel.of(file("${projectDir}/test-data/toy-data/empty_file.txt")))
+                input[1] = Channel.of(true)
```
katherine-stansifer, line 55: same comment about being sure you want to test on single-end data

tests/subworkflows/local/profile/main.nf.test
```diff
@@ -5,6 +5,54 @@ nextflow_workflow {
...
+            // Output files should have expected structure and line count
+            def kraken_file = path(workflow.out.kraken[0]).linesGzip
+            def bracken_file = path(workflow.out.bracken[0]).linesGzip
+            assert kraken_file.size() == 1 // Just header
+            assert bracken_file.size() == 0 // Completely empty
+
+            // Kraken output should have expected fields
+            def kraken_headers = kraken_file[0].split("\t")
+            assert kraken_headers.size() == 10
+            def kraken_headers_exp = ["pc_reads_total", "n_reads_clade", "n_reads_direct", "n_minimizers_total", "n_minimizers_distinct", "rank","taxid", "name", "sample","ribosomal"]
```
katherine-stansifer, line 52: we repeat this a bunch of times already in the file, there's not a way to reuse this variable between tests is there?

katherine-stansifer (commented):
Added a number of small comments/questions; generally looks good. summarize-multiqc.R is the file where I am least confident of my ability to comprehensively review.

Overall, single-end illumina seems like a wacky choice for test cases; why did you go with this?

Also, how did you pick what to add an empty file test for?
- countReads I think should get a module-level empty file test
- If you tested only subworkflows that directly use the changed modules, you would have countTotalReads, qc, and profile
    -  You didn't add a test for qc, and you did add tests for runQc, subsetTrim, and blastViral, why?

tests/modules/local/addFixedColumn/main.nf.test
```diff
@@ -6,6 +6,69 @@ nextflow_process {
...
+            }
+        }
+        then {
+            // Should run without failures
+            assert process.success
+            
+            // Output should have correct header only
+            def output_file = path(process.out.output[0])
+            
+            // Read the file manually to avoid issues with compressed file handling
```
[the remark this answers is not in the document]
willbradshaw: Turns out it isn't, fixed.

willbradshaw:
> Also, how did you pick what to add an empty file test for?
> 
> countReads I think should get a module-level empty file test
> If you tested only subworkflows that directly use the changed modules, you would have countTotalReads, qc, and profile
> You didn't add a test for qc, and you did add tests for runQc, subsetTrim, and blastViral, why?

countReads actually already has empty-file tests from a previous PR!

For the rest, I went through and tested the subworkflows that are called by RUN. If they work with empty files, then so does everything called by that subworkflow, so I moved on. If not, I went deeper.

tests/subworkflows/local/profile/main.nf.test
katherine-stansifer: I'd be in favor of that! Using the heuristic that 2 copies of a code snippet is fine but when you get up to 3 it's worth refactoring to avoid. (Though up to your judgement; happy to approve the PR without it.)

katherine-stansifer (dismissed):

katherine-stansifer (approved):

PR #289, opened 2025-04-22
Katherine ont empty

katherine-stansifer:
This may have substantially more work, but I wanted to get your first-pass review before I go further.

Without any changes, the end-to-end empty file test of extractViralReadsONT ran fine! So I'm not sure I actually have any work to do.
MINIMAP2, MINIMAP2_NON_STREAMED, FILTLONG, PROCESS_VIRAL_MINIMAP2_SAM, MASK_FASTQ_READS, EXTRACT_SHARED_FASTQ_READS were the ONT-specific processes I identified.
I added a warning message about empty files to PROCESS_VIRAL_MINIMAP2_SAM because you'd done something similar for PROCESS_VIRAL_BOWTIE2_SAM; the others are command-line utilities (and it didn't look like you'd added empty file warnings for similar processes like BOWTIE2) so I left them unchanged
Should I be adding empty file tests for all of these? None of these? just PROCESS_VIRAL_MINIMAP2_SAM? Unsure how you're thinking about this.
Anything else I need to do that I've missed?
(Will update changelog once you've given me this initial feedback)

### DIFF
--- a/modules/local/processViralMinimap2Sam/resources/usr/bin/process_viral_minimap2_sam.py
+++ b/modules/local/processViralMinimap2Sam/resources/usr/bin/process_viral_minimap2_sam.py
@@ -109,7 +109,9 @@ def process_sam(sam_file, out_file, genbank_metadata, viral_taxids, clean_read_d
         out_fh.write(header)
         try:
             with pysam.AlignmentFile(sam_file, "r") as sam_file:
+                num_reads = 0
                 for read in sam_file:
+                    num_reads += 1
                     print_log(f"Processing read: {read.query_name}")
                     if read.is_unmapped or read.is_secondary or read.is_supplementary:
                         continue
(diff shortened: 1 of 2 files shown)


willbradshaw:
Yeah, I set an inconsistent example here; in #283 I carefully tested every process, but in #286 I only tested the top-level subworkflows and drilled down only if those failed. I understand why you're unsure.

If the extractViralReadsONT empty-file test still runs fine after the changes to PROCESS_VIRAL_MINIMAP2_SAM, I think we're fine to merge and move on.

One thing I'm a bit nervous about both here and for the Illumina PRs is checking that the pipeline works when empty and non-empty files are mixed together. But that should probably be its own PR.

willbradshaw (dismissed):

katherine-stansifer:
Yep, extractViralReadsONT for empty and non-empty files still passes (as does module level test for PROCESS_VIRAL_BOWTIE2_SAM, though that doesn't check for empty files). Updated changelog so will need you to re-approve and merge.

willbradshaw (approved):

PR #290, opened 2025-04-23
Add LCA algorithm

harmonbhasin:
Added function for running LCA and two tests, one that checks that it succeeds and fails successfully. I think the success test might need more conditions, but we'll see.

An improvement on this would be taking a divide and conquer approach.


harmonbhasin:
Unassigning Will as I made an update locally and I need to update the remote repo.

Issue #284, opened 2025-04-17
test_component_dependencies.py doesn't test dependencies of the target component


willbradshaw:
I've now experimented with test_component_dependencies.py in a couple of different contexts, and concluded that it's currently very useful for  isolating and running tests associated with a particular module/process, but much less so for workflows. This is because, despite its name, it doesn't actually pull in tests for components the target component depends on (the conventional meaning of "dependency" in programming); rather, it only pulls in those for components that depend on the target component.

For example, if I try running it with EXTRACT_VIRAL_READS_SHORT as the target component, it only identifies two test files to run:

   • tests/subworkflows/local/extractViralReadsShort/main.nf.test
   • tests/workflows/run.nf.test

i.e. the target component itself and the main workflow that calls it.

But EXTRACT_VIRAL_READS_SHORT depends on a bunch of processes and subworkflows whose tests I'd want to run before I felt comfortable saying the workflow was fully tested. Without that functionality I'm left running the whole local test suite to check I'm not missing something important.

simonleandergrimm:
I see why the name is confusing in that way and it shouldn't take that long
to implement.

But I wouldn't think that modules should break when editing a subworkflow?

In any case, I'll add a flag to test *actual* dependencies (or just make it
the default behavior), as you suggested.

simonleandergrimm:
Yeah I think adding this would save time insofar as you don't need to call
the function several times on different components when your work clusters
around one subworkflow!

simonleandergrimm:
@willbradshaw Adressed in #291.

PR #291, opened 2025-04-23
Adding ability to test all subcomponents of a component recursively

simonleandergrimm:
As discussed in #284 it would be useful to be able to test all the subcomponents that are used within a component. I.e., all the processes and subworkflows that are used by some subworkflow or workflow. I've now added the ability to do this.

Specifically, I:
- Introduced a function to get the file name of the component we want tested.
- Added code that parses the file to search for processes and subworkflows used in that component.
- Added code that recursively searches the components that are used by a component or its subcomponents.
- Added command-line argument that activates this entire process.

This code only works for workflows and subworkflows. @willbradshaw Can you provide an initial review?

### DIFF
--- a/bin/test_component_dependencies.py
+++ b/bin/test_component_dependencies.py
@@ -47,6 +47,7 @@ def identify_component_file(component):
         for hit in out:
             path, line_content = hit.split(":", 1)
             if line_content.lstrip().startswith("workflow "):
+                # TODO: This could pose a problem when there are multiple matches, and the first match is erroneous..
                 return path
     raise ValueError(f"Could not find workflow/subworkflow with name '{component}'.")
 
--- a/bin/test_component_dependencies.py
+++ b/bin/test_component_dependencies.py
@@ -34,6 +34,49 @@ def find_dependency(directory, component):
 
     return file_paths
 
+def identify_component_file(component):
+    # Search in subworkflows and modules directories
+    search_dirs = ["subworkflows/local/", "modules/local/", "workflows/"]
+
+    pattern = f"\\b{component}\\b"
+    for d in search_dirs:
+        out = subprocess.run(
+            ["grep", "-r", pattern, d], capture_output=True, text=True
+        ).stdout.splitlines()
+
+        for hit in out:
+            path, line_content = hit.split(":", 1)
+            if line_content.lstrip().startswith("workflow "):
+                return path
+    raise ValueError(f"Could not find workflow/subworkflow with name '{component}'.")
+
+def get_subcomponents(component_path):
+    with open(component_path, "r") as f:
+        component_content = f.read()
+
+    modules, workflows = set(), set()
+    for line in component_content.splitlines():
+        if "include" in line:
+            # Extract the first all caps variable in the include line
+            match = re.search(r'include\s*{\s*([A-Z][A-Z0-9_]*)', line)
+            if match:
+                if "subworkflows" in line:
+                    workflows.add(match.group(1))
+                else:
+                    modules.add(match.group(1))
+                continue
+    return modules, workflows
+
+def collect_component_dependencies(component):
+    file_path = identify_component_file(component)
+    modules, workflows = get_subcomponents(file_path)
+
+    for wf in list(workflows):
+        sub_mods, sub_wfs = collect_component_dependencies(wf)
+        modules.update(sub_mods)
+        workflows.update(sub_wfs)
+
+    return modules, workflows
 
 def workflow_uses_subworkflow(workflow, subworkflow_path):
     workflow_path = f"workflows/{workflow}"
@@ -92,14 +135,23 @@ def parse_args():
         default=False,
     )
 
+    parser.add_argument(
+        "-t",
+        "--test-subcomponents",
+        action="store_true",
+        help="Test subcomponents of the component. Only applicable if the component is a workflow/subworkflow.",
+        default=False,
+    )
+
     args = parser.parse_args()
     component = args.component
     verbose = args.verbose
-    return component, verbose
+    test_subcomponents = args.test_subcomponents
+    return component, verbose, test_subcomponents
 
 
 def main():
-    component, verbose = parse_args()
+    component, verbose, test_subcomponents = parse_args()
 
     # Validate component name - only allow alphanumerics and underscores
     if not re.match(r'^[a-zA-Z0-9_]+$', component):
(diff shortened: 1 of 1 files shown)


simonleandergrimm:
FWIW, I think the entire code could use some refactoring, but I want to make this PR too complex.

bin/test_component_dependencies.py
```diff
@@ -34,6 +34,50 @@ def find_dependency(directory, component):
...
+    return modules, workflows
+
+def collect_component_dependencies(component):
+    file_path = identify_component_file(component)
+    modules, workflows = get_subcomponents(file_path)
+
+    for wf in list(workflows):
+        sub_mods, sub_wfs = collect_component_dependencies(wf)
+        modules.update(sub_mods)
+        workflows.update(sub_wfs)
```
willbradshaw, line 78: Note that this will only go one layer deep (i.e. it will test dependencies of dependencies, but not dependencies of dependencies of dependencies). @simonleandergrimm is this intended behavior?

bin/test_component_dependencies.py
```diff
@@ -92,14 +136,23 @@ def parse_args():
         default=False,
     )
 
+    parser.add_argument(
+        "-t",
+        "--test-subcomponents",
+        action="store_true",
+        help="Test subcomponents of the component. Only applicable if the component is a workflow/subworkflow.",
+        default=False,
```
willbradshaw, line 144: I would probably have this on by default unless it creates speed problems for testing module components.

bin/test_component_dependencies.py
```diff
@@ -164,13 +217,34 @@ def main():
...
         print(f"   • {test}")
     print()
 
+
+    # Identifying tests of modules and subworkflows within the component (Optional)
+    if test_subcomponents:
+        subcomp_tests = set()
+        modules, workflows = collect_component_dependencies(component)
+        for module in modules:
+            subcomp_tests.update(find_dependency("tests/", module))
```
willbradshaw, line 234: Hmm, I can see the thinking here but it can quite easily lead to an explosion in the number of tests being run. E.g. if the component is a workflow that calls a subworkflow that calls a widely-used utility module, which is then called by a bunch of other workflows or tests, etc etc.

bin/test_component_dependencies.py
simonleandergrimm: I don't think this is right? What the code is doing is that it recursively checks for subworkflows in each subworkflow, until the `get_subcomponents` call returns no more workflows. (at which the for loop doesn't do anything any more).

E.g., if you run the script on `RUN` it pulls out modules used in `QC`.

willbradshaw: `list(workflows)` is defined at the start of the for loop and doesn't get re-evaluated as you add elements to `workflows` during the loop.

Here's a toy example to illustrate the behavior:

```
x = {1} # Starting set
for y in list(x):
    x.add(y+1)
    print(y)
```

If the list was updated iteratively the loop would run forever, but in fact it just runs once and prints "1".

bin/test_component_dependencies.py
simonleandergrimm: Your first question sounds more like you want recursion. Let me know what you prefer, I can drop this to only search for the subcomponents in the component (and nothing deeper).

bin/test_component_dependencies.py
simonleandergrimm: And also switched the argument to `--skip-subcomponents`.

bin/test_component_dependencies.py
simonleandergrimm: If we add a print statement, we get the expected behavior:
```
def collect_component_dependencies(component):
    file_path = identify_component_file(component)
    if file_path is None:
        return None, None
    modules, workflows = get_subcomponents(file_path)

    for wf in list(workflows):
        print(f"Collecting dependencies for {wf}")
        sub_mods, sub_wfs = collect_component_dependencies(wf)
        modules.update(sub_mods)
        workflows.update(sub_wfs)

    return modules, workflows
```

```
[ec2-user@ip-172-31-40-150 mgs-workflow]$ bin/test_component_dependencies.py -c RUN
[...]
Collecting dependencies for EXTRACT_VIRAL_READS_SHORT
Collecting dependencies for TAXONOMY
Collecting dependencies for MERGE_JOIN_READS
Collecting dependencies for COUNT_TOTAL_READS
Collecting dependencies for LOAD_SAMPLESHEET
Collecting dependencies for PROFILE
Collecting dependencies for TAXONOMY
Collecting dependencies for MERGE_JOIN_READS
Collecting dependencies for EXTRACT_VIRAL_READS_ONT
Collecting dependencies for RUN_QC
Collecting dependencies for QC
Collecting dependencies for BLAST_VIRAL
Collecting dependencies for CHECK_VERSION_COMPATIBILITY
Collecting dependencies for SUBSET_TRIM
========================================================================
Found 54 tests for subcomponents of RUN:
========================================================================
 [...]
```

bin/test_component_dependencies.py
willbradshaw: We should distinguish here between "downward recursion" (identifying dependencies of dependencies etc) and "upward recursion" (identifying dependents of dependents etc). I think it's reasonable to have both upward and downward recursion from the target component (identifying and testing all dependencies and dependents). But what's happening here is you're chaining upward and downward recursion together: testing dependents of dependencies and dependencies of dependents. This can quickly "zig-zag" out to a very large number of tests with only a very indirect relationship with the target component.

I don't think this is obviously wrong, to be clear. But I think it makes the expected number of tests being run much bigger in some cases, and should maybe be configurable.

simonleandergrimm: Ok so I now addressed this with a fairly minimal change:

For modules and subworkflows that are dependencies for the tested workflow, I just search tests in either the `modules` or `subworkflows` directories. 

This should reduce the number of tests that aren't directly related (e.g., by making it less likely that a utility module triggers a lot of subworkflow tests).

@willbradshaw let me know if you want something more that is more stringent.

I'd need to take a fairly detailed look at the scripts to check if there is some way of only selecting the test that is directly testing some module/subworkflow. File name matching alone doesn't do it, as not all tests are called `main.nf.test`.

willbradshaw: Sorry for delay!

This is better but it still has quite a bit of explosion potential; some subworkflows are very big.

Naive question: what would go wrong if we just cut this whole new block?

willbradshaw:
@simonleandergrimm this has been sitting with you for three months untouched. Can you handle it?

simonleandergrimm:
Yes, will handle once I’m back. Apologies for shirking this!

willbradshaw:
@simonleandergrimm back to you for getting up to date with `dev` so we can merge.

simonleandergrimm:
@willbradshaw Now merged with dev. Can we pull in?

PR #297, opened 2025-05-01
Katherine contam bugfix

katherine-stansifer:
I realized we weren't actually passing the output of contaminant filtering into MINIMAP2_VIRUS, so we were effectively skipping the contaminant filtering. I talked with simon and confirmed it was a bug. Ran local test for extractViralReadsONT and it passed


harmonbhasin:
@katherine-stansifer Will is gone today and tomorrow, so I would be happy to review this PR.

harmonbhasin (approved):
LGTM!

PR #300, opened 2025-05-05
Katherine multiqc empty

katherine-stansifer:
fix issue where multiqc basic stats has missing columns on empty input files (which leads to failure of concat when there are a mix of empty and non-empty files).

Ran local tests for summarizeMultiQC module and for qc and runQC subworkflows

### DIFF
--- a/tests/modules/local/summarizeMultiqc/main.nf.test
+++ b/tests/modules/local/summarizeMultiqc/main.nf.test
@@ -53,7 +53,7 @@ nextflow_process {
             def tsv_basic  = path(process.out[0][0][0]).csv(sep: "\t", decompress: true)
             with (tsv_basic) {
                 assert rowCount == 1
-                assert columnCount == 15
+                assert columnCount == 17
                 def zero_cols = ["percent_gc", "mean_seq_len", "n_reads_single", "percent_duplicates", "n_bases_approx"]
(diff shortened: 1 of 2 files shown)


modules/local/summarizeMultiqc/resources/usr/bin/summarize-multiqc.R
```diff
@@ -72,11 +72,21 @@ basic_info_fastqc <- function(fastqc_tsv, multiqc_json, single_end){
...
     select(-any_of(columns_exclude)) %>%
     select(n_bases_approx, everything()) %>%
     summarize_all(function(x) paste(x, collapse="/"))
+  
+  # Ensure per_base_sequence_quality and per_sequence_quality_scores are present 
+  # (they are missing from multiqc JSON if multiqc was run on empty file, but we always want them)
+  required_columns <- c("per_base_sequence_quality", "per_sequence_quality_scores")
+  missing_cols <- setdiff(required_columns, colnames(tab_tsv))
+  if (length(missing_cols) > 0) {
+    tab_tsv[missing_cols] <- NA
```
willbradshaw, line 87: Huh, I would not have predicted that would work, but I checked and it does.

willbradshaw (approved):

PR #303, opened 2025-05-08
Katherine extractviral single

katherine-stansifer:
Realized I needed to make extract viral reads support single-end reads; pulling this out as a separate PR to merge into my big column-changing PR.
Unsure of best way to specify endedness--for RUN we infer it from the samplesheet, but run_validation and downstream don't get a samplesheet. currently I have a boolean single_end in the configs; if preferred we could just infer from whether the viral hits TSV has query_seq_rev column.


modules/local/extractViralHitsToFastqNoref/resources/usr/bin/extract_viral_hits.py
willbradshaw, line 1: Couldn't this script just infer endedness on the basis of the columns present in the TSV?

katherine-stansifer: yes! I did by presence of query_seq_rev (you'd advised also checking that it was not NA for some rows, but we're processing the file line-by-line to avoid OOM errors, so I'm just going by presence of the field)

katherine-stansifer:
@willbradshaw back to you, changed to infer single end from file, all tests passing (except downstream due to snapshot issue). I'll do changelog updates as part of katherine-runvalidation-ont (the big column change PR)

PR #304, opened 2025-05-09
Post-hoc validation 3a: BLAST alignment

willbradshaw:
Coming back to post-hoc validation after a month away, and cleaning up + checking in the changes I'd already made but not merged before dropping it previously; I think getting these out of the way before doing new stuff makes the most sense.

This PR does two fairly simple things:

- Splits out core of BLAST_VIRAL subworkflow into a new BLAST_FASTA subworkflow that is called by both BLAST_VIRAL and VALIDATE_VIRAL_ASSIGNMENTS
- Adds tests for BLAST_FASTA and updated tests for VALIDATE_VIRAL_ASSIGNMENTS

Things the code does not yet do:

- Apply LCA to the tabular BLAST outputs (I have the code from @harmonbhasin to do this and will add it soon, but not in this PR)
- Actually do anything with the BLAST outputs in the DOWNSTREAM workflow, other than return them for testing


modules/local/downsampleFastnById/main.nf
```diff
@@ -9,7 +9,7 @@ process DOWNSAMPLE_FASTN_BY_ID {
         tuple val(sample), path("input_${fastn}"), path("input_${ids}"),  emit: input
     shell:
         '''
-        seqkit grep -f !{ids} !{fastn} | !{fastn.toString().endsWith(".gz") ? 'gzip -c' : 'cat'} > downsampled_!{fastn}
+        seqkit grep -f !{ids} !{fastn} | seqkit rmdup | !{fastn.toString().endsWith(".gz") ? 'gzip -c' : 'cat'} > downsampled_!{fastn}
```
harmonbhasin, line 12: why add this now?

harmonbhasin (approved):
I have one question, but that shouldn't block the merge. This all looks good to me!

modules/local/downsampleFastnById/main.nf
willbradshaw: Having duplicate sequence IDs was causing some tests to fail.

harmonbhasin: ah, sounds good thanks!

PR #310, opened 2025-05-14
Add column name customization features to LCA_TSV

willbradshaw:
This small PR modifies LCA_TSV to change the output column names:

1. It adds a "prefix" argument that allows all columns to receive a common prefix (e.g. "blast" for BLAST validation); this is important to distinguish e.g. Bowtie2 LCA from BLAST LCA.
2. It modifies the output column names to preserve input information -- e.g. if the input groups column name is "seq_id" then the corresponding output column will now also be "seq_id".


harmonbhasin (dismissed):
LGTM

harmonbhasin (approved):
LGTM

PR #315, opened 2025-05-20
Modify BLAST_FASTA to make LCA column prefix configurable

willbradshaw:
Tiny PR to make LCA column prefix in BLAST_FASTA configurable, and to modify the default in VALIDATE_VIRAL_ASSIGNMENTS to "validation" instead of "blast". This means we can change the aligner used for post-hoc validation in future without needing to rename columns (and so hopefully avoid a schema update).

Local tests for BLAST_FASTA all pass.

### DIFF
--- a/subworkflows/local/blastViral/main.nf
+++ b/subworkflows/local/blastViral/main.nf
@@ -37,7 +37,7 @@ workflow BLAST_VIRAL {
         // 3. Run BLAST and process output
         blast_ch = BLAST_FASTA(fasta_ch, ref_dir, blast_db_prefix,
             perc_id, qcov_hsp_perc, blast_max_rank, blast_min_frac,
-            taxid_artificial)
+            taxid_artificial, "blast")
         // 4. Rename subset FASTA file for output
         copy_ch = COPY_FILE(fasta_ch, "blast_input_subset.fasta.gz")
     emit:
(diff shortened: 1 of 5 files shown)


subworkflows/local/blastViral/main.nf
katherine-stansifer, line 1: do we also want to change blastViral to take a parameter?

katherine-stansifer (approved):
Looks good (assuming the local tests for BlastViral, ValidateViralAssignments, and anything else that uses BlastFasta pass in addition to the direct tests for BlastFasta)

subworkflows/local/blastViral/main.nf
willbradshaw: Hmm, why? ISTM we want the prefix to be predictable for a given subworkflow?

katherine-stansifer: ah, that makes sense! I think then I don't fully understand why we're not changing the prefixes of blastViral_ to be validation_ (just because we figure that sort of schema update can wait until later, if/when we change our valdiation alignment tool?)

willbradshaw: That's fair, I can change it to "validation".

katherine-stansifer (approved):
LGTM!

Issue #298, opened 2025-05-02
Failure in RUN:EXTRACT_VIRAL_READS_SHORT:SORT_BOWTIE_VIRAL


jeffkaufman:
This fails with OutOfMemoryError on some datasets in v2.9.0.0. 

See internally-filed issue https://twist.com/a/197793/ch/619193/t/6938296/

jeffkaufman:
Changing `modules/local/sortTsv/main.nf` as:

```
 process SORT_TSV {
     label "python"
-    label "single"
+    label "single_cpu_16GB_memory"
```

works for me, though it's probably overkill.

willbradshaw:
Hmm, it looks like this process is just calling GNU `sort` via `subprocess.run`, so will inherit whatever memory requirements that has.

@jeffkaufman how large are the samples for which this is failing? I agree 16GB is probably overkill, though also not the end of the world.

jeffkaufman:
That is surprising: IIRC `sort` is quite clever and seamlessly falls back to a disk-based merge sort if the input is too big for memory. 

Looking at https://github.com/naobservatory/mgs-workflow/blob/master/modules/local/sortTsv/resources/usr/bin/sort_tsv.py I think the problem is actually how we are wrapping `sort`:

```
sort_cmd = ["sort", "-t", "\t", "-k", f"{col_index+1},{col_index+1}", temp_sorted]
        sorted_data = subprocess.run(sort_cmd, capture_output=True, text=True, check=True).stdout
        
        # Write header and sorted data to output
        with open_by_suffix(output_file, "w") as out:
            out.write(header + "\n")
            out.write(sorted_data)
```

This buffers the entire output in memory. Instead, we could tell `sort` to write to another temp file and then concatenate that file with some headers to make the final output.

willbradshaw:
Seems right! @harmonbhasin assigning this to you.

jeffkaufman:
@willbradshaw is your `/tmp` backed by shared memory?  If so that would explain it:

* `tempfile.TemporaryDirectory()` will default to something under `/tmp`
* `sort` will handle files bigger than memory by using `/tmp`

You could fix this by either making `/tmp` not memory-backed or by picking a directory that's disk-backed and passing `dir` to `TemporaryDirectory` and `--temporary-directory` to `sort`.

PR #318, opened 2025-05-21
Rework SORT_TSV to avoid memory overload

willbradshaw:
Addresses #298 by reworking SORT_TSV to write intermediate sorted file body to a temporary file, rather than storing in memory. Also improves logging and code structure.


willbradshaw:
Ran local tests with `bin/test_component_dependencies.py -c SORT_TSV`; all tests passed.

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # v2.9.0.1-dev
 
+- Modified Github Actions to pull specific Nextflow version (rather than "latest")
```
katherine-stansifer, line 3: should include a note that this is a temp fix, right?

modules/local/sortTsv/resources/usr/bin/sort_tsv.py
```diff
@@ -1,133 +1,193 @@
...
-    converter = time.gmtime
 
-# Configure logger
-logger = logging.getLogger(__name__)
-logger.setLevel(logging.INFO)
+#=======================================================================
+# Configure logging
+#=======================================================================
+
+class UTCFormatter(logging.Formatter):
```
katherine-stansifer, line 28: this is fine and I think the class-based approach has advantages, but FWIW 
logging.Formatter.converter = time.gmtime (like I do in mgs-import https://github.com/naobservatory/mgs-import/blob/main/seq_import/__init__.py) is a more concise way to get logging in UTC.

modules/local/sortTsv/resources/usr/bin/sort_tsv.py
```diff
@@ -1,133 +1,193 @@
...
+    logger.info(f"Input file: {args.input_file}")
+    logger.info(f"Sort field: {args.sort_field}")
+    logger.info(f"Output file: {args.output_file}")
+    # Sort the TSV file
+    logger.info("Sorting TSV file.")
+    sort_tsv_file(args.input_file, args.output_file, args.sort_field)
+    # Finish time tracking
+    logger.info("Script completed successfully.")
+    end_time = time.time()
+    logger.info(f"Total time elapsed: {end_time - start_time} seconds")
```
katherine-stansifer, line 187: I know it's not new, but why do we have elapsed time logging for sort_tsv.py? We don't have this for our other python scripts.

modules/local/sortTsv/resources/usr/bin/sort_tsv.py
```diff
@@ -1,133 +1,193 @@
 #!/usr/bin/env python3
 
+"""
+Sort a TSV file by a specific column header using GNU sort.
+"""
+
+#=======================================================================
```
katherine-stansifer, line 7: this kind of annotation feels stylistically weird/un-pythonic and doesn't add any extra info (though this is a taste thing so if you want to keep it I won't argue)

modules/local/sortTsv/resources/usr/bin/sort_tsv.py
```diff
@@ -1,133 +1,193 @@
...
+        # Define temporary file paths
+        logger.debug(f"Temporary directory: {temp_dir}")
+        temp_body = os.path.join(temp_dir, "body.tsv")
+        temp_body_sorted = os.path.join(temp_dir, "body_sorted.tsv")
+        logger.debug(f"Temporary body files: {temp_body}, {temp_body_sorted}")
+        # Read the header and extract the sort field index
+        header = infile.readline().strip()
+        logger.debug(f"Header: {header}")
+        col_index = process_header(header, sort_field)
+        if col_index is None: # If no header, write empty output file
```
katherine-stansifer, line 125: is this really writing an empty output file? seems like it's not writing anything...

modules/local/sortTsv/resources/usr/bin/sort_tsv.py
```diff
@@ -1,133 +1,193 @@
...
+        return None
+    # Split the header line into fields
+    header_fields = header_line.split('\t')
+    # Look for the sort field in the header fields
+    if sort_field not in header_fields:
+        msg = f"Could not find sort field in input header: '{sort_field}', {header_fields}"
+        logger.error(msg)
+        raise ValueError(msg)
+    index = header_fields.index(sort_field)
+    logger.info(f"Found '{sort_field}' in column {index+1}. Sorting by column {index+1}.")
```
katherine-stansifer, line 96: this should include a note that sort_fields are 1-indexed (I'm assuming the reason for the +1 is that the sort uses 1-indexing? Otherwise I'd expect since this is a python script for it to be giving 0-indexed column positions.

katherine-stansifer (approved):
Looks good--left some comments but I don't need to re-review (though I'm happy to if the merge requires it :) )

modules/local/sortTsv/resources/usr/bin/sort_tsv.py
willbradshaw: We have it for a random subset of python scripts that happened to inherit their python code from other scripts that have it. I'm pro having it everywhere (I think it's useful and harmless) but not worth the effort of going through and adding it to everything at once.

modules/local/sortTsv/resources/usr/bin/sort_tsv.py
willbradshaw: Do you mean the docstring at the top, or the comment with the equals signs?

CHANGELOG.md
willbradshaw: I don't think so; we'll want to increment the specific version we're pulling, but I don't think we want to just pull "latest"; if a new version comes out we want to be intentional about switching to it.

modules/local/sortTsv/resources/usr/bin/sort_tsv.py
willbradshaw: If the former, I think this is an established practice; Gemini says: "At the top of a Python script, a docstring serves to describe the script's overall purpose and functionality. It should be the first statement in the file, placed after the shebang line (if present) and before any imports or code. The docstring is enclosed in triple quotes (""") and can be single-line or multi-line."

modules/local/sortTsv/resources/usr/bin/sort_tsv.py
willbradshaw: Cool, thanks! I don't want to change this here but I'll try this out in a future script.

modules/local/sortTsv/resources/usr/bin/sort_tsv.py
willbradshaw: Yeah, if you open a file object with `open` in write mode and then don't write anything to it, the file will exist but be empty.

modules/local/sortTsv/resources/usr/bin/sort_tsv.py
willbradshaw: Done

willbradshaw:
@katherine-stansifer need your re-approval

modules/local/sortTsv/resources/usr/bin/sort_tsv.py
katherine-stansifer: Sorry to be unclear---I like the docstring at the top, it definitely adds info and is very pythonic! I meant the ============== things that say stuff like "main function" and "imports"--we can see that it's they are imports/main function from the code

katherine-stansifer (approved):

modules/local/sortTsv/resources/usr/bin/sort_tsv.py
willbradshaw: Ah, okay. I find it much easier to read code when there are clear section delimiters like this.

PR #301, opened 2025-05-06
column name documentation

katherine-stansifer:

PR #321, opened 2025-05-23
Concatenate and clean up post-hoc validation output

willbradshaw:
- Switched selectTsvColumns drop mode to proceed with a warning if attempting to drop columns that aren't there (rather than throwing an error)
    - This lets me use it for "guarantee no such column is present" steps in subworkflows that might be called in multiple contexts
- Split out final concatenation step from validateViralAssignments into its own subworkflow, concatenateTsvsAcrossSpecies, and implemented tests for it.
- Applied concatenateTsvsAcrossSpecies to intended output files for validateViralAssignments.
- Reviewed and cleaned up post-hoc validation output files
    - Dropped columns from the viral taxonomy DB
    - Added "vsearch_" prefix to cluster information columns for clarity
    - Renamed "taxid" column back to "bowtie2_taxid_best"
    - Simplified output filenames to "<group>_validation_hits.tsv.gz"


modules/local/processVsearchClusterOutput/resources/usr/bin/process_vsearch_cluster_output.py
```diff
@@ -134,9 +134,12 @@ def extract_representative_ids(output_db, n_clusters):
     logger.info(f"Extracted representative IDs: {representative_ids.tolist()}.")
     return representative_ids
 
-def save_output_db(output_db, out_path):
+def save_output_db(output_db, out_path, output_prefix):
     """Save processed DataFrame to output path."""
     logger.info(f"Saving output DB to {out_path}.")
+    if output_prefix: # Add prefix to all column names except for seq_id
+        output_db = output_db.add_prefix(output_prefix + "_")
+        output_db = output_db.rename(columns={output_prefix + "_seq_id": "seq_id"})
```
katherine-stansifer, line 142: this is something I should have noticed earlier (not new in this PR)--but I thought we were trying to avoid pandas dependencies and also read/write all tsvs streaming? (hence the duplicated tsv parsing code everywhere). Why pandas here?

modules/local/selectTsvColumns/resources/usr/bin/select_tsv_columns.py
```diff
@@ -109,11 +129,12 @@ def select_columns(input_path: str, output_path: str, fields: list[str], mode: s
             raise ValueError("No header to select fields from.")
         headers_in = header_line.split("\t")
         # Get indices of selected fields
-        indices = [get_header_index(headers_in, field) for field in fields]
+        indices = [get_header_index(headers_in, field, mode) for field in fields]
         # Get indices of fields to keep
         if mode == "keep":
             indices_keep = indices
         elif mode == "drop":
+            indices = [i for i in indices if i is not None]
```
katherine-stansifer, line 137: I think this can just be ...if i] rather than ...if i is not None]

modules/local/processVsearchClusterOutput/main.nf
```diff
@@ -5,16 +5,19 @@ process PROCESS_VSEARCH_CLUSTER_OUTPUT {
...
         tuple val(sample), path("input_${db}"), emit: input
-    shell:
-        '''
-        in=!{db}
-        out_db=!{sample}_vsearch_tab.tsv.gz
-        out_id=!{sample}_vsearch_ids.txt
-        process_vsearch_cluster_output.py -n !{n_clusters} ${in} ${out_db} ${out_id}
-        ln -s ${in} input_${in}
-        '''
+    script:
```
katherine-stansifer, line 13: I was wondering about this shell->script...but then I looked at nextflow documentation and it looks like the shell block is deprecated! Do we need an issue to switch over other modules? I think we have a ton of shell blocks.

subworkflows/local/concatenateTsvsAcrossSpecies/main.nf
```diff
@@ -0,0 +1,56 @@
...
+| WORKFLOW |
+***********/
+
+workflow CONCATENATE_TSVS_ACROSS_SPECIES {
+    take:
+        tsvs // Channel of 2-tuples, each containing a combined group/species label and a TSV path
+        filename_suffix // Suffix for output filenames
+    main:
+        // 0. Remove existing group and group_species columns, if present
+        drop_group_ch = DROP_GROUP_COLUMN(tsvs, "group", "drop").output
```
katherine-stansifer, line 33: when/why would these already be present?

subworkflows/local/concatenateTsvsAcrossSpecies/main.nf
```diff
@@ -0,0 +1,56 @@
...
+        // 2. Restructure channel to replace group/species label with group label only
+        split_label_ch = label_ch.map{
+            label, path ->
+                def pattern = /^(.*?)_(\d+)$/
+                def matcher = (label =~ pattern)
+                if (!matcher) {
+                    def msg = "Group label doesn't match required pattern: ${label}, ${path}, ${pattern}"
+                    throw new IllegalArgumentException(msg)
+                }
+                [matcher[0][1], path]
```
katherine-stansifer, line 46: I'm sure this is right/tested but for my own edification why is this matcher[0][1] rather than just matcher[0]?

subworkflows/local/propagateValidationInformation/main.nf
```diff
@@ -27,15 +27,16 @@ workflow PROPAGATE_VALIDATION_INFORMATION {
...
         validation_tsv // Validation TSV, including cluster_rep_id and validation status
+        taxid_column // Column header for taxid in hits TSV
     main:
         // 0. Drop redundant columns from validation TSV
-        select_ch = SELECT_TSV_COLUMNS(validation_tsv, "taxid", "drop").output
+        select_ch = SELECT_TSV_COLUMNS(validation_tsv, taxid_column, "drop").output
         // 1. Sort cluster and validation TSVs by cluster_rep_id
-        sort_cluster_tsv = SORT_CLUSTER_TSV(cluster_tsv, "cluster_rep_id").sorted
-        sort_validation_tsv = SORT_VALIDATION_TSV(select_ch, "cluster_rep_id").sorted
+        sort_cluster_tsv = SORT_CLUSTER_TSV(cluster_tsv, "vsearch_cluster_rep_id").sorted
```
katherine-stansifer, line 35: do you want to pass the "vsearch_" prefix as a parameter as well since it's now configurable?

subworkflows/local/splitViralTsvBySpecies/main.nf
```diff
@@ -32,27 +33,28 @@ workflow SPLIT_VIRAL_TSV_BY_SPECIES {
...
-            Channel.of("db").combine(db),
-            "taxid"
-        ).sorted.map{ group, db -> db }
-        join_prep_ch = rehead_sorted_ch.combine(db_sorted_ch)
-        join_ch = JOIN_TSVS(join_prep_ch, "taxid", "left", "taxonomy").output
-        // 2. Partition on species taxid
+        // 1. Prepare taxonomy DB for joining
+        db_ch = Channel.of("db").combine(db)
+        db_select_ch = SELECT_TSV_COLUMNS(db_ch, "taxid,taxid_species", "keep").output
+        db_rehead_ch = REHEAD_TSV(db_select_ch, "taxid", "bowtie2_taxid_best").output
```
katherine-stansifer, line 39: note to self that when we merge #301 we'll need to change this to aligner_taxid; hopefully this won't mess anything up too much

katherine-stansifer (approved):
Looks great!

modules/local/processVsearchClusterOutput/main.nf
willbradshaw: Yeah I think this happened fairly recently. I'm sad about this because I liked the shell block, but ah well. We probably do want to switch over other modules at some point, yeah, though I don't think it's urgent. It's also very claude-code-able.

subworkflows/local/splitViralTsvBySpecies/main.nf
willbradshaw: Yeah, I was conscious of this when I was writing it, but needed to make it work with the current version of `dev`.

subworkflows/local/propagateValidationInformation/main.nf
willbradshaw: I don't think so. It's configurable at the process level (PROCESS_VSEARCH_CLUSTER_OUTPUT) but not the subworkflow level (CLUSTER_VIRAL_ASSIGNMENTS).

subworkflows/local/concatenateTsvsAcrossSpecies/main.nf
katherine-stansifer: I wonder if raising an error if present (and fixing the tests so it doesn't happen there) would be a better solution--I think that "we don't know if the column will ever be present already" is not a great situation to be it. But I don't feel strongly--up to you!

PR #329, opened 2025-05-30
Rework SORT_TSV to avoid memory overload (attempt 2)

willbradshaw:
My last attempt to address #298 (#318) was only partially successful; some runs still failed with OOM errors. This PR takes a more thorough approach, explicitly specifying GNU sort's memory allocation as well as using a custom temporary directory that's definitely writing to disk. I've tested it on one of the runs that was previously failing and it now succeeds, so I'm optimistic that this time will work.

### DIFF
--- a/modules/local/sortTsv/main.nf
+++ b/modules/local/sortTsv/main.nf
@@ -12,7 +12,7 @@ process SORT_TSV {
     shell:
         '''
         # Run the Python script to sort the TSV
-        sort_tsv.py !{tsv} !{sort_field} sorted_!{sort_field}_!{tsv}
+        sort_tsv.py -m !{task.memory.toGiga()} !{tsv} !{sort_field} sorted_!{sort_field}_!{tsv}
         
         # Link input to output for testing
         ln -s !{tsv} input_!{tsv}
--- a/modules/local/sortTsv/main.nf
+++ b/modules/local/sortTsv/main.nf
@@ -2,7 +2,7 @@
 // Uses Python script to handle empty files properly
 process SORT_TSV {
     label "python"
-    label "single"
+    label "single_cpu_16GB_memory"
     input:
         tuple val(sample), path(tsv)
         val(sort_field)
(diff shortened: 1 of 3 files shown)


modules/local/sortTsv/resources/usr/bin/sort_tsv.py
```diff
@@ -136,28 +145,49 @@ def sort_tsv_file(input_file: str,
...
+             memory_limit: int) -> None:
     """
     Run the sort command on a TSV file with a specified field separator,
     handling errors and logging, and write to an output file.
     Args:
         input_file (str): The input TSV file path.
         output_file (str): The output TSV file path.
         sort_index (int): The index of the column to sort by.
+        temp_dir (str): The temporary directory for GNU sort to use.
+        memory_limit (int): The memory limit in GB.
```
jeffkaufman, line 171: How about `memory_limit_gb`?  Always good to include units.

modules/local/sortTsv/resources/usr/bin/sort_tsv.py
```diff
@@ -48,6 +49,7 @@ def parse_args() -> argparse.Namespace:
     parser.add_argument("input_file", help="Input TSV file path")
     parser.add_argument("sort_field", help="Column header to sort by")
     parser.add_argument("output_file", help="Output TSV file path")
+    parser.add_argument("--memory-limit", "-m", type=int, help="Memory limit in GB", required=True)
```
jeffkaufman, line 52: How about `--memory-limit-gb`?  Always good to include units.

modules/local/sortTsv/main.nf
```diff
@@ -2,7 +2,7 @@
 // Uses Python script to handle empty files properly
 process SORT_TSV {
     label "python"
-    label "single"
+    label "single_cpu_16GB_memory"
```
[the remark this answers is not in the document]
willbradshaw: It's not _needed_, but it makes it run faster.

PR #338, opened 2025-06-03
Modify post-hoc validation workflows to reduce BLAST fixed costs

willbradshaw:
- Rearranges VALIDATE_VIRAL_ASSIGNMENTS to concatenate across species before rather than after BLAST_VIRAL, thus dramatically reducing the number of BLAST processes that need to run and so reducing per-process fixed costs.
- Also involves modifying PROPAGATE_VALIDATION_INFORMATION and introducing a new subworkflow, CONCATENATE_FASTA_ACROSS_SPECIES (which in turn depends on a new process, CONCATENATE_FASTN_LABELED).


modules/local/concatenateFastn/main.nf
katherine-stansifer, line 1: We already have a lot of very similar processes to concatenate files, and I don't like adding one more. And I don't think this has any FASTA/FASTQ specific logic, does it? I'd prefer to make CONCATENATE_FILES labeled and rejigger the subworkflows that use it to deal with labeled input/output. Would be great if we could replace CONCATENATE_FASTA and/or CONCATENATE_GENOME_FASTA too but that may be more than is reasonable for this PR.

katherine-stansifer: though I guess including label in name is not what we want for those...but it does seem like there should be a way to make this configurable rather than having a bunch of very similar wrappers around `cat`

subworkflows/local/propagateValidationInformation/main.nf
```diff
@@ -30,21 +31,23 @@ workflow PROPAGATE_VALIDATION_INFORMATION {
...
+        sort_validation_tsv = SORT_VALIDATION_TSV(drop_taxid_ch, "vsearch_cluster_rep_id").sorted
         // 2. Left-join cluster and validation TSVs by cluster_rep_id
         combine_1_ch = sort_cluster_tsv.combine(sort_validation_tsv, by: 0)
         join_1_ch = JOIN_CLUSTER_VALIDATION(combine_1_ch, "vsearch_cluster_rep_id", "left", "representative").output
         // 3. Sort intermediate TSV by seq_id (hits TSV should already be sorted)
         sort_join_1_tsv = SORT_JOIN_TSV(join_1_ch, "seq_id").sorted
         // 4. Strict-join intermediate and hits TSVs by seq_id
         // Both tables should contain the same sequences in the same order
-        combine_2_ch = hits_tsv.combine(sort_join_1_tsv, by: 0)
+        drop_group_ch = DROP_GROUP(sort_join_1_tsv, "group", "drop").output
```
katherine-stansifer, line 45: why do we now have to drop the group, and why couldn't you do it at the same time you drop taxid?

subworkflows/local/concatenateFastaAcrossSpecies/main.nf
```diff
@@ -0,0 +1,43 @@
...
+| MODULES AND SUBWORKFLOWS |
+***************************/
+
+include { CONCATENATE_FASTN_LABELED } from "../../../modules/local/concatenateFastn"
+
+/***********
+| WORKFLOW |
+***********/
+
+workflow CONCATENATE_FASTA_ACROSS_SPECIES {
```
katherine-stansifer, line 21: this is very similar logic to CONCATENATE_TSVS_ACROSS_SPECIES since the most annoying code is to replace group/species label with group label only-- is there a way to reuse code, or (probably my preference) would it make sense to have one subworkflow, CONCATENATE_ACROSS_SPECIES, that concatenates both the fastas and tsvs? 

(and a general comment -- you're ending up with more subprocesses than I would have expected devoted to post hoc validation -- probably fine but wanted to make sure you've thought in general whether any processes could be removed or combined

subworkflows/local/validateViralAssignments/main.nf
```diff
@@ -81,6 +81,8 @@ workflow VALIDATE_VIRAL_ASSIGNMENTS {
         test_cluster_ids = cluster_ch.ids
```
katherine-stansifer, line 81: this is a LOT of test outputs, do we really need all of them?

tests/modules/local/concatenateFastn/main.nf.test
```diff
@@ -0,0 +1,213 @@
+nextflow_process {
```
katherine-stansifer, line 1: I don't think your process actually has any fasta/fastq specific logic, nor does it require your extensions to include fasta/fastq, so I think there's no reason to test on both fasta and fastq. (though testing on fasta.gz makes sense to ensure it handles double extensions properly)

modules/local/concatenateFastn/main.nf
```diff
@@ -0,0 +1,22 @@
...
+    label "coreutils"
+    input:
+        tuple val(label), path(files)
+        val(name)
+    output:
+        tuple val(label), path("${label}_${name}.*"), emit: output
+        tuple val(label), path("${label}_input_${files[0]}"), emit: input
+    script:
+        // Extract extension from first file, handling double extensions like .fasta.gz
+        def first_file = files[0].toString()
```
katherine-stansifer, line 13: should document behavior with a mix of extensions (if it's easy to add a check that they are all the same I would, but if that's annoying/complex I'd simply add a comment saying that it expects all files to have the same extension and if they are mixed will arbitrarily take first extension)

katherine-stansifer (commented):
looks good, left some comments mostly around ideas to simplify/reduce # of modules and subprocesses

subworkflows/local/propagateValidationInformation/main.nf
willbradshaw: I think we could drop it at the same time, though, will try it.

subworkflows/local/concatenateFastaAcrossSpecies/main.nf
willbradshaw: You're right that I could integrate them both into one subworkflow and define a function for the label wrangling that I called twice. My main reason not to do this is that I'm not confident I won't need this functionality in the future, and if I do, having a basic concatenate-across-species workflow that doesn't involve specialised TSV wrangling seems useful. 

On that note, I don't think this workflow is actually specific to FASTA (it will work on any process that can be simply concatenated) so I renamed it to CONCATENATES_FILES_ACROSS_SPECIES.

tests/modules/local/concatenateFastn/main.nf.test
willbradshaw: You're right, this is unnecessary. Removed.

subworkflows/local/propagateValidationInformation/main.nf
willbradshaw: Consolidating the calls seems to work!

willbradshaw: Ah, spoke too soon, never mind. The two columns that need to be dropped are coming from different inputs, so we need two calls either way.

subworkflows/local/validateViralAssignments/main.nf
willbradshaw: I removed the ones that aren't being used by the test.

modules/local/concatenateFastn/main.nf
katherine-stansifer: I do think it's worth it (though not urgent, so could make a P1 issue and get to it in the next week or so). I think having multiple near-identical processes is a quite bad form of technical debt.

subworkflows/local/propagateValidationInformation/main.nf
katherine-stansifer: sounds good! thanks for trying!

modules/local/concatenateFastn/main.nf
willbradshaw: Filed #342

katherine-stansifer (approved):
LGTM!

PR #348, opened 2025-06-09
Overhaul duplicate marking to improve computational efficiency

willbradshaw:
Several changes to the main Rust script for the MARK_ALIGNMENT_DUPLICATES module to improve performance through multithreading and algorithmic improvements.


harmonbhasin:
I see that the downstream test is failing, I don't think that should be happening.

harmonbhasin:
I'll let you decide what you need to do.

PR #351, opened 2025-06-12
Make trace files unique across runs

willbradshaw:
Modifies Nextflow trace files output by RUN and DOWNSTREAM workflows to be unique across multiple runs in the same directory, by appending the timestamp to the file name. This was previously implemented by @harmonbhasin and then taken out by me; I'm now restoring it. This is important for effectively monitoring costs when running the workflow with mgs-orchestrator.


harmonbhasin (approved):
LGTM

PR #360, opened 2025-06-20
Implementation of LCA for ONT

harmonbhasin:
Mostly consists of:
- Making minimap2 give multiple alignments
- Changing the viral Minimap2 script to keep secondary + supplementary alignments
- Running LCA on the Minimap2 TSV

### DIFF
--- a/tests/modules/local/minimap2/align.nf.test
+++ b/tests/modules/local/minimap2/align.nf.test
@@ -68,6 +68,20 @@ nextflow_process {
 
             // Verify input FASTQ contains all reads
             assert input_reads == fastq_read_ids_mapped + fastq_read_ids_unmapped
+
+            // Verify that some reads have multiple alignments (due to -N 10 parameter)
+            def read_alignment_counts = samlines
+                .collect { line -> line.split('\\t')[0] }  // Get read IDs
+                .groupBy { it }  // Group by read ID
+                .collectEntries { k, v -> [k, v.size()] }  // Count alignments per read
+            
+            // With -N 10, at least some reads should have multiple alignments
+            def reads_with_multiple_alignments = read_alignment_counts.values().findAll { it > 1 }
+            assert reads_with_multiple_alignments.size() > 0, "Expected some reads to have multiple alignments with -N 10 parameter"
+            
+            // Verify no read exceeds the -N 10 limit
+            def max_alignments_per_read = read_alignment_counts.values().max() ?: 0
+            assert max_alignments_per_read <= 10, "Expected no read to have more than 10 alignments with -N 10 parameter"
         }
     }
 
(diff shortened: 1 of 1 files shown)


harmonbhasin:
Passed both the tests I created.

harmonbhasin:
Confirmed that the normal and snapshot test for the ONT workflow still works. Had to run these as updating the normal minimap2 process to pass in alignment parameters requires me passing in an empty string to those processes.

harmonbhasin:
just realized that i need to udpate the process tests and that the contaminant is using a non-streamed version.

modules/local/minimap2/main.nf
```diff
@@ -51,7 +52,7 @@ process MINIMAP2 {
         #   - Second branch (samtools view -u -F 4 -) filters SAM to aligned reads and saves FASTQ
         #   - Third branch (samtools view -h -F 4 -) also filters SAM to aligned reads and saves SAM
         zcat ${reads} \
-            | minimap2 -a ${idx} /dev/fd/0 \
+            | minimap2 -a !{alignment_params} ${idx} /dev/fd/0 \
```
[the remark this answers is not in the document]
harmonbhasin: good question, i mainly did this to allow the user to pass in more settings in-case they would like to (similar to what we do for Bowtie2).

subworkflows/local/extractViralReadsONTLca/main.nf
```diff
@@ -0,0 +1,84 @@
...
+        // Generate TSV of viral hits, and sort
+        tsv_ch = PROCESS_VIRAL_MINIMAP2_SAM(sam_fastq_ch, genome_meta_path, virus_db_path)
+        tsv_sorted_ch = SORT_TSV(tsv_ch.output, "seq_id")
+        // Run LCA on viral hits TSV
+        lca_ch = LCA_TSV(tsv_sorted_ch.sorted, nodes_db, names_db,
+            "seq_id", "aligner_taxid", "aligner_length_normalized_score", taxid_artificial,
+            "aligner")
+        // Sort both TSV files by seq_id for joining
+        lca_sorted_ch = SORT_LCA(lca_ch.output, "seq_id")
+        minimap2_sorted_ch = SORT_MINIMAP2_VIRAL(tsv_ch.output, "seq_id")
```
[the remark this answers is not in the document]
harmonbhasin: good point, we actually totally can!

harmonbhasin:
Passes:
- PROCESS_VIRAL_MINIMAP2_SAM_LCA
- MINIMAP2
- extractViralReadsONTLca

harmonbhasin:
extractViralReadsONTLca passes after this change.

tests/modules/local/minimap2/align.nf.test
```diff
@@ -31,6 +31,58 @@ nextflow_process {
                 input[1] = "${params.ref_dir}/results/mm2-virus-index"
                 input[2] = "test"
                 input[3] = false
+                input[4] = "-N 10"
```
katherine-stansifer, line 34: would be good to add a check here that some reads DO have multiple alignments, and in the next test than each read has at most 1 alignment.

katherine-stansifer, line 34: would be good to add a check here that some reads DO have multiple alignments, and in the next test than each read has at most 1 alignment.

katherine-stansifer (dismissed):
Looks very good! I have one suggestion for the align.nf.test.

tests/modules/local/minimap2/align.nf.test
harmonbhasin: I just remembered that we can't look for at most 1 alignment since Minimap2 by default does return up to 5 secondary alignments, but I can check that the number is below 5.

> Minimap2 rates an alignment by the score of the max-scoring sub-segment, excluding introns, and marks the best alignment as primary in SAM. When a spliced gene also has unspliced pseudogenes, minimap2 slightly prefers the spliced alignment. By default, minimap2 outputs up to five secondary alignments (i.e. likely pseudogenes in the context of RNA-seq mapping). This can be tuned with option -N.

https://github.com/lh3/minimap2?tab=readme-ov-file#started

PR #362, opened 2025-06-20
Backport nextflow 25.04 compatibility and fixed release of virushostdb to stable 

katherine-stansifer:
Local tests for top-level workflows pass (Will said this was sufficient for this PR and I didn't need to run all local tests). Small benchmark dataset runs successfully through RUN and DOWNSTREAM.
@harmonbhasin we may need to get an admin to approve the merge--not sure how stable is set up--but I'm hoping you can take a first look at least.

### DIFF
--- a/.github/workflows/end-to-end.yml
+++ b/.github/workflows/end-to-end.yml
@@ -5,6 +5,7 @@ on:
     branches:
       - master
       - dev
+      - stable
 
 jobs:
   test-index:


.github/workflows/end-to-end.yml
```diff
@@ -21,16 +22,16 @@ jobs:
...
-          java-version: '11'
-          distribution: 'adopt'
+          java-version: '21'
+          distribution: 'temurin'
 
       - name: Setup Nextflow latest (stable)
         uses: nf-core/setup-nextflow@v1
         with:
-          version: "24.10.6"
+          version: "25.4.3"
```
harmonbhasin, line 34: Prior to the breaking of code, we had this as "latest", do you think we should set a specific version, or use "latest". Furthermore, if we do keep "25.4.3", we should come up with some system for making sure that we do update this version at some frequency.

harmonbhasin (approved):
Looks good to me, I just have one question.

.github/workflows/end-to-end.yml
katherine-stansifer: Will specifically said specific version so we don't have the kind of surprise breakage that happened with 24.10.6, but good point that we should update! I added "think about updating" as a step during releases https://docs.google.com/document/d/1stErYWczyx-9z8kAsSGFXRcoEWhTqB_RwgqMnXnA-4s/edit?tab=t.0

Issue #363, opened 2025-06-25
Change our "shell" blocks in nextflow to "script" blocks


katherine-stansifer:
The "shell" block is now deprecated and means our linting in nextflow 25 produces a TON of warnings. This is a straightforward migration, we just have a lot to migrate.

PR #364, opened 2025-06-26
compliance with strict syntax

katherine-stansifer:
Made pipeline compliant with strict syntax:
* Linting no longer gives errors (except on config files with their "<>" placeholders; I don't think this is worth fixing right now)
* Only remaining warnings are shell blocks (opened https://github.com/naobservatory/mgs-workflow/issues/363) 
     * And one for subworkflows/local/extractViralReadsShortLca/main.nf: host_taxon variable was declared but not used (@harmonbhasin I didn't want to mess with your LCA code but you should fix this when you fully integrate LCA)

* I added tests for maskGenomeFasta (because my strict syntax compliance required editing this process and I wanted confidence that it still worked)

I also folded in a few tiny improvements that don't really fit in this PR (let me know if you want me to separate these out):
* Defined seed for tests of subsetFastn (because it routinely gives stochastic failures when i run all tests and it's annoying)
* end-to-end tests now execute when merging into stable. 

All local tests pass.


harmonbhasin:
In the future, those changes should probably be separate PRs (I need to also work on this, but i think this is best practice); however, they're super small and don't really make it harder to review the current code so they're fine by me!

harmonbhasin (commented):
Everything looks good! I like having the "_" syntax for unused variables, and knowing that we don't need to import specific Java libraries is also useful. Excited to use the linter going forward!

harmonbhasin (approved):

PR #367, opened 2025-07-01
Katherine add run_ont.config

katherine-stansifer:
To allow us to run ONT data through orchestrator, it's very helpful to have an ONT config file with the correct defaults

(see https://twist.com/a/197793/ch/619193/t/7080407/ for discussion).

Added such a file, renamed run.config to run_illumina.config, and updated documentation accordingly.

Note that tests/configs has run.config and run_ont.config; I wasn't sure whether to rename tests/configs/run.config to tests/configs/run_illumina.config (I'm leaning towards yes but didn't want to do so until I discussed because it's a lot of file changes)


harmonbhasin:
> Note that tests/configs has run.config and run_ont.config; I wasn't sure whether to rename tests/configs/run.config to tests/configs/run_illumina.config (I'm leaning towards yes but didn't want to do so until I discussed because it's a lot of file changes)

The existing changes looks good to me. I do think we should update the tests. (also, even though there are a lot of files to change, it hopefully shouldn't take too long using something like find + replace in VS Code or your other editor of choice; although one annoying thing is that you will have to rerun all of the existing tests).

I can look at this again once the tests have been changed.

katherine-stansifer:
Note that @harmonbhasin and I talked in person: since we want to support aviti/pacbio as well, we decided to use run.config and run_ont.config, both for the main config files and for the test config files

harmonbhasin (approved):
LGTM! I was going to say that we needed to update the tests to use `run_ont.config`, but we already have that there (thanks Katherine for confirming this).

PR #375, opened 2025-07-10
Adjusting processLcaAlignerOutput to be more flexible such that we can use it for ONT LCA

harmonbhasin:
As the title says, due to ONT having different output columns than the short read pipeline (no `rev` columns from aligner), we need to adjust the workflow a bit.

Tests passing:
- processLcaAlignerOutput
- extractViralReadsShortLca


katherine-stansifer (approved):
LGTM!

PR #381, opened 2025-07-14
Adjusting SORT_FASTQ to use alphanumeric sorting of read id's instead of dictionary sorting

harmonbhasin:
Our python script in filter_viral_sam checks that the read id's are sorted alphanumerically, however SORT_FASTQ sorts dictionary-wise, so I'm changing this to keep things consistent. SORT_FASTQ is using in processLcaAlignerOutput, blastViral, extractViralReadsShortLca, and the tests pass for all of these, including the test for SORT_FASTQ.


willbradshaw:
Actually, @harmonbhasin you need to bump the CHANGELOG (and possibly pipeline-version)

PR #386, opened 2025-07-22
Swap normal run workflow for lca-based run workflows

harmonbhasin:
This PR removes the LCA run workflows for the normal run workflows, getting rid of the old one as we will no longer do this anymore. This includes functionality for both the short and long read pipeline. The functionality for downstream and docs will come in a later PR (or previous depending on when this is committed).


harmonbhasin:
I have run ALL workflow level tests. The subworkflow and process tests are currently running.

tests/subworkflows/local/extractViralReadsONT/main.nf.test
```diff
@@ -22,37 +22,38 @@ nextflow_workflow {
 
     test("Should run without failures") {
         tag "expect_success"
-        tag "ont"
+        tag "ont_lca"
```
willbradshaw, line 25: Change to "ont" now that LCA is the default

tests/subworkflows/local/extractViralReadsONT/main.nf.test
```diff
@@ -62,7 +63,7 @@ nextflow_workflow {
     test("Should handle empty input file") {
         tag "empty_file"
         tag "expect_success"
-        tag "ont"
+        tag "ont_lca"
```
willbradshaw, line 66: Change to "ont"

configs/run.config
```diff
@@ -21,10 +21,7 @@ params {
     // Other required parameters
     n_reads_profile = 1000000 // Number of reads per sample to run through taxonomic profiling
     bt2_score_threshold = 20 // Normalized score threshold for calling a host-infecting virus read (typically 15 or 20)
-    cutadapt_error_rate = 0.33 // Maximum mismatch fraction between read and adapter for cutadapt trimming;
```
willbradshaw, line 24: Conversation with @harmonbhasin: this isn't doing anything b/c the parameter is hardcoded.
I'm pretty sure @katherine-stansifer meant for this to be configurable, so probably we should make it configurable rather than removing the param in the config file? @katherine-stansifer let me know your thoughts.

willbradshaw (changes requested):
Looks good, a couple minor things

configs/run.config
harmonbhasin: I've made this configurable.

willbradshaw (approved):

PR #387, opened 2025-07-22
Add clade counting to DOWNSTREAM

### LINKED ISSUE #374: Generate clade counts from LCA assignments
Several applications, including dashboards and reference-based growth detection, require counts of reads assigned to clades of the viral tree. There are two sub-tasks here:

1. Read the viral hits table and count reads assigned to each taxid by bowtie LCA
2. Aggregate counts up the tree

@katherine-stansifer and I talked today and decided that I should take this on since it's a blocker for RBGD. We decided on the following specs:

- Clade counting should run in DOWNSTREAM
- Counts should be generated per group
- Counts should be of deduplicated reads
- Initial implementation in python
- Output is a TSV with taxid, direct LCA assignment counts, and clade counts

dp-rice:
Resolves #374.

# Changes

- Add and test a nextflow module to run clade counting python script (#385)
- Add module to DOWNSTREAM workflow (directly rather than making a single-module subworkflow)
- Update docs
  - Describe module in DOWNSTREAM docs
  - Change description of DOWNSTREAM in the README to remove the comment that it only marks duplicates
- Update changelog and `pipeline-version.txt` to `v3.0.0.1-dev`

# Tests

I have added and run module-level tests:
```
nf-test test tests/modules/local/countReadsPerClade/main.nf.test
```

and updated and run the DOWNSTREAM workflow tests:
```
nf-test test tests/workflows/downstream.nf.test
```
I confirmed that the old output does not change, that the new output is correct, and updated the snapshot to include the new output file.

All tests complete successfully.


dp-rice:
@willbradshaw two questions for you when you review:

1. Is it right to make this a new version or should we roll these changes into `v3.0.0.0`?
2. I used `sample` in the nextflow module, following `markAlignmentDuplicates/main.nf`, but I think these properly refer to a sample group. I'm thinking that consistency between modules is more important at this stage, but maybe we should open an issue to clean up how we refer to these? Or maybe `sample` is the right way to think about it from the module's perspective and `group` is a concept that only applies when you're thinking about the higher level pipeline.

dp-rice:
It just occurred to me: should we be using the post-hoc validation output instead? We could take a taxonomic distance cutoff as a parameter and only count reads whose representative validates to within that distance. I think this is more complicated than we need for now, but it could be an option for an update.

modules/local/countReadsPerClade/main.nf
```diff
@@ -0,0 +1,19 @@
+// Count reads per clade using taxonomic LCA assignments
+// Note that if any reads are assigned to taxids not in the taxonomy, these reads will not be counted
+process COUNT_READS_PER_CLADE {
+    label "python"
+    label "single"
+    input:
+        tuple val(sample), path(reads_tsv) // read table with columns: seq_id, prim_align_dup_exemplar, aligner_taxid_lca
```
willbradshaw, line 7: Just checking: it can tolerate other columns, right? It just needs to have at least these columns?

(Ditto for taxdb)

pipeline-version.txt
```diff
@@ -1 +1 @@
-3.0.0.0-dev
+3.0.0.1-dev
```
willbradshaw, line 1: See above, no need to increment

docs/downstream.md
```diff
@@ -20,12 +20,15 @@ flowchart LR
...
+E --> K(Summary TSVs)
 E --> F[VALIDATE_VIRAL_ASSIGNMENTS]
 G(Viral taxonomy DB) --> F
 F --> H(Validation hits TSV)
 F --> I(BLAST results TSV)
-E --> J(Annotated hits TSVs)
-E --> K(Summary TSVs)
+G --> L[COUNT_READS_PER_CLADE]
+E --> L
+L --> M(Read count TSVs)
```
willbradshaw, line 31: Maybe "Clade count TSVs" for clarity?

tests/modules/local/countReadsPerClade/main.nf.test
```diff
@@ -0,0 +1,219 @@
...
+            
+            // Output and input files should exist
+            assert path(process.out.output[0][1]).exists()
+            assert path(process.out.input[0][1]).exists()
+            
+            // Parse the gzipped output TSV
+            def output_tsv = path(process.out.output[0][1]).csv(sep: "\t", decompress: true)
+            
+            // Should have 4 rows (header + 4 taxids: 0, 1, 2, 3)
+            assert output_tsv.rowCount == 4
```
willbradshaw, line 35: I don't love this kind of data-dependent hardcoding in tests. In this case you could just read the input and convert to a set, then find the size, right?

tests/modules/local/countReadsPerClade/main.nf.test
```diff
@@ -0,0 +1,219 @@
...
+            def rows_by_taxid = [:]
+            output_tsv.rows.each { row ->
+                rows_by_taxid[row.taxid] = row
+            }
+            
+            // Validate taxid 0 (root)
+            def row0 = rows_by_taxid[0]
+            assert row0 != null
+            assert row0.parent_taxid == "."
+            assert row0.reads_direct_total == 0
```
willbradshaw, line 51: My preference here would be to add expected-output columns to the taxdb file, which you can then read and compare to the observed output for each taxon.

tests/modules/local/countReadsPerClade/main.nf.test
```diff
@@ -0,0 +1,219 @@
...
+                '''
+                input[0] = Channel.of("test")
+                    | combine(Channel.of("${projectDir}/test-data/toy-data/count-reads-per-clade/test-reads-missing-seq-id.tsv"))
+                input[1] = "${projectDir}/test-data/toy-data/count-reads-per-clade/test-taxonomy.tsv"
+                '''
+            }
+        }
+        then {
+            assert process.failed
+            assert process.exitStatus == 1
```
willbradshaw, line 138: For expect_failed tests, I recommend asserting that the error message is what you expect; I've been bitten before by tests failing for other reasons but getting past my insufficiently-specific failure tests.

You can do this with `assert process.errorReport.contains("something")`

For extra credit, also add an `assert !process.errorReport.contains("something else")` to get around cases where the process just spits out the whole script.

willbradshaw (changes requested):
Core code looks fine, but I have some requests for the tests.

modules/local/countReadsPerClade/main.nf
dp-rice: Clarified

dp-rice: (the answer is yes)

tests/modules/local/countReadsPerClade/main.nf.test
dp-rice: I decided to address this and the next comment by included an expected output file in the toy data and using that for comparison.

tests/modules/local/countReadsPerClade/main.nf.test
dp-rice: See above. The advantage this has over adding to the taxdb file is that if we add other test cases we can add more expected output files more cleanly than adding more columns to taxdb

tests/modules/local/countReadsPerClade/main.nf.test
dp-rice: Good catch! I've:
1. Updated the python script to catch the KeyErrors and exit with an informative error message.
2. asserted that the error report contains the correct error message.
I don't think the negative test is necessary: because I'm using f strings for the error messages, and the message includes the input filename, the literal error message does not appear in the source code.

dp-rice:
Comments have been addressed. Tests have been rerun.

PR #405, opened 2025-08-06
Make clade counting use the NCBI root taxid

dp-rice:
The NCBI taxonomy uses 1 as the taxid for the root of the entire tree. It is also its own parent. Modify the clade counting code and tests to handle this. The key changes are:
- Set `parent_taxid` to 1 for root taxids in the viral tree (should just be `taxid` 1)
- Avoid creating a cycle when reading the root taxid pointing to itself 

Update snapshot for DOWNSTREAM nf test.

(Also makes a minor change to the nf-test toy data that tests a slightly more general tree: one that has a node with clade assignments but no direct assignments.)

# Tests

I have run the tests for the module and for downstream.

### DIFF
--- a/tests/workflows/downstream.nf.test.snap
+++ b/tests/workflows/downstream.nf.test.snap
@@ -3,13 +3,13 @@
         "content": [
             "gs1_duplicate_reads.tsv.gz:md5,3e2753611473e9bc77a036b060d6aca2",
             "gs1_duplicate_stats.tsv.gz:md5,48f655dffa173d1c8b23572ab2039e34",
-            "gs1_clade_counts.tsv.gz:md5,7939f8f7475146ff7ad4dfaa0ff70ba5",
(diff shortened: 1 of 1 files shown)


test-data/toy-data/count-reads-per-clade/test-reads.tsv
```diff
@@ -1,7 +1,7 @@
 seq_id	prim_align_dup_exemplar	aligner_taxid_lca	group
 read1	read1	1	test
 read2	read1	1	test
-read3	read3	2	test
+read3	read3	4	test
```
dp-rice, line 4: As above.

test-data/toy-data/count-reads-per-clade/test-taxonomy.tsv
```diff
@@ -1,4 +1,5 @@
 taxid	parent_taxid
-1	0
+1	1
 2	1
 3	1
+4	2
```
dp-rice, line 5: As above.

willbradshaw (approved):

PR #407, opened 2025-08-06
Update expected output files

dp-rice:
- Update the expected output files with new output
- Add a line to the developer docs reminding contributers to do this when submitting PRs


willbradshaw:
@dp-rice are you sure you have the right destination branch?

willbradshaw (changes requested):
One minor typo (I think). Agree on not including trace files.

PR #410, opened 2025-08-07
Merge staging back into dev (v3.0.0.0)

willbradshaw:


pipeline-version.txt
```diff
@@ -1 +1 @@
-3.0.0.0-dev
+3.0.0.0
```
dp-rice, line 1: Should be `3.0.1.0-dev` I think

test-data/gold-standard-results-downstream/gs1_clade_counts.tsv
dp-rice, line 1: When did these files get added to the repo?

dp-rice: Looks like
```
6fccb832 Removing prim_align from query related columns in the run short read workflow
```
Which was one of @harmonbhasin's fixes on staging. ✅

dp-rice (approved):
By eye, all of these changes look like the ones we made on the staging branch. I've also confirmed that staging and master are equal.

Double check that the version in `pipeline-version.txt` is what you want then good to go.

pipeline-version.txt
willbradshaw: Ah whoops thanks

PR #413, opened 2025-08-11
parameter maps

katherine-stansifer:


tests/modules/local/bbduk/bbduk.nf.test
```diff
@@ -35,12 +35,15 @@ nextflow_process {
             }
             process {
                 '''
+                params_map = [
```
willbradshaw, line 38: Huh, I didn't know you could do this. Why not in params?

subworkflows/local/blastFasta/main.nf
```diff
@@ -24,15 +24,25 @@ workflow BLAST_FASTA {
...
+                   // perc_id: Minimum %ID required for BLAST to return an alignment
+                   // qcov_hsp_perc: Minimum query coverage required for BLAST to return an alignment
+                   // blast_max_rank: Only keep alignments that are in the top-N for that query by bitscore
+                   // blast_min_frac: Only keep alignments that have at least this fraction of the best bitscore for that query
+                   // taxid_artificial: Parent taxid for artificial sequences in NCBI taxonomy
+                   // lca_prefix: Prefix for LCA column names (e.g. "blast")
     main:
-        // 0. Get reference paths
+        // Extract parameters from map
+        blast_db_prefix = params_map.blast_db_prefix
```
willbradshaw, line 37: FWIW I find immediately extracting all the map elements like this kind of ugly, and would rather have them passed directly to the processes instead. But it's not a big deal.

tests/modules/local/extractViralHitsToFastqNoref/main.nf.test
```diff
@@ -25,15 +25,18 @@ nextflow_process {
                 script "subworkflows/local/extractViralReadsShort/main.nf"
                 process {
                     '''
+                    def short_params = [
```
willbradshaw, line 28: Why is this a `def` when the others were undecorated definitions?

workflows/downstream.nf
```diff
@@ -36,12 +36,18 @@ workflow DOWNSTREAM {
         // 4. Generate clade counts
         COUNT_READS_PER_CLADE(dup_ch, viral_db)
         // 5. Validate taxonomic assignments
-        VALIDATE_VIRAL_ASSIGNMENTS(dup_ch, viral_db,
-            params.validation_cluster_identity, 15, params.validation_n_clusters,
-            params.ref_dir, params.blast_db_prefix,
-            params.blast_perc_id, params.blast_qcov_hsp_perc,
-            params.blast_max_rank, params.blast_min_frac,
-            params.taxid_artificial)
+        validation_params = [
```
willbradshaw, line 39: I think you could just set a value for `params.cluster_min_len` in the workflow and then pass `params` to `VALIDATE_VIRAL_ASSIGNMENTS`?

willbradshaw: (Ditto for other workflows)

willbradshaw (approved):
Minor comments only, assuming all affected tests pass.

tests/modules/local/bbduk/bbduk.nf.test
katherine-stansifer: huh, I didn't know you could do it the other way! params makes more sense; moving it there.

tests/modules/local/extractViralHitsToFastqNoref/main.nf.test
katherine-stansifer: oops, that's a claude-codeism, removed

katherine-stansifer: however I'm not sure how to make moving to a params block work for setup--the naive thing of just making a params {} block under run before the process {} block gives errors--so I've left this in process. happy to discuss tomorrow!

willbradshaw:
@katherine-stansifer a test is failing

willbradshaw (approved):

willbradshaw:
@katherine-stansifer back to you for squash-merge

PR #423, opened 2025-08-16
Staging

katherine-stansifer:
Full test suite passes; currently running test datasets through RUN and DOWNSTREAM, but I think you can review in parallel.  I've discussed merge strategy with Dan already!


CHANGELOG.md
```diff
@@ -1,5 +1,29 @@
-# v3.0.0.1
+# v3.0.1.0
```
willbradshaw, line 1: Could we try writing this with a user rather than developer in mind, like the 3.0.0.0 update?

katherine-stansifer: done, take a look

katherine-stansifer:
Re-wrote CHANGELOG; also added a couple comments to configs/profiles.config (I realized there were some notes in changelog that make more sense there as persistent comments) 
Small test datasets run successful, so I think after your re-review this is finally ready to merge

PR #429, opened 2025-08-19
Fixing a bug in prepareGroupTsv

harmonbhasin:
Bug: VALIDATE_GROUPING defines `tuple val(label), path("zero_vv_${input_file}"), emit: zero_vv_log` as one of the outputs, then these output files then get concatenated in PREPARE_GROUP_TSVS: 

```
zero_vv_log_ch = VALIDATE_GROUPING.out.zero_vv_log
// 7. Concatenate zero VV logs into single consolidated file
zero_vv_log_grouped_ch = zero_vv_log_ch.map { label, file -> ["all_samples", file] }.groupTuple()
consolidated_zero_vv_ch = CONCATENATE_EMPTY_SAMPLES_LIST(zero_vv_log_grouped_ch,"zero_vv_consolidated").output
```

For the other subworkflows in downstream, we use `test-data/downstream/input_file.csv`, which has the hits_tsv defined as 
`s3://nao-testing/gold-standard-test/processed/virus_hits_final.tsv.gz` for the three groups. This leads to a file clash when trying to run CONCATENATE_EMPTY_SAMPLES_LIST as each channel in `zero_vv_log_ch` will have the same file name.

Solution: Add the label name to the file, such that each file is unique:

`tuple val(label), path("zero_vv_${label}_${input_file}"), emit: zero_vv_log`


harmonbhasin:
Verified that the main downstream test still passes, and that all the tests that use PREPARE_GROUP_TSVS pass by running this command ([rg](https://github.com/BurntSushi/ripgrep) is a newer version of grep):

`rg "PREPARE_GROUP_TSVS" -l | rg "tests" | xargs -P 5 -I {} nf-test test {}`

willbradshaw (approved):

PR #431, opened 2025-08-20
Adding official formatting for ruff, and mypy through the inclusion of a pyproject.toml

harmonbhasin:
The pyproject.toml will allow us to specify the settings that are used for ruff, mypy, and pytest across all python files in this repo. 

I additionally updated the developer docs to note this, and mentioned that ruff formatting should be checked as we ran into an issue where ruff was setting the [`strict` attribute to false by default if it was not specified in `zip`](https://github.com/naobservatory/mgs-orchestrator/pull/47#discussion_r2261190503).

I ran pytest, ruff, and mypy and they all worked. There were issues (by issues I mean they both said that linting/typing was violated) that came up with ruff, and mypy; however, updating all these files might take a lot of time, so I'd figure we can just have people run `ruff` and `mypy` on scripts as we update them. If you would like me to just run it now and update all the python scripts, I can also do that.

### DIFF
--- a/docs/developer.md
+++ b/docs/developer.md
@@ -59,7 +59,9 @@ These guidelines represent best practices to implement in new code, though some 
 - Python style: 
     - Loosely follow PEP 8 conventions.
     - Type hints are encouraged but not currently required.
-    - Linting is encouraged (our go-to tool is `ruff`), but not currently required. 
+    - Linting and type checking are encouraged (our go-to tools are `ruff` for linting and `mypy` for type checking), but not currently required.
+    - Formatting applied (including in nested directories) will follow the configuration in `pyproject.toml`.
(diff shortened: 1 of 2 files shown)


harmonbhasin:
For future reference, these settings for ruff and mypy were pulled from [mgs orchestrator](https://github.com/naobservatory/mgs-orchestrator/blob/dev/pyproject.toml).

willbradshaw (dismissed):

willbradshaw:
> There were issues (by issues I mean they both said that linting/typing was violated) that came up with ruff, and mypy; however, updating all these files might take a lot of time

@harmonbhasin can you make an issue for doing this? I'd especially like to get to the point where Mypy is happy.

willbradshaw (approved):

PR #432, opened 2025-08-20
Adding preference for using pytest for processes that use python over nf-test for unit tests

harmonbhasin:
Pytest tests are faster, easier to write, and allow more complicated logic than nf-test. For unit tests, these will now be preferred.

### DIFF
--- a/docs/developer.md
+++ b/docs/developer.md
@@ -70,7 +70,7 @@ Build and push the custom containers using the script `bin/build-push-docker.sh`
     
 ## Testing
 
-All tests use the [nf-test](https://www.nf-test.com/) framework. (We do not currently have any unit tests of Python/Rust/R scripts.)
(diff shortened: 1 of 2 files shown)


willbradshaw (approved):

docs/developer.md
```diff
@@ -72,7 +72,7 @@ Build and push the custom containers using the script `bin/build-push-docker.sh`
     
 ## Testing
 
-All tests use the [nf-test](https://www.nf-test.com/) framework. (We do not currently have any unit tests of Python/Rust/R scripts.)
+We use [nf-test](https://www.nf-test.com/) for integration and end-to-end tests. For Python process unit tests, we use pytest. As we continue developing, we will transition away from nf-test unit tests for Python scripts.
```
willbradshaw, line 75: I think this needs to be refined a little to reflect current reality. Currently, we use nftest for all tests for nearly all modules. We intend to transition to using pytest, and developers should use pytest going forward for unit tests, but saying "for unit tests, we use pytest" isn't really true.

willbradshaw (changes requested):

willbradshaw (approved):

Issue #438, opened 2025-08-28
Bugfix in DOWNSTREAM to allow viral hits table to have samples not found in the grouping file


harmonbhasin:
In the latest release, 3.0.1.0, I incorrectly introduced logic to the downstream workflow that required the samples in the viral hits table to be a subset of samples in the grouping table. This logic is problematic as we want to be able to exclude undetermined (non-demultiplexed) samples.

PR #439, opened 2025-08-28
Fix bug that prevents DOWNSTREAM workflow from running when the viral hits table has samples that are not in the grouping file.

harmonbhasin:
Addressing #438. Instead of raising an error when a sample is found in the viral hits table, and not in the grouping file, we now log a  statement for debugging. I also updated the tests to reflect this new functionality. 

The validateGrouping test works, and so do all tests that use prepareGroupingTsv.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # v3.0.1.1-dev
 - Updated `docs/developer.md` to reflect the new branch naming convention.
 - Added pyproject.toml at the top level directory to set ruff, and mypy settings.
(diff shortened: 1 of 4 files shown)


harmonbhasin:
Index test was failing due to a genome failing to download, running the test again to see if the issue resolves itself.

willbradshaw (changes requested):
Code looks fine. Watch out for typos.

willbradshaw (dismissed):

willbradshaw (approved):

Issue #192, opened 2025-02-11
Check compatible taxids across index tools


willbradshaw:
The INDEX workflow has several components that use the NCBI taxonomy database, including Kraken, BLAST and Virus-Host DB. If these use incompatible versions of the taxonomy structure, it could cause downstream issues. Currently we have no way to avoid this except carefully checking that each input to the database was created at a similar time; it would be better if we could automatically check the taxonomies of the different components and confirm that they are compatible.

PR #445, opened 2025-09-05
Update Index urls and ignored viral genomes

harmonbhasin:
Addresses #192. By updating the NCBI taxonomy and Virus-Host DB to use the daily release, we can reduce the inconsistency between tools vastly. Although Kraken2 will still remain out of sync, it is not used in identifying viral hits, so we are fine with it being a bit of date. More information about this solution can be found in a [private NAO internal doc](https://docs.google.com/document/d/1zMkYxXNzsW5_OBRyK3P_6pw2pzw6M-rMB3K2J4rRKeI/edit?usp=sharing).


willbradshaw:
What about `index-for-run-test.config`?

Issue #448, opened 2025-09-15
VALIDATE_GROUPING tsv for samples with no viral hits is missing header.


harmonbhasin:
As the title says.

willbradshaw:
@harmonbhasin remember to add tags to your issues

PR #449, opened 2025-09-15
Fix `VALIDATE_GROUPING` bug where headerless log files can't be concatenated as the concatenation script expects headers.

harmonbhasin:
Addresses #448. 

In some DOWNSTREAM runs, we were getting errors from CONCATENATE_EMPTY_SAMPLES_LIST which concatenates the log files from VALIDATE_GROUPING.  CONCATENATE_EMPTY_SAMPLES_LIST requires there to be a header; however, the log files from VALIDATE_GROUPING didn't have headers, which is why this code failed. The test didn't catch this as the test data doesn't have samples with no viral hits.


harmonbhasin:
Local tests for this process pass.

willbradshaw (dismissed):

PR #454, opened 2025-09-18
Merge dev to main -- Release 3.0.1.2

harmonbhasin:
All local tests passed, and the composite dataset ran without error.


bin/test_component_dependencies.py
dp-rice, line 1: If the changes to this file are covered by the changelog, it's not obvious to me.

pyproject.toml
```diff
@@ -1,6 +1,14 @@
...
 name = "mgs-workflow"
+version = "3.0.1.2"
 requires-python = ">=3.12"
+dependencies = [
+    "biopython>=1.85",
+    "mypy>=1.18.1",
+    "pandas>=2.3.2",
+    "pytest>=8.4.2",
+    "ruff>=0.13.0",
+]
```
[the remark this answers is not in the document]
harmonbhasin: Given that the `pyproject.toml` isn't used for anything other than testing code, I thought it would be fine to put the dependencies in the main section so that users have less to type to get their env working. Let me know what you think.

bin/test_component_dependencies.py
dp-rice: Sorry I missed that on changelog PR review!

harmonbhasin:
> Probably we should move towards recording the version only in pyproject.toml; We'll need to make changes to orchestrator before we change this. @harmonbhasin can you make an issue for that in both repos?

Done see #459, which also links to the orchestrator issue (not going to link it here as that's a private repo).

willbradshaw (approved):

Issue #455, opened 2025-09-18
Release 3.0.1.2 -- CHANGELOG update


harmonbhasin:
Updating the CHANGELOG.md before merging dev into main.

harmonbhasin:
@willbradshaw yes.

PR #456, opened 2025-09-18
Addressing Reviewer Comments for CHANGELOG.md for Release 3.0.1.2

harmonbhasin:
Addressses #455 

Updated based on @dp-rice comments in #454. The other thing I did was fix random spacing added in during #291. Now if you diff the changelog with main, the only difference is the new section.


harmonbhasin:
I think that would be good, made an issue #457.

PR #471, opened 2025-10-07
Remove per-line logging in filter_virus_reads.py

### LINKED ISSUE #469: Heavy logging in FILTER_VIRUS_READS causing AWS Batch jobs to hang
On version 2.9.0.6, which is currently `stable`, the process FILTER_VIRUS_READS sometimes causes its AWS Batch job to run indefinitely even though the nextflow process completes in ~5m.

I've tracked this issue down to excessive logging. The process's python script writes a line to stdout for every line of input, sometimes generating over 100MB of output. On batch, this gets captured and sent to the CloudWatch logs. If the process exits before the log buffer is flushed, the job never gets marked as complete.

Here's a minimal workflow that reproduces the issue:

```
process LOG_SPAM {
    container 'ubuntu:latest'

    output:
    path 'output.txt'

    script:
    """
    # Generate ~100MB of stdout (1 million lines × ~100 bytes each)
    for i in {1..1000000}; do
        echo "Log line \$i: Simulating verbose process output to reproduce AWS Batch CloudWatch streaming issue"
    done 

    # Write some minimal output
    echo "Hello" > output.txt

    echo "Process completed successfully"
    """
}

workflow {
    LOG_SPAM()
}
```

To fix it, we just need to delete logging lines in the python script.

dp-rice:
Resolves #469 by deleting the logging statements for each import line.

I've verified that it fixes the problem in a minimal example with full-sized data ([private repo](https://github.com/securebio/nao-debug-filter-virus-reads)). Currently re-running a failing run to make sure that it solves the problem in production.

Updates version to 2.9.0.7. This is a point release because it doesn't change anything except logging.

## Testing

Ran:
```
nf-test test tests/modules/local/filterVirusReads/main.nf.test
```
All tests pass.


willbradshaw (approved):
Approved. For for the 3.0.1 PR I'd prefer to move logging to debug rather than deleting it entirely, but as discussed I think it's fine here to just delete.

PR #510, opened 2025-11-14
Convert concatenateTsvs to Rust

### LINKED ISSUE #480: Concatenation processes are slow
I haven't measured this carefully, but from running the pipeline a bunch since I've taken over delivery processing, I've noticed that processes that concatenate files seem to be a bottleneck that contribute significantly wall clock time. For example, my current run was stuck here for a while:

```
[73/8f525d] process > RUN:EXTRACT_VIRAL_READS_SHORT:PROCESS_LCA_ALIGNER_OUTPUT:CONCATENATE_ALIGNER              [  0%] 0 of 1, retries: 2
[82/cb1e55] process > RUN:EXTRACT_VIRAL_READS_SHORT:PROCESS_LCA_ALIGNER_OUTPUT:CONCATENATE_VIRAL_HITS           [  0%] 0 of 1, retries: 2
```

Looks like `CONCATENATE_VIRAL_HITS` took 22 minutes to complete, but it also failed a few times when spot instances got reclaimed, so took more like 60 min of wall clock time.

`CONCATENATE_ALIGNER` has been in progress with various restarts for almost two hours.

I'm guessing we could improve our end-to-end time by making these faster. I haven't looked at whether these are the same module or not. See also #342.

jo-delia-faraguna:
Goal was to resolve #480. Explored:

1. Python: multi-threaded file reading with a queue for a single file writer.
2. Python: single-threaded chunked file reading and writing.
3. Bash: bare `head` / `tail` commands (only on uncompressed files).

Selected this Rust implementation which generally improves across different input conditions.


willbradshaw:
Given how small the speedup from switching to Rust ended up being, I think we should explore other options before making this change. I'm going to close it for now, but we can revisit later.

PR #520, opened 2025-11-19
Replace more nf-test tests with pytest

willbradshaw:
# Description

This PR replaces nf-test tests with 1-to-1 replacements in Pytest, to help streamline execution of the test suite. The bulk of the code was written by Claude Code (prompt below). In total, 85 tests were converted. I also added a Github Actions configuration for running Pytest on pull requests.

I was quite strict in this PR about keeping Claude to 1:1 replacements for existing tests, rather than adding new tests it thought were useful. Adding marginal Pytest tests is _much_ cheaper than marginal nf-test tests, however, so it likely makes sense to add more detailed tests in a future PR.

# Prompt

> Many of the modules in this repository are thin wrappers around Python scripts. Currently, most of these are tested using nf-test, which is slow and clunky. I want to convert most of these tests to run with Pytest instead. Please do the following:
> 
> 1. Scan through `modules/local/moduleName/resources/usr/bin` directories to identify modules that have an associated Python script (`module_name.py`) but lack a corresponding Pytest file (test_module_name.py) in the same directory.
> 2. For each of these modules, look for nf-test test files in `tests/modules/local/moduleName` directories.
> 3. Write a Pytest file in the same directory as the Python script that covers only and exactly the test cases covered by the nf-test file (we might want to add other tests later, but for now we just want to make sure we don't lose coverage).
> 4. Run the tests and check they pass.
> 5. Edit the nf-test file to delete the tests that were transferred, leaving only a minimal end-to-end crash test for each module.
> 
> In writing the Pytest files, you should use fixtures and pytest.mark.parameterize to reduce redundancy and improve readability. Some common fixtures can be found at `modules/local/conftest.py`; you're encouraged to use these and add to them where appropriate. You should avoid introducing test cases that aren't covered by the original nf-test file; we'll add these at a later time.
> 
> There are a number of modules to address here; unless I tell you otherwise, you should keep going until all of them have been converted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # v3.0.1.5-dev
 - Fixed `SPLIT_VIRAL_TSV_BY_SELECTED_TAXID` failing when input TSV has no data rows. Empty partition files (`partition_empty_*`) are now filtered out before downstream processing. (#524)
+- Converted many nf-test tests to Pytest to ease pre-release review.
 
 # v3.0.1.4
 - Removed `nextflow.preview.output` statement from `main.nf` for compatibility with Nextflow 25.10


pyproject.toml
dp-rice: Why these changes? Is it for GHA deployment?

PR #530, opened 2025-11-26
Merge dev into main

willbradshaw:
Merging main into dev so we can complete the release


evanfields:
@willbradshaw The target branch here is `dev` not `main`?

PR #532, opened 2025-11-30
Fix nf-test load failures on Github Actions

### LINKED ISSUE #388: Add a test retry for Github Actions if it fails to download `nf-test`
Almost every time I've run tests in the past week, I've had one test fail instantly as the Github Actions instance was unable to download `nf-test` during the setup process. It would be nice to configure Github Actions such that it retries a test if it fails at a specific step, or of it fails within a specific period of time.

jeffkaufman:
Use the official Github Action to install nf-test https://github.com/nf-core/setup-nf-test

This should solve the problem where we keep getting download errors.

Fixes #388

### DIFF
--- a/.github/workflows/end-to-end.yml
+++ b/.github/workflows/end-to-end.yml
@@ -36,14 +36,8 @@ jobs:
         with:
           version: "25.10.0"
 
-      - name: Install nf-test
-        run: |
-          for i in 1 2 3 4 5; do
-            echo "Attempt $i to download nf-test..."
-            wget -qO- https://get.nf-test.com | bash && break
-            sleep 10
-          done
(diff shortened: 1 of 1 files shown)


jeffkaufman:
@simonleandergrimm FYI

jeffkaufman:
@willbradshaw Sorry, I forgot about this!  Would you like me to close and re-open, or is this a "for next time" request?

Is this also something you'd like for the other repos (and belongs in https://data.securebio.org/handbook/ ?) or is this only an mgs-workflow convention?

willbradshaw:
@jeffkaufman noting for future reference that we both forgot to add a CHANGELOG entry for this PR.

Issue #534, opened 2025-12-01
csv.writer in filter_tsv_column_by_value.py adds erroneous quotation marks to quality scores.


simonleandergrimm:
It'd be good to hot-fix this. Looks like `csv.writer` is only used in `filter_tsv_column_by_value.py`.

## Original discovery

When running `DOWNSTREAM`, I got an error in `VALIDATE_VIRAL_ASSIGNMENTS:CLUSTER_VIRAL_ASSIGNMENTS:VSEARCH_CLUSTER`, which complained about sequence and quality lines not having the same length:

```
  Reading file af3c57b480b1e07e8fc60fe0b71b81_3052619_hits_out.fastq.gz 100%
  2509 nt in 4 seqs, min 623, max 630, avg 627
  Masking 100%
  Sorting by length 100%
  Counting k-mers 100%
  Clustering 100%
  Sorting clusters 100%
  Writing clusters 100%
  Clusters: 1 Size min 4, max 4, avg 4.0
  Singletons: 0, 0.0% of seqs, 0.0% of clusters
  vsearch v2.30.0_linux_x86_64, 369.8GB RAM, 96 cores
  https://github.com/torognes/vsearch
  Reading file af3c57b480b1e07e8fc60fe0b71b81_3418604_hits_out.fastq.gz
  Fatal error: Invalid line 33156 in FASTQ file: Sequence and quality lines must be equally long
```

Looking into this, the underlying read in the original `virus_hits_final.tsv.gz` did indeed show a difference in the length of the sequence and the quality score. 

## Origin of the bug

Checking the diff between the overly long quality score and the original quality score, I get an introduction of 14 `”` characters, one at the start and end, the others added to a section of `"` in the middle.

Reviewing the RUN workflow with Claude Code, the bug originates in `filter_tsv_column_by_value.py`, used by modules/local/filterTsvColumnByValue.

As per Claude, by default csv.writer uses the `QUOTE_MINIMAL` setting. This in turn wraps fields containing `"` in double quotes and doubles all internal `"` characters.

Anyway, I tested this, and it seems to be right. Running the string through the script and checking length before and after, I get added `”` chars:

```
% ./filter_tsv_column_by_value.py   -i test-bug.tsv \
    -o test-bug-output.tsv \
    -c filter_col \
    -v true
[...]
% awk -F'\t' '{print length($3)}' test-bug.tsv
10
1312
% awk -F'\t' '{print length($3)}' test-bug-output.tsv
10
1326
[...]

PR #535, opened 2025-12-01
Handle quote characters in filterTsvColumnByValue

willbradshaw:
- Addresses #534 by treating quotation characters as regular characters rather than delimiters.
- Adds a test to the relevant Pytest file to catch regressions in future.
- Note that, currently, only this module uses the `csv` library; if we add it to other modules in the future we'll need to include similar safeguards.
- Once this is merged I'll cherrypick a hotfix onto stable.


katherine-stansifer (approved):
Looks great!

PR #551, opened 2025-12-18
Move ADD_CONDITIONAL_TSV_COLUMN to Python

willbradshaw:
Python implementation is more readable and allows us to move unit tests to Pytest.


dp-rice (approved):
Converting from inline script to python seems like an improvement.

My big-picture question (not necessary to merge this PR) is why we need this process it at all. It seems like a lot of boilerplate for something really trivial. Could the logic be merged into an upstream or downstream process?

PR #559, opened 2025-12-25
Switch containers from Seqera Containers to AWS ECR

willbradshaw:
We've repeatedly run into issues with Seqera Containers pull limits, especially while testing. Fixing this is a necessary but not sufficient condition for streamlining our release process. This PR switches the codebase over from Seqera Containers to AWS ECR, while still keeping programmatic container generation from YAML files. The whole test suite passes, and I was able to run the small benchmarking dataset on Batch with the new containers.

Importantly, while this PR disables Seqera Wave when running locally (e.g. for tests) it remains enabled when running on Batch, as it's needed for the Fusion filesystem we use to work with S3. In the process of working on this PR, I discovered that, if Wave is enabled, the containers are passed through Seqera Containers even if they are hosted on ECR, so we can still run into pull rate limits when running on Batch. Fortunately this hasn't been a big problem so far; if it becomes one in the future, we'll need to move off Fusion in order to fully disentangle from Wave.

### DIFF
--- a/bin/build_ecr_container.py
+++ b/bin/build_ecr_container.py
@@ -216,7 +216,7 @@ def build_docker_image_from_spec(
     logger.info(f"Building Docker image: {image_tag}")
     try:
         subprocess.run(
-            ["docker", "build", "-t", image_tag, str(build_dir)],
+            ["docker", "build", "--platform", "linux/amd64", "-t", image_tag, str(build_dir)],
             check=True,
         )
         logger.info(f"Built image: {image_tag}")


willbradshaw:
Note that the Trivy test is still expected to fail, see notes in #476 for details.

bin/build_ecr_container.py
```diff
@@ -0,0 +1,449 @@
...
+    spec_filename = spec_file.name
+    shutil.copy(spec_file, build_dir / spec_filename)
+    dockerfile_path = build_dir / "Dockerfile"
+    dockerfile_path.write_text(generate_dockerfile(spec_filename))
+    logger.info(f"Building Docker image: {image_tag}")
+    try:
+        subprocess.run(
+            ["docker", "build", "-t", image_tag, str(build_dir)],
+            check=True,
+        )
```
dp-rice: Claude pointed out that we lost the platform specification

```suggestion
        subprocess.run(
            ["docker", "build", "--platform", "linux/amd64", "-t", image_tag, str(build_dir)],
            check=True,
        )
```

PR #568, opened 2026-01-08
Update tests to use local tiny index

willbradshaw:
This PR updates the test suite to actually use the new tiny test datasets generated in #567, and deletes references to the old gold-standard test set, which is no longer used.

In some cases, small changes to the actual modules and workflows was required to get everything working.

This change reduces the runtime of the entire test suite from >4500 seconds to ~2800 seconds.

There remain a lot of cheap opportunities to save runtime on the test suite, but I'm leaving those to a future PR; this one solely makes the changes required for the whole suite to run successfully with the tiny datasets.

### DIFF
--- a/tests/configs/index.config
+++ b/tests/configs/index.config
@@ -10,7 +10,7 @@ params {
     // Directories
     base_dir = "./" // Parent for working and output directories (can be S3)
 
-    // Cow & E. coli
+    // Other contaminants (phage)
     genome_urls = [
         phage: "https://eutils.ncbi.nlm.nih.gov/entrez/eutils/efetch.fcgi?db=nucleotide&id=NC_000866.4&rettype=fasta&seq_start=1&seq_stop=1000"
     ]
@@ -18,7 +18,7 @@ params {
     // URLs for downloading reference genomes etc
     taxonomy_url = "https://nao-testing.s3.amazonaws.com/test-databases/tiny-taxonomy.zip"
     virus_host_db_url = "https://www.genome.jp/ftp/db/virushostdb/virushostdb.daily.tsv"
-    human_url = "https://eutils.ncbi.nlm.nih.gov/entrez/eutils/efetch.fcgi?db=nucleotide&id=NC_000021.9&rettype=fasta&seq_start=10000000&seq_stop=10001000"
+    human_url = "https://eutils.ncbi.nlm.nih.gov/entrez/eutils/efetch.fcgi?db=nucleotide&id=NC_000021.9&rettype=fasta&seq_start=10000000&seq_stop=10001000" // 1kb of human chr21
     ssu_url = "https://www.ebi.ac.uk/ena/browser/api/fasta/AB065370.1"
     lsu_url = "https://www.ebi.ac.uk/ena/browser/api/fasta/AB065370.1"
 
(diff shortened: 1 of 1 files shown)


tests/modules/local/minimap2/align.nf.test
```diff
@@ -73,20 +73,6 @@ nextflow_process {
 
             // Verify input FASTQ contains all reads
             assert input_reads == fastq_read_ids_mapped + fastq_read_ids_unmapped
-
```
willbradshaw: These are tests for multiple mapping that are no longer fulfilled by the new test data, because it's too small. I don't think this was especially important to test for before, and certainly isn't worth artificially inflating our test set now.

test-data/results/run-short/subset_qc_adapter_stats.tsv
katherine-stansifer: same question as above, is it fine that this file is empty?

tests/subworkflows/local/splitViralTsvBySelectedTaxid/main.nf.test
```diff
@@ -120,6 +120,7 @@ nextflow_workflow {
 
     test("Should run without failures on valid input data") {
         tag "expect_success"
+        tag "valid_data"
```
katherine-stansifer: how is this tag used? is there anywhere else we need to add it?

tests/subworkflows/local/qc/main.nf.test
```diff
@@ -63,9 +63,11 @@ nextflow_workflow {
             // Other TSVs should have expected structure and content
             def tsv_adapt = path(workflow.out.qc[0][1]).csv(sep: "\t", decompress: true)
             with (tsv_adapt) {
-                assert columnCount == 6
-                assert columns["stage"][0] == stage
-                assert columns["sample"][0] == sample
+                assert columnCount == 6 || columnCount == 0
```
katherine-stansifer: again, if you keep the empty adapter QC files I'd add a comment in every test file where this comes up--it will help our future selves remember what was going on.

tests/subworkflows/local/blastFasta/main.nf.test
```diff
@@ -272,7 +188,8 @@ nextflow_workflow {
                     lca_prefix: params.lca_prefix,
                     db_download_timeout: params.db_download_timeout
                 ]
-                input[0] = CONVERT_FASTQ_FASTA.out.output
+                input[0] = Channel.of("test")
```
katherine-stansifer: I really like this simplification!

tests/subworkflows/local/processLcaAlignerOutput/main.nf.test
```diff
@@ -155,7 +155,7 @@ nextflow_workflow {
                         group_field: "seq_id",
                         taxid_field: "taxid",
                         score_field: "length_normalized_score",
-                        taxid_artificial: "12908",
+                        taxid_artificial: "81077",
```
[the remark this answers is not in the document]
willbradshaw: It was wrong before! https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?command=show&mode=node&id=81077&lvl=

tests/subworkflows/local/qc/main.nf.test
willbradshaw: You're right, I think there's a better way to handle this. Will change.

test-data/results/run-ont/subset_qc_adapter_stats.tsv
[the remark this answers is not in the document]
willbradshaw: It would be better to simulate adapter contamination in our test reads, but talking to Claude Code, this seems to be nontrivially complex. I think this is okay.

tests/modules/local/replaceString/replace_string.nf.test
```diff
@@ -13,8 +13,8 @@ nextflow_process {
 
         when {
             params {
-                old_string = "gs3"
-                new_string = "gs4"
+                old_string = "tt3"
```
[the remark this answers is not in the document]
willbradshaw: It's not essential to change them, but the old name was a reference to "gold standard" which isn't a good description of the new test data.

tests/configs/index.config
```diff
@@ -12,19 +12,15 @@ params {
 
     // Cow & E. coli
```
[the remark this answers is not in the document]
willbradshaw: Will do

Issue #560, opened 2025-12-26
Cut down test suite runtime


willbradshaw:
As of #559 we can now run the entire test suite with one command, but it still takes over an hour to run (4680s). It would be a lot more convenient to prepare new releases if we could run the test suite more quickly.

PR #570, opened 2026-01-10
Remove setup statements from subworkflow tests

willbradshaw:
Continuing with the work of #560, this test removes `setup` statements from all the tests in `tests/subworkflows`, replacing them with plaintext data from `test-data`. This ended up a bigger task than initially planned, because fully removing the `setup` blocks required converting various modules to handle plaintext inputs rather than just gzipped ones -- which in turn required changing some module tests to fit the new behavior. (I think this is a good change, but it undeniably increased the scale of this PR.)

In total, this reduced total test suite runtime from ~2400 to ~2100 seconds (i.e. about 5 minutes quicker).

There's certainly more that could be done to reduce runtimes here, including removing `setup` blocks from module tests and converting more nf-test tests to pytest. I might do some of this as easy work while I'm sick. But all else equal, I suspect at this point it's a better use of time to work on parallelization and automation rather than directly tinkering with tests.


tests/modules/local/bbduk/bbduk_hits.nf.test
katherine-stansifer: claude points out: why do you still need the setup block in this file?

modules/local/convertFastqFasta/main.nf
```diff
@@ -27,21 +29,41 @@ process CONVERT_FASTQ_FASTA_LIST {
...
-        tuple val(sample), path("${sample}_*_in.fastq.gz"), emit: input
-    shell:
-        '''
-        for fastq in !{fastqs}; do
-            species=$(basename ${fastq} | grep -oP '!{sample}_\\K\\d+(?=_)')
-            if [ -z "$species" ]; then
-                >&2 echo "Error: Could not extract species from filename: ${fastq}"
+        tuple val(sample), path("converted_*"), emit: output
+        tuple val(sample), path("input_*"), emit: input
+    script:
```
katherine-stansifer: in the future I'd consider making something like this its own PR

modules/local/convertFastqFasta/main.nf
```diff
@@ -27,21 +29,41 @@ process CONVERT_FASTQ_FASTA_LIST {
...
+        def inputs = fastqs.collect { fastq -> "input_${fastq}" }
+        """
+        fastqs_array=(${fastqs})
+        outputs_array=(${outputs.join(" ")})
+        inputs_array=(${inputs.join(" ")})
+        for ((i=0; i<\${#fastqs_array[@]}; i++)); do
+            fastq="\${fastqs_array[i]}"
+            output="\${outputs_array[i]}"
+            input="\${inputs_array[i]}"
+            species=\$(basename \${fastq} | grep -oP '${sample}_\\K\\d+(?=_)')
```
katherine-stansifer: I think species is unused?

test-data/prepareGroupTsvs/tt1_groups.tsv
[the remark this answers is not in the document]
willbradshaw: Nextflow requires unique filenames for all the inputs to a process. We could have a single input and then use a setup statement to copy it to a different filepath, but I'm trying to get away from using setup statements for subworkflow tests.

modules/local/convertFastqFasta/main.nf
```diff
@@ -27,21 +29,41 @@ process CONVERT_FASTQ_FASTA_LIST {
...
-                >&2 echo "Error: Could not extract species from filename: ${fastq}"
+        tuple val(sample), path("converted_*"), emit: output
+        tuple val(sample), path("input_*"), emit: input
+    script:
+        // Extract extension from first file and check that all files have the same extension
+        def first_file = fastqs[0].toString()
+        if (!first_file.contains('.')) {
+            throw new Exception("Input file ${first_file} has no extension")
+        }
+        def extension = first_file.substring(first_file.indexOf('.') + 1)
```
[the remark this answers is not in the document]
willbradshaw: The intent is to catch multipart extensions like `.fastq.gz`, which I believe this achieves?

modules/local/convertFastqFasta/main.nf
willbradshaw: Fair enough!

modules/local/subsetReads/main.nf
```diff
@@ -126,31 +130,34 @@ process SUBSET_READS_SINGLE_TARGET {
...
-        tuple val(sample), path("${sample}_subset.${suffix}.gz")
-    shell:
-        '''
-        # Define input/output
-        in=!{reads}
-        out=!{sample}_subset.!{suffix}.gz
+        tuple val(sample), path("subset_${reads}"), emit: output
+        tuple val(sample), path("input_${reads}"), emit: input
+    script:
+        def in = reads
```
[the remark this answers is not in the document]
willbradshaw: Good point, will fix

Issue #494, opened 2025-11-04
DOWNSTREAM skipping samples with no vertebrate viral reads


jeffkaufman:
I have some water controls which don't have any viral reads.  These result in no DOWNSTREAM output, but empty output would be better.  Otherwise there's no good way for me to distinguish "DOWNSTREAM didn't complete" from "DOWNSTREAM completed but these files were empty".

See https://twist.com/a/197793/ch/789673/t/7367576/c/99459800

dp-rice:
I'm trying to restart the backlog run and I'm running into a consequence of this. I started by using mgs-orchestrator to submit deliveries from the last month to make sure I haven't missed any and it started trying to run DOWNSTREAM on water samples.

PR #605, opened 2026-01-24
Fix DOWNSTREAM empty group bug

### LINKED ISSUE #526: PARTITION_TSV uses hardcoded 'empty' string that can be misinterpreted as data
## Problem

When `PARTITION_TSV` receives a TSV with headers but no data rows, it creates an output file with the hardcoded string `"empty"` as the partition identifier:

```python
# partition_tsv.py lines 55-58
if fields is None: # Empty apart from headers
    outf = initialize_output_file(input_path, "empty", headers)
    outf.close()
    return
```

This creates a file named `partition_empty_<input_filename>.tsv.gz`.

## Why this is problematic

1. **Collision with legitimate data**: If a partition column actually contained `"empty"` as a value, it would be indistinguishable from the no-data case.

2. **Misinterpreted downstream**: In `PREPARE_GROUP_TSVS`, group names are extracted from partition filenames using regex:
   ```groovy
   def pattern = "^partition_(.*?)_sorted_group_..."
   [matcher[0][1], path]  // extracts group name from filename
   ```
   When `partition_empty_...` is created, `"empty"` gets extracted as a group name and flows through the pipeline as if it were a real group.

3. **Not clearly a sentinel value**: The string `"empty"` doesn't clearly indicate it's a special marker rather than real data.

## Related

This issue was discovered while investigating #524. The immediate symptom (EXTRACT_FASTQ_LIST failing on non-numeric partition identifiers) was fixed in #525, but the root cause remains.

## Recommended fix

**Option A (preferred)**: Don't create any output file when there's no data. Downstream processes should handle the case where no partition files are produced.

**Option B**: Use a clearly distinguishable sentinel value like `__no_data__` that cannot collide with real column values.

Either fix would require:
1. Updating `partition_tsv.py`
2. Updating the `PARTITION_TSV` process output declaration to handle empty output
3. Auditing all callers of `PARTITION_TSV` to ensure they handle the empty case

---

*Generated with Claude Code*

willbradshaw:
Fixes #526 and #494 by introducing dedicated handling for groups for which RUN returns no vertebrate-viral hits. Specifically, this PR:

- Modifies `validateGrouping` to separately identify and return (1) empty samples with nonempty group-mates, and (2) empty samples with only empty group mates (empty groups).
- Modifies `partitionTsv` to return nothing if given empty input (instead of creating an artificial "empty" group).
- Adds a `createEmptyGroupOutputs` module which iterates over identified empty groups and creates empty output files with the expected filenames.
- Modifies `DOWNSTREAM` to tie the above changes together and publish empty outputs.


dp-rice:
My naive question is why can't we have a convention that each process that outputs a TSV write the header whether or not it has any input data? Then null results should propagate "naturally" through the pipeline. Since the headers aren't data dependent (right?) there's nothing to derive. If we update a process to have different columns, the header would get updated too.

Issue #611, opened 2026-01-27
Add ONT benchmark test


willbradshaw:
Since #588 we've had a long-running Illumina benchmark test set up to run before merging to `main`. We now also have a public benchmarking dataset prepared for ONT. We should set up a benchmark test for ONT similar to that for Illumina.

PR #612, opened 2026-01-27
Add ONT benchmark test file

willbradshaw:
Addresses #611 by adding an ONT benchmark test similar to the pre-existing Illumina benchmark. Also fixes a fiddly bug that was causing the benchmark run to fail.

Originally I wanted to use a 2M-read benchmark here, but that was timing out after 2 hours without even finishing RUN. Currently this is using a 100k-read benchmark, which is smaller than I want but has the great virtue of finishing on a similar timeframe to the other integration tests. In the near future, we should look into what's bottlenecking the ONT pipeline, resolve it, then increase the size of the benchmark.

### DIFF
--- a/.github/workflows/check-release.yml
+++ b/.github/workflows/check-release.yml
@@ -1,0 +1,49 @@
+name: Check release readiness
+
+on:
+  pull_request:
+    branches: [main]
+
+jobs:
+  check-release:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+
+      - name: Extract version from pyproject.toml
+        id: version
+        run: |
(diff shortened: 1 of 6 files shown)


modules/local/addConditionalTsvColumn/resources/usr/bin/add_conditional_tsv_column.py
```diff
@@ -61,13 +61,13 @@ def add_conditional_column(input_path, chk_col, match_val, if_col, else_col, new
         # Validate required columns exist
         validate_columns(reader.fieldnames, chk_col, if_col, else_col)
 
-        # Write header with new column
+        # Write header with new column (use simple tab-join to avoid CSV quoting)
         fieldnames_out = list(reader.fieldnames) + [new_hdr]
-        writer = csv.DictWriter(outf, fieldnames=fieldnames_out, delimiter='\t', lineterminator='\n')
-        writer.writeheader()
+        outf.write("\t".join(fieldnames_out) + "\n")
```
katherine-stansifer: why this solution rather than user DictWriter with csv.QUOTE_NONE ? (https://docs.python.org/3/library/csv.html)

PR #626, opened 2026-01-29
Move DB download functionality to Python

willbradshaw:
Following on from #624, moves DB download functionality from `bin/download-db.sh` to `bin/download_db.py` and adds unit tests at `bin/test_download_db.py`. This will hopefully help us catch issues like the one fixed in #624 more easily.

### DIFF
--- a/bin/conftest.py
+++ b/bin/conftest.py
@@ -1,2 +1,0 @@
-# pytest configuration for bin/ directory
-# This file's presence adds bin/ to sys.path automatically
--- a/bin/download_db.py
+++ b/bin/download_db.py
@@ -93,13 +93,12 @@ def file_lock(lock_file: Path, timeout_seconds: int | None = None):
         TimeoutError: If the lock cannot be acquired within the timeout period
     """
     lock_file.parent.mkdir(parents=True, exist_ok=True)
-    fd = os.open(str(lock_file), os.O_RDWR | os.O_CREAT)
-    try:
+    with open(lock_file, "w") as f:
         if timeout_seconds is not None:
             start_time = time.time()
             while True:
                 try:
-                    fcntl.flock(fd, fcntl.LOCK_EX | fcntl.LOCK_NB)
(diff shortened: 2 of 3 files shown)


willbradshaw:
@katherine-stansifer no urgency on this one, fine to merge it next week

bin/conftest.py
katherine-stansifer: harmless, but I don't think you actually need this for pytest to find the tests in the bin directory?

bin/download_db.py
```diff
@@ -0,0 +1,240 @@
...
+
+def sync_from_s3(source_path: str, local_path: Path) -> None:
+    """
+    Sync a database from S3 using 'aws s3 sync --delete'.
+    The --delete flag removes local files not present in the source,
+    preventing stale files from previous runs.
+    """
+    logger.info(f"Syncing from S3: {source_path} -> {local_path}")
+    configure_aws_s3_transfer()
+    result = subprocess.run(
```
katherine-stansifer: nit: consider using check=True rather than manually checking the return code

katherine-stansifer (approved):
Code looks very nice! python is much easier to read than bash, and I like the hashes to keep different download directories separate.

Concurrency is tricky so I think it would be a good idea to, at some point soon, run a recent ONT delivery with this new code to make sure:
- there are no unexpected bugs
- performance (ie, time taken up by kraken) hasn't substantially regressed

This seems like it should just work absolutely fine with comparable performance to the old script--but the sharding project gave me a enough counterintuitive results that I think it would be smart to check. 
[ this doesn't need to happen before merge to dev, and if you want to delegate this to me as part of my "work with Alessandro on better ZEPHYR computational support" project I'm happy to help]

bin/test_download_db.py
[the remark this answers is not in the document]
willbradshaw: I do actually think heavily mocked tests for high-level functions add value; I've been bitten before by unit-testing all the helper functions and then failing due to a dumb typo in `main().

PR #627, opened 2026-01-30
327: process vsearch cluster output linewise

evanfields:
Convert processVsearchOutput to small Rust streaming utility, using the new Rust build system. Memory goes from O(reads) to O(number of clusters). The coefficient in front of that big-O also improves since we're no longer using Python.


evanfields:
Now that I've changed the base branch to dev (and the Rust build system is merged), I'm deleting the prior Gemini comments and asking it for a new review.

willbradshaw (approved):
Approved conditional on fixing conflicts with dev (+ updated tests passing)

PR #638, opened 2026-01-31
Remove BLAST validation from RUN workflow

willbradshaw:
Now that we have BLAST validation in DOWNSTREAM, the functionality in RUN is redundant complexity. This PR removes that functionality from the RUN workflow, deletes obsoleted processes & subworkflows, & updates tests accordingly.


willbradshaw:
> Do we really need to bump the version? I get that it's technically changing the output, but if no one was running the optional blast validation in RUN, then does it actually matter? It would cause us significantly less headache to just call this a point release.

I think this is a good point, but also, I'm planning to roll this in with a bunch of other changes that are definitely compatibility-breaking, so it will end up being moot.

PR #644, opened 2026-02-04
Change QC outputs to per-sample format

willbradshaw:
Changes QC outputs from concatenated to per-sample format, with separate outputs for pre- and post-cleaning, and updates associated tests and documentation.

- Updates `SUMMARIZE_MULTIQC` to emit tuples with sample name
- Removes concatenation steps from `QC` and `RUN_QC`
- Updates associated snapshot data

Since `DOWNSTREAM` does not currently consume any of these outputs, no changes to that workflow are needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # v3.1.0.0-dev
 
+- Add datapackage table-schema support for workflow outputs (proof-of-concept with `duplicate_stats`):
+    - Created `schemas/` directory with `duplicate_stats.schema.json` table-schema definition.
+    - Added `bin/validate_schemas.py` script to validate output files against schemas using frictionless library.
+    - Updated `CREATE_EMPTY_GROUP_OUTPUTS` to generate headers from schemas for empty output files.
+    - Added schema validation step to DOWNSTREAM workflow CI after nf-test.
+    - Created reusable `.github/actions/setup-python` composite action for Python environment setup.
 - Remove BLAST validation from RUN workflow. BLAST validation is now only available in the DOWNSTREAM workflow via VALIDATE_VIRAL_ASSIGNMENTS.
     - Deleted `BLAST_VIRAL` subworkflow, `SUBSET_FASTN` module, and `RUN_VALIDATION` workflow.
     - Removed `blast_viral_fraction` and related BLAST parameters from RUN workflow configs.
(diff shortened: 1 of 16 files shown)


willbradshaw:
@dp-rice most of the file changes are tiny QC files in the snapshot. These have mostly undergone one of a few operations:

1. Change to filename without change to contents (same hash in `.snap` file)
2. Split into pre- and post-cleaning files (new hash in `.snap` file)

Substantively, I think you can pretty much ignore these.

CHANGELOG.md
```diff
@@ -17,7 +17,9 @@
 - Convert `setup-rust-container` from reusable workflow to composite action, simplifying CI check reporting.
 - Remove confusing `workflow_run` triggers from integration tests (benchmark and test-chained workflows).
 - Refactored processVsearchClusterOutput module to use streaming Rust implementation rather than memory-intensive Python/Pandas.
-- Changed read counts output from concatenated `read_counts.tsv.gz` to per-sample `{sample}_read_counts.tsv` files.
+- Changed RUN outputs to per-sample format:
+    - Read counts: `read_counts.tsv.gz` → `{sample}_read_counts.tsv`
```
dp-rice: Is `sample` guaranteed to only have valid characters for a filename? What happens if it doesn't?

Also, are sample names allowed to have `_`? Even of so, I think this is fine because we can still pattern match on the suffixes.

pyproject.toml
```diff
@@ -90,14 +90,19 @@ expected-outputs-run = [
...
+    "results/{SAMPLE}_qc_adapter_stats_cleaned.tsv.gz",
+    "results/{SAMPLE}_qc_adapter_stats_raw.tsv.gz",
+    "results/{SAMPLE}_qc_basic_stats_cleaned.tsv.gz",
+    "results/{SAMPLE}_qc_basic_stats_raw.tsv.gz",
+    "results/{SAMPLE}_qc_length_stats_cleaned.tsv.gz",
+    "results/{SAMPLE}_qc_length_stats_raw.tsv.gz",
+    "results/{SAMPLE}_qc_quality_base_stats_cleaned.tsv.gz",
+    "results/{SAMPLE}_qc_quality_base_stats_raw.tsv.gz",
+    "results/{SAMPLE}_qc_quality_sequence_stats_cleaned.tsv.gz",
+    "results/{SAMPLE}_qc_quality_sequence_stats_raw.tsv.gz",
```
dp-rice: ~Surprised to see files being added here rather than just removed. Are we adding new output with cleaned vs raw?~ Update: Just reread the PR description.

dp-rice (approved):
Looks good modulo the sample-as-filename question.

PR #647, opened 2026-02-04
Add datapackage table-schema support for workflow outputs

willbradshaw:
This PR implements a proof-of-concept for using datapackage table-schemas to document and enforce workflow output file formats, starting with the `duplicate_stats` output.

**Changes:**
- Created `schemas/` directory with `duplicate_stats.schema.json` table-schema definition (v2 profile)
- Added `bin/validate_schemas.py` script to validate output files against schemas using the frictionless library
- Added `bin/test_validate_schemas.py` with pytest tests for the validation script
- Updated `CREATE_EMPTY_GROUP_OUTPUTS` module to read headers from schemas when generating empty output files
- Added schema validation step to DOWNSTREAM workflow CI after nf-test completes
- Created reusable `.github/actions/setup-python` composite action for Python environment setup
- Updated `.github/workflows/pytest.yml` to use the new setup-python action
- Added `frictionless>=5.18.0` dependency to `pyproject.toml`
- Added `.nf-test` directory cleanup to setup-nf-test action

### DIFF
--- a/bin/test_validate_schemas.py
+++ b/bin/test_validate_schemas.py
@@ -191,6 +191,98 @@ class TestValidateFile:
         assert not is_valid
         assert len(errors) > 0
 
+    def test_empty_data_file_is_valid(self, tmp_path: Path, simple_schema: Path) -> None:
+        """Header-only files should validate successfully."""
+        data_file = tmp_path / "test.tsv"
+        data_file.write_text("col1\tcol2\n")
+        is_valid, errors = validate_file(data_file, simple_schema)
+        assert is_valid
+        assert errors == []
+
+    @pytest.mark.parametrize(
+        "schema_fields,data,should_pass",
+        [
+            # Required constraint
+            (
+                [{"name": "col1", "type": "string", "constraints": {"required": True}}],
+                "col1\n\n",
+                False,
+            ),
+            # Pattern constraint - invalid
+            (
+                [{"name": "id", "type": "string", "constraints": {"pattern": "^[^/]+(/[^/]+)?$"}}],
+                "id\nfoo/bar/baz\n",
+                False,
+            ),
+            # Pattern constraint - valid
+            (
+                [{"name": "id", "type": "string", "constraints": {"pattern": "^[^/]+(/[^/]+)?$"}}],
+                "id\nNC_001234.1\nAB/CD\n",
+                True,
+            ),
+            # Enum constraint - invalid
+            (
+                [{"name": "status", "type": "string", "constraints": {"enum": ["pass", "fail"]}}],
+                "status\nunknown\n",
+                False,
+            ),
+            # Minimum constraint - invalid
+            (
+                [{"name": "count", "type": "integer", "constraints": {"minimum": 0}}],
+                "count\n-5\n",
+                False,
+            ),
+            # Maximum constraint - invalid
+            (
+                [{"name": "frac", "type": "number", "constraints": {"maximum": 1.0}}],
+                "frac\n1.5\n",
+                False,
+            ),
+        ],
+        ids=["required", "pattern_invalid", "pattern_valid", "enum", "minimum", "maximum"],
+    )
+    def test_constraint_validation(
+        self, tmp_path: Path, schema_fields: list, data: str, should_pass: bool
+    ) -> None:
+        schema = {
+            "$schema": "https://datapackage.org/profiles/2.0/tableschema.json",
+            "fields": schema_fields,
+        }
+        schema_path = tmp_path / "test.schema.json"
+        schema_path.write_text(json.dumps(schema))
+        data_file = tmp_path / "test.tsv"
+        data_file.write_text(data)
+        is_valid, errors = validate_file(data_file, schema_path)
+        assert is_valid == should_pass
+        if not should_pass:
+            assert len(errors) > 0
+
+    def test_missing_values_handled(self, tmp_path: Path) -> None:
+        """Values in missingValues should be treated as null."""
+        schema = {
+            "$schema": "https://datapackage.org/profiles/2.0/tableschema.json",
+            "missingValues": ["", "NA"],
+            "fields": [
+                {"name": "col1", "type": "integer"},
+                {"name": "col2", "type": "integer"},
+            ],
+        }
+        schema_path = tmp_path / "test.schema.json"
+        schema_path.write_text(json.dumps(schema))
+        data_file = tmp_path / "test.tsv"
+        # Both "NA" and "" should be treated as null, not as invalid integers
+        data_file.write_text("col1\tcol2\n42\tNA\nNA\t99\n100\t\n")
+        is_valid, errors = validate_file(data_file, schema_path)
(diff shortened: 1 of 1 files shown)


schemas/duplicate_stats.schema.json
```diff
@@ -0,0 +1,53 @@
...
+  "$schema": "https://datapackage.org/profiles/2.0/tableschema.json",
+  "title": "Summary Statistics for Alignment-Based Duplicate Marking",
+  "description": "Summary statistics for groups of duplicate reads identified by the DOWNSTREAM workflow. Each row represents a single duplicate group: a set of reads aligned to the same genome, with matching start coordinates (within configurable tolerance) for both forward and reverse reads. Groups are defined based on the primary alignment of each read and do not account for LCA information.",
+  "missingValues": ["", "NA"],
+  "primaryKey": ["prim_align_dup_exemplar"],
+  "fields": [
+    {
+      "name": "prim_align_genome_id_all",
+      "title": "Genome ID",
+      "description": "NCBI accession ID(s) of the genome(s) that reads in this duplicate group aligned to in their primary alignment. For split alignments where forward and reverse reads map to different genomes, genome IDs are joined with '/' and sorted alphabetically.",
```
dp-rice: Would be good to enforce

> genome IDs are joined with '/'

less important to enforce:

> sorted alphabetically

dp-rice: Sorry I meant the "/" separated list, which we might want to parse.

dp-rice: I just read your comment better. Don't we just need to know that they don't have slashes? If they do we shouldn't use slashes to separate, right?

willbradshaw:
> What an mgs-workflow developer is meant to do with this. When do you add a new schema/change a schema? Do you have to do anything else to have it be used?

I was previously thinking of this as one of our many "don't worry about it unless it fails CI" constraints, but I think in fact we should document this in the developer docs so people know to check the schema when altering output fields. Will do.

> What's the status of this change? Can the schema be adopted piecewise? What's a minimal piece that that can be added at once.

I'm not sure I understand the question correctly, but: this change introduces and enforces a schema on DOWNSTREAM duplicate_stats files. In addition to the schema itself, it also introduces the code needed to (1) ensure no-data stats files have the right columns (otherwise the schema would fail), and (2) validate the schema in CI. I think this is roughly the minimal change required to get a working schema that actually does anything.

Once this is merged, adding schemas for other DOWNSTREAM outputs should be as simple as defining a new schema file; this should be consumed by both CI and `createEmptyGroupOutputs` without further code changes.

> How are the previously-empty files generated? Are they guaranteed to have the same header as the files with reads? Which files does this apply to?

I'm a bit confused about this question as I think it's covered in the PR description. These files are generated by `createEmptyGroupOutputs`, as before. If a schema is available, that module derives its column names directly from the schema fields. Currently, only a schema for duplicate stats is available, so other outputs are still empty.

> What does this mean for the consumers of the output? If I want to understand or use an output, how should I go about that? How should I track schema changes? What level of change merits a schema version bump? Any change to the schema files? Any change to column names? Etc

Consumers of the output can (a) refer to the schema file to understand the file contents, and (b) build dependent tooling that relies on the schema being correct. Any future change to the schema (except purely documentation changes like modifying `title` or `description`) will require a schema bump.

dp-rice:
> I'm not sure I understand the question correctly, but: this change introduces and enforces a schema on DOWNSTREAM duplicate_stats files. In addition to the schema itself, it also introduces the code needed to (1) ensure no-data stats files have the right columns (otherwise the schema would fail), and (2) validate the schema in CI. I think this is roughly the minimal change required to get a working schema that actually does anything.
> 
> Once this is merged, adding schemas for other DOWNSTREAM outputs should be as simple as defining a new schema file; this should be consumed by both CI and createEmptyGroupOutputs without further code changes.

Thanks! This answers my question. I was confused by the phrase "proof of concept" in the PR description, which made me wonder how complete the implementation was.

modules/local/createEmptyGroupOutputs/resources/usr/bin/create_empty_group_outputs.py
```diff
@@ -102,33 +104,90 @@ def get_group_output_patterns(pyproject_path: str, platform: str) -> list[str]:
...
+        schema_name (str): Base name of the schema (e.g., 'duplicate_stats').
+    Returns:
+        list[str] | None: List of column names, or None if no schema found.
+    """
+    schema_path = schema_dir / f"{schema_name}.schema.json"
+    if not schema_path.exists():
+        return None
+    with open(schema_path) as f:
+        schema = json.load(f)
+    fields = schema.get("fields", [])
```
dp-rice: Can a valid schema lack the `"fields"` entry?

dp-rice (approved):
Thanks for the docs updates! I now understand the plan better and think it's a good one. Left some comments, but I don't need to re-review. LMK if there are places where I suggested a change or an issue and you decide on neither.

One high-level comment:
Including this in CI is good, but fundamentally limited by what we include in our test dataset there. It seems very likely to me that things will show up in real data that cause schema violations and it would be good to know that ASAP. I think it would be very good to add schema validation as a final step in the pipeline. Probably the pipeline should not fail if there's a schema violation but it should alert us very loudly. (This is obviously a separate issue/PR).

bin/validate_schemas.py
```diff
@@ -0,0 +1,278 @@
...
+        Tuple of (is_valid, list of error messages).
+    """
+    with decompressed_path(data_file) as file_to_validate:
+        dialect = Dialect(controls=[formats.CsvControl(delimiter="\t")])
+        resource = Resource(
+            path=str(file_to_validate),
+            schema=str(schema_path),
+            dialect=dialect,
+        )
+        with system.use_context(trusted=True):
```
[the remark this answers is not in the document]
willbradshaw: I went back and forth with Claude about this some, and concluded that we do need it and that it's safe in this context. Other implementations are possible but substantially more complex.

schemas/duplicate_stats.schema.json
```diff
@@ -0,0 +1,53 @@
+{
+  "$schema": "https://datapackage.org/profiles/2.0/tableschema.json",
+  "title": "Summary Statistics for Alignment-Based Duplicate Marking",
+  "description": "Summary statistics for groups of duplicate reads identified by the DOWNSTREAM workflow. Each row represents a single duplicate group: a set of reads aligned to the same genome, with matching start coordinates (within configurable tolerance) for both forward and reverse reads. Groups are defined based on the primary alignment of each read and do not account for LCA information.",
+  "missingValues": ["", "NA"],
```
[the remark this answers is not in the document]
willbradshaw: I did, but I wonder if we should actually ban empty strings and have `NA` as our only null value. It's moot for this file since all columns are required (i.e. non-nullable) but matters for other outputs.

schemas/duplicate_stats.schema.json
willbradshaw: Added

PR #650, opened 2026-02-06
Switch DOWNSTREAM test data from S3 to local paths

willbradshaw:
- Renames `test-data/results` subdirectories and `tests/workflows/*.snap` files for mutual consistency.
- Adds validation that md5 sums match between files in workflow snapshots and corresponding `test-data/results` subdirectories
- Updates `DOWNSTREAM` input configuration to handle local relative paths, removing dependency on S3 inputs
- Updates `DOWNSTREAM` snapshot to bring it into sync with `RUN` outputs (previously got out of sync over the past few PRs)

Collectively, these changes ensure that `RUN` outputs and `DOWNSTREAM` inputs stay in sync and removes the possibility of drift between local and S3 test inputs. They also eliminate race conditions in the event that two developers are working on `DOWNSTREAM` inputs at the same time (previously both would have needed to edit the same S3 files).

Most of the changed files are simple renamings, either of the files themselves or of paths they reference. The main exceptions are:

- Modifying `‎subworkflows/local/loadDownstreamData/main.nf‎` to handle relative paths
- New `‎bin/validate_test_data_sync.py‎` and `‎.github/workflows/validate-test-data.yml‎` validation code.
- Updated `DOWNSTREAM` snapshot (`‎tests/workflows/downstream.nf.test.snap‎`) and associated output files (‎`test-data/results/downstream_output_shortread/tt1_validation_hits.tsv`, ‎ `‎test-data/results/downstream_output_shortread/tt1_duplicate_reads.tsv`, `‎test-data/results/downstream_output_ont/tt1_validation_hits.tsv‎`‎) (not much to be gained by reviewing these)

### DIFF
--- a/bin/test_validate_test_data_sync.py
+++ b/bin/test_validate_test_data_sync.py
@@ -26,26 +26,26 @@ class TestParseSnapshot:
     @pytest.mark.parametrize(
         "content_entries,expected_md5_map",
         [
-            # Single file
-            (["file1.tsv.gz:md5,abc123"], {"file1": "abc123"}),
+            # Single file (.gz stripped)
+            (["file1.tsv.gz:md5,abc123"], {"file1.tsv": "abc123"}),
             # Multiple files
             (
                 ["file1.tsv.gz:md5,abc123", "file2.tsv.gz:md5,def456"],
-                {"file1": "abc123", "file2": "def456"},
+                {"file1.tsv": "abc123", "file2.tsv": "def456"},
             ),
-            # Various extensions stripped
+            # Only .gz is stripped, other extensions preserved
             (
                 [
                     "file.tsv.gz:md5,abc123",
                     "another.csv.gz:md5,def456",
                     "simple.txt:md5,ghi789",
                 ],
-                {"file": "abc123", "another": "def456", "simple": "ghi789"},
+                {"file.tsv": "abc123", "another.csv": "def456", "simple.txt": "ghi789"},
             ),
             # Entries without :md5, are ignored
             (
                 ["file1.tsv.gz:md5,abc123", "not_an_md5_entry", "another:entry:without:md5"],
-                {"file1": "abc123"},
+                {"file1.tsv": "abc123"},
             ),
             # Empty content
             ([], {}),
@@ -61,7 +61,8 @@ class TestParseSnapshot:
         result = parse_snapshot(snapshot_path)
         assert result == {"test_output": expected_md5_map}
 
-    def test_parses_multiple_snapshots(self, tmp_path: Path) -> None:
+    def test_parses_multiple_snapshot_names(self, tmp_path: Path) -> None:
+        """Parametrized cases above use a single snapshot name; verify multiple names in one file."""
         snapshot = make_snapshot({
             "snapshot_one": ["file1.tsv.gz:md5,aaa111"],
             "snapshot_two": ["file2.tsv.gz:md5,bbb222"],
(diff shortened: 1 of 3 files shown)


test-data/results/downstream_output_ont/tt1_validation_hits.tsv
evanfields: Commenting here but it's a generic question about downstream outputs: why do we need the DOWNSTREAM output files themselves, not just hashes thereof? Seems like we need the RUN outputs as inputs to DOWNSTREAM, but nothing is consuming the outputs of DOWNSTREAM?

(Sort of exception: `validate_test_data_sync.py` reads the DOWNSTREAM outputs, but wouldn't have to if we didn't keep them in git.)

bin/validate_test_data_sync.py
```diff
@@ -0,0 +1,167 @@
...
+    for snapshot_name, md5_map in snapshot_data.items():
+        logger.info(f"Validating {snapshot_name}")
+        snapshot_results_dir = results_dir / snapshot_name
+        if not snapshot_results_dir.exists():
+            msg = f"Snapshot directory not found: {snapshot_results_dir}"
+            logger.error(msg)
+            raise FileNotFoundError(msg)
+        errors: list[str] = []
+        for snapshot_stem, expected_md5 in md5_map.items():
+            # Find file matching the stem (any extension)
```
evanfields: Why do we strip the extension in `parse_snapshot` then look for any extension here, rather than just looking for exact filename matches (including extensions)?

bin/test_validate_test_data_sync.py
```diff
@@ -0,0 +1,184 @@
...
+    def test_parses_content_entries(
+        self, tmp_path: Path, content_entries: list[str], expected_md5_map: dict[str, str]
+    ) -> None:
+        snapshot = make_snapshot({"test_output": content_entries})
+        snapshot_path = tmp_path / "test.snap"
+        snapshot_path.write_text(json.dumps(snapshot))
+        result = parse_snapshot(snapshot_path)
+        assert result == {"test_output": expected_md5_map}
+
+    def test_parses_multiple_snapshots(self, tmp_path: Path) -> None:
```
evanfields: This isn't already handled by the parameterization on `test_parses_content_entries`?

evanfields (approved):
Nice cleanup overall! It'll be a solid upgrade to not risk test data getting out of sync, as we've recently seen.

Besides mostly-optional inline nits, my two main questions are:
1. Do we need DOWNSTREAM output files at all? Why? (Plausible answer: consistent handling of different workflows, even though we don't use them right now.)
2. We now have two representatives of each workflow output: the actual raw output files, plus the MD5 hashes thereof. This seems duplicative; if we have raw files, why do we need hashes at all -- can't we just diff generated outputs against expected outputs? My understanding is no, we can't; nf-test requires the hash format. 
  * This seems like a case where doing the reasonable-on-the-margin thing of keeping both file types so we can use nf-test is increasing the CI complexity: now we have multiple output records to record, an extra CI step, and hundreds of extra lines of CI code (+test thereof).
  * I don't think there's anything actionable here besides "Will as repo maintainer thinks hard about this and perhaps makes a Linear project for future work."

I don't need to re-review!

subworkflows/local/loadDownstreamData/main.nf
```diff
@@ -5,6 +5,7 @@
 workflow LOAD_DOWNSTREAM_DATA {
     take:
         input_file
+        input_base_dir  // Base directory for resolving relative paths in input CSV (use projectDir or launchDir)
```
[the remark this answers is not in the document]
willbradshaw: [Claude Code] Done — removed the parenthetical.

PR #651, opened 2026-02-06
Address processViralMinimap2Sam OOM

evanfields:
Address processViralMinimap2Sam OOM:
* Send only viral reads, not all reads, to the processViralMinimap2Sam process;
* Internally, the process now sorts its inputs and uses a two-pointer merge algorithm for O(1) memory.

### DIFF
--- a/modules/local/processViralMinimap2Sam/resources/usr/bin/process_viral_minimap2_sam.py
+++ b/modules/local/processViralMinimap2Sam/resources/usr/bin/process_viral_minimap2_sam.py
@@ -1,12 +1,14 @@
 #!/usr/bin/env python3
 
 import argparse
-import datetime
 import gzip
+import logging
 import math
 import sys
+from typing import Any
 import tempfile
 import time
+from datetime import datetime, timezone
 
 import pandas as pd
 import pysam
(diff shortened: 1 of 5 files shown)


modules/local/processViralMinimap2Sam/main.nf
```diff
@@ -12,12 +12,37 @@ process PROCESS_VIRAL_MINIMAP2_SAM {
...
+        rm raw.sam
+
+        # Sort FASTQ by read id.
+        # 4-line records are combined into one tap-separated line, sorted with C locale,
+        # and then split back into 4-line records.
+        zcat ${clean_reads} \
+            | paste - - - - \
+            | LC_ALL=C sort -k1,1 -S 1G \
+            | tr '\t' '\n' \
+            > sorted.fastq
```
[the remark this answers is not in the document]
willbradshaw: Yeah, this goes quite far beyond my preferred limit for Nextflow-level complexity.

My first preference would be to use the pre-existing sorting processes for this (`sortFastq`, `sortFile`, `sortTsv`); second preference would be to move it into Python. Distant third preference would be to add additional test coverage to nf-test to cover the sorting behavior.

modules/local/processViralMinimap2Sam/resources/usr/bin/process_viral_minimap2_sam.py
```diff
@@ -1,267 +1,184 @@
...
-        "-a",
-        "--sam",
-        type=str,
-        required=True,
-        help="Path to Minimap2 SAM alignment file.",
-    )
-    parser.add_argument(
-        "-r",
-        "--reads",
-        type=lambda f: open_by_suffix(f, "r"),
```
willbradshaw: Why drop this?

modules/local/processViralMinimap2Sam/resources/usr/bin/process_viral_minimap2_sam.py
```diff
@@ -1,267 +1,184 @@
...
-        print_log("Output path: {}".format(out_file))
 
-        # Import metadata and viral DB
-        print_log("Importing Genbank metadata file...")
-        meta_db = pd.read_csv(meta_path, sep="\t", dtype=str)
+        meta_db = pd.read_csv(args.metadata, sep="\t", dtype=str)
         genbank_metadata = {
-            genome_id: [taxid, species_taxid]
-            for genome_id, taxid, species_taxid in zip(
+            gid: [tid, stid]
```
willbradshaw: A bit confused at the decision to change this; seems strictly less readable to me.

modules/local/processViralMinimap2Sam/resources/usr/bin/process_viral_minimap2_sam.py
willbradshaw: I had a range of comments of the form "this script does not meet our usual quality standards" (around e.g. logging & type hints), to which the predictable response would be "not changed in this PR, it didn't meet those standards before". We should sync up on whether the resolution should be "let's keep this PR minimal" versus "we should take the opportunity to bring this script up to code".

modules/local/processViralMinimap2Sam/resources/usr/bin/test_process_viral_minimap2_sam.py
willbradshaw: I'd like to have more comprehensive testing here; Claude Code can do a pretty good job of this as long as you sternly warn it to use `@pytest.mark.parametrize`, and I've caught several edge cases this way in the past couple months.

willbradshaw (changes requested):
Most important issue is that I don't like having this much functionality in long Nextflow `shell` blocks.

modules/local/processViralMinimap2Sam/main.nf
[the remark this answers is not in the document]
evanfields: The above could also be done all in this PR, which would make this PR larger in semantic scope but smaller in lines.

PR #652, opened 2026-02-08
Convert viral hits outputs to per-sample format

willbradshaw:
Changes viral hits outputs in RUN workflow from concatenated to per-sample format, and adapts DOWNSTREAM workflow to ingest per-sample inputs.

- Removes concatenation steps from `processLcaAlignerOutput`.
- Updates DOWNSTREAM input CSV format to take a `results_dir` and discover any viral hits files in that directory, rather than taking a single `hits_tsv`.
- Updates DOWNSTREAM test input files in `test-data/downstream` to use the new format.
- Updates `loadDownstreamData` to (1) ingest the new input CSV format, and (2) concatenate incoming virus hits files into a single file per input row.

This results in a somewhat absurd configuration where the RUN workflow splits the viral hits across samples, only to have DOWNSTREAM concatenate them anyway. The purpose of this is to aid review of the changes to RUN by minimizing consequent changes to DOWNSTREAM. A future PR will add in the changes to DOWNSTREAM required to take full advantage of the new per-sample inputs from RUN.

### DIFF
--- a/subworkflows/local/loadDownstreamData/main.nf
+++ b/subworkflows/local/loadDownstreamData/main.nf
@@ -11,14 +11,14 @@ include { CONCATENATE_TSVS_LABELED } from "../../../modules/local/concatenateTsv
 workflow LOAD_DOWNSTREAM_DATA {
     take:
         input_file
-        input_base_dir  // Base directory for resolving relative paths in input CSV (use projectDir or launchDir)
+        input_base_dir  // Base directory for resolving relative paths in input CSV
     main:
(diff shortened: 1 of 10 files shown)


subworkflows/local/loadDownstreamData/main.nf
dp-rice: I haven't parsed all the logic here, but catching bugs in something like this is what integration tests are for. Do we have a regression test case for the issue Gemini pointed out?

PR #659, opened 2026-02-15
Add group-level read counts to DOWNSTREAM output

willbradshaw:
- High-level change: Added group-level read-count outputs to DOWNSTREAM workflow (`{GROUP}_read_counts.tsv.gz`), with schema validation & enforcement (`schemas/read_counts.schema.json`).
- Simplified `LOAD_DOWNSTREAM_DATA` to emit `run_dirs` (unique resolved directory paths) and `groups` (label/sample/group tuples), removing hardcoded virus hits discovery.
- Created reusable `DISCOVER_RUN_OUTPUT` subworkflow that discovers all per-sample TSV files from RUN output directories and matches them to sample groups using filename prefixes.
- Created general-purpose `CONCAT_BY_GROUP` subworkflow (replacing `PREPARE_GROUP_TSVS`) that filters by file suffix, concatenates by group, adds a group column, and renames to clean output filenames.
- Future PRs will add more group-level outputs using the infrastructure developed here.


willbradshaw (commented):

## Correctness

Good test coverage overall. Unit tests exist for all three subworkflows (`LOAD_DOWNSTREAM_DATA`, `DISCOVER_RUN_OUTPUT`, `CONCAT_BY_GROUP`), and the downstream integration tests validate end-to-end output with MD5 snapshots for both shortread and ONT platforms.

1. **nit**: `CONCAT_BY_GROUP` unit test only exercises the `hits.tsv` suffix, not `read_counts.tsv`. The integration test does cover read counts end-to-end (the snapshot asserts `tt1_read_counts.tsv.gz` with an expected MD5), so this is not a gap in practice, but a dedicated unit test case with read counts data would make the suffix-filtering logic easier to verify in isolation.

2. **nit**: The regex in `CONCAT_BY_GROUP` interpolates the suffix string directly into a regex pattern (`f.name ==~ /.*_${suffix}(\.gz)?$/`), so the dots in suffixes like `virus_hits.tsv` act as wildcard matchers rather than literal dots. In practice this won't cause false matches since RUN output filenames follow predictable conventions, but it's worth noting if the set of suffixes ever grows.

## Documentation

No issues found. The CHANGELOG, `docs/downstream.md` flowcharts (both shortread and ONT), subworkflow descriptions, and all textual references to `PREPARE_GROUP_TSVS` have been updated to reflect the new subworkflow names. The new `read_counts.schema.json` is well-documented with field descriptions and examples.

## Quality

1. **nit**: Indentation is inconsistent in the `emit` block of `workflows/downstream.nf`. Lines 68–69 use 7 spaces of indentation while the surrounding `main:` block uses 8. Compare:
   ```
          input_downstream = ...   // 7 spaces
          logging_downstream = ... // 7 spaces
   ```
   vs. the `main:` body which uses 8 spaces throughout.

2. **suggestion**: `workflows/downstream.nf` mixes two styles for accessing `VALIDATE_VIRAL_ASSIGNMENTS` outputs: line 70 uses `VALIDATE_VIRAL_ASSIGNMENTS.out.blast_results` directly, while line 73 uses the stored reference `validate_ch.annotated_hits`. Consider using `validate_ch` consistently for both to match the pattern established by `load_ch` and `discover_ch` earlier in the workflow.

willbradshaw:
*Automated response from Claude Code*

Addressed the following suggestions from det-code-review in 77046a47:

- **Escape dots in suffix regex**: Added `suffix.replace('.', '\\.')` before regex interpolation in `CONCAT_BY_GROUP` to prevent unescaped dots from matching any character.
- **Fix emit block indentation**: Normalized all emit entries to consistent 8-space indentation.
- **Use validate_ch consistently**: Changed `VALIDATE_VIRAL_ASSIGNMENTS.out.blast_results` to `validate_ch.blast_results` to match the rest of the emit block.

dp-rice (approved):
Please address comments C1 and T1 in the [last auto-review](https://github.com/securebio/nao-mgs-workflow/pull/659#pullrequestreview-3810423938).

Also, I know we want to get away from `nf-test`, but I'd consider adding a test that would catch issue C1 and confirm that the fix makes it pass.

Mostly focused my manual review on the schema. If you can please the auto-reviewer after those changes, I'm satisfied.

PR #658, opened 2026-02-13
Merge 3.1.0.0 into main

willbradshaw:

PR #662, opened 2026-02-16
Reduce maxRetries to 1 and add job queue closure documentation

jo-delia-faraguna:
## Summary
- Reduce `process.maxRetries` from 3 to 1 in `standard` and `batch` profiles
  - Failures could be either spot reclamations, non-stochastic process errors, or stochastic process errors
  - Reducing from 3 to 1 will handle spot reclamations and surface non-stochastic errors faster
  - Reducing from 3 to 1 might reduce our ability to overcome stochastic process errors. **It's not clear to me how common these are and if the current pattern (retry the entire process 3 times) is the best way to handle them.**
- Add documentation about using closure to switch to backup job queue after one failure

## Acceptance criteria
- [x] Verify retry behavior with a spot reclamation scenario
- [x] Confirm on-demand fallback triggers on first retry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# v3.1.0.1-dev
+# v3.1.0.1
 
 - Reduced `maxRetries` from 3 to 2 in `standard` and `batch` profiles for spot-to-on-demand fallback (#662)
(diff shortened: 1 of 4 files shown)


willbradshaw (approved):
Approved, but please don't merge until #658 is in

PR #665, opened 2026-02-18
Add group-level QC outputs to DOWNSTREAM

willbradshaw:
- Added group-level QC outputs (`{GROUP}_qc*.tsv.gv`) to DOWNSTREAM workflow for both short-read and ONT platforms.
- Implements table schemas for all group-level QC outputs.


subworkflows/local/concatRunOutputsByGroup/main.nf
dp-rice: As I commented in the last PR, I don't love having to list everything here manually, but c'est la vie Nextflow.

schemas/qc_basic_stats.schema.json
```diff
@@ -0,0 +1,161 @@
...
+      "example": "sample_1"
+    },
+    {
+      "name": "group",
+      "title": "Group",
+      "description": "Group identifier assigned to this sample in the grouping TSV.",
+      "type": "string",
+      "constraints": { "required": true, "minLength": 1 },
+      "example": "group_1"
+    }
```
dp-rice: Not for this PR, but we're using these fields a lot. Does the schema framework we're using allow us to define common columns and include them?

schemas/qc_basic_stats.schema.json
```diff
@@ -0,0 +1,161 @@
...
+      "title": "Percent Duplicates",
+      "description": "Percent duplicates as measured by FASTQC.",
+      "type": "number",
+      "constraints": { "required": true, "minimum": 0, "maximum": 100 },
+      "example": 8
+    },
+    {
+      "name": "n_bases_approx",
+      "title": "Approximate Base Count",
+      "description": "Approximate number of base pairs in reads.",
```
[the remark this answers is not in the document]
willbradshaw: [Claude Code] Good question — this is a throughput metric, not a per-read metric. `mean_seq_len` is the average length of individual reads (e.g. 151 bp), while `n_bases_approx` is the total number of bases across all reads (≈ n_reads_single × mean_seq_len). Updated description in 8744b46f to clarify this.

schemas/qc_basic_stats.schema.json
willbradshaw: [Claude Code] The Frictionless Data Table Schema spec (v2.0) doesn't support field-level `$ref` or composition, so common columns have to be repeated in each schema file. A build-time generation step could reduce this repetition.

PR #671, opened 2026-02-20
Remove redundant duplicate_reads from DOWNSTREAM outputs

willbradshaw:
The `{GROUP}_duplicate_reads.tsv.gz` DOWNSTREAM output is a strict subset of `{GROUP}_validation_hits.tsv.gz` (same rows, fewer columns) and is thus redundant. This PR removes it from published outputs.

**Changes:**
- Stop publishing the duplicate_reads file by filtering it out of `dup_output_ch` in `downstream.nf`
- Remove `duplicate_reads` from expected-outputs lists in both `pyproject.toml` files
- Update the downstream snapshot test
- Bump version to `3.2.0.0-dev` (schema-level change: removing an output file)

Note: the `MARK_VIRAL_DUPLICATES` subworkflow still produces the reads file internally as an intermediate for `viral_hits_ch`, which feeds into `VALIDATE_VIRAL_ASSIGNMENTS` and `COUNT_READS_PER_CLADE`.


.github/workflows/manual-reset.yml
jo-delia-faraguna: Do we want to restrict who can run this to just admins? If so, update `docs/developer.md` to reflect this.

PR #673, opened 2026-02-21
Add index age tracking and rebuild infrastructure

willbradshaw:
Adds infrastructure for tracking about the freshness of the benchmark index, and a workflow for rebuilding it when it gets too stale.

The benchmark index at `s3://nao-testing/mgs-workflow-test/index-latest/` is used by integration tests but had no mechanism to detect staleness or to rebuild it in a controlled way. This PR adds both, gated behind a configurable `max-stable-index-age-days` setting (default: 90 days).

**Changes:**
- Add `bin/check_index_age.py` script and unit tests: reads the index `time.txt` from S3 and compares against the max age in `pyproject.toml`
- Add `check-index-age.yml` CI workflow: runs on PRs to `main`/`stable` as a release gate
- Add `rebuild-benchmark-index.yml` workflow: manually-triggered, runs INDEX nf-tests as preflight, builds the index, cleans up the work directory, and verifies freshness
- Add `workflow_call` trigger to `nf-test-workflows-index.yml` so the rebuild workflow can reuse it as a preflight gate
- Fix Groovy date format in INDEX, RUN, and DOWNSTREAM workflows: `YYYY` (week-year) → `yyyy` (calendar year) in `time.txt` timestamps
- Add `max-stable-index-age-days = 90` to `pyproject.toml`
- Clarify job queue placeholder in `configs/index.config`

**Security:**
- The rebuild workflow is `workflow_dispatch`-only, restricted to users with repo write access, and requires a typed confirmation string to prevent accidental triggers.
- AWS access uses the repo's existing OIDC role (`AWS_OIDC_ROLE_ARN`); no new credentials are introduced.
- The workflow writes only to the test bucket (`s3://nao-testing/`); promoting to the production index bucket (`s3://nao-mgs-index/`) is a separate manual step.


willbradshaw:
[Claude Code] Addressed review feedback from det-code-review:

- **C1**: Replaced the pre-build `aws s3 rm` with a post-build stale file cleanup. The workflow now records the build start time, and after a successful build removes any S3 objects older than that timestamp. On build failure, the old index remains intact.
- **C2**: Removed the `branch` input entirely. The workflow now always checks out the dispatch branch (`github.ref`), which is the same branch the preflight tests validate.
- **D1**: Updated `docs/developer.md` to reference the CI rebuild workflow instead of the manual process.

PR #684, opened 2026-03-07
Add COMBINE_SAMPLE_JSONS and CONCAT_JSON_BY_GROUP

willbradshaw:
Introduces new module and subworkflow for combining per-sample JSON files into per-group JSON outputs, paralleling the existing TSV concatenation functionality. In a future PR, these will be used to produce per-group FASTP JSONs in DOWNSTREAM.

**Changes:**
- `COMBINE_SAMPLE_JSONS` module: Python script that takes per-sample JSON files, extracts sample names from filenames, injects `sample` and `group` fields, and writes a combined JSON keyed by sample name
- `CONCAT_JSON_BY_GROUP` subworkflow: Filters input files by suffix, groups by group label, combines via `COMBINE_SAMPLE_JSONS`, and renames output — mirrors `CONCAT_BY_GROUP` for TSVs
- Pytest suite for `combine_sample_jsons.py` covering sample name extraction, JSON combining, and CLI entrypoint
- nf-tests for both the module and subworkflow with dynamic input-based verification
- Test data: minimal FASTP-like JSON fixtures in `test-data/combineSampleJsons/`


willbradshaw:
[Claude Code] Responses to the second review:

**C1** (switch to `shell:`/`!{...}`): The `shell:` directive is now deprecated in Nextflow. `script:` with `${...}` is the correct approach going forward.

**C2** (comment about `.gz` omission): Done — added a comment explaining that `.gz` matching is deliberately omitted since JSON outputs are never gzipped.

**C3** (docstring note about key overwrite): Done — added a note to the `combine_sample_jsons()` docstring that `sample` and `group` keys are unconditionally overwritten.

**T1** (pytest case for overwrite): Skipping — the overwrite is a side effect rather than explicitly desired behavior, so codifying it in a test would over-commit to the current implementation.

PR #686, opened 2026-03-07
Add FASTP JSON Schema and extend validation

willbradshaw:
- Extends existing schema validation code to handle JSON files using the `jsonschema` package.
- Adds & enforces a schema for FASTP JSON output, verified against fastp v0.23.4 C++ source and real pipeline output.
    - Note that schema fields reflect the use of FASTP in the pipeline, so fields that are produced by arguments hardcoded in the pipeline are marked as required.
    - Updated `test-data/downstream/empty/empty_sample_fastp.json` from minimal stub to realistic zero-reads output.
- Updated DOWNSTREAM workflow test CI to run on changes to schemas to ensure validation code is executed.

**Note:** during development of this PR, I realized much of the JSON handling code in `createEmptyGroupOutputs` is dead code that will never actually run. For the sake of keeping this PR minimal, I decided not to fix this here, but will do so in a later PR. (It's not actually breaking anything, it's just cruft.)

### DIFF
--- a/docs/developer.md
+++ b/docs/developer.md
@@ -279,7 +279,7 @@ All new output files should have a complete schema, and all existing schemas sho
 - A `primaryKey` entry describing a set of fields that are collectively guaranteed to uniquely identify each row (can often be a single field).
 - A `missingValues` entry listing permitted null values (typically `""` and `"NA"`).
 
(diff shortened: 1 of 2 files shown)


willbradshaw:
@mikemc I let Claude Code post replies to your comments since its drafts seemed good; I did review all the drafts before posting. Thanks for your careful review here! Like CC says, the main change on our end is that I want the schema to reflect usage in the pipeline, rather than the full possibility space of FASTP outputs; this meant adding a few more required columns and removing some that are never produced in the pipeline.

PR #689, opened 2026-03-11
Replace ncbi-genome-download with NCBI datasets CLI

willbradshaw:
Replaces the unreliable `ncbi-genome-download` tool with NCBI's official `datasets` CLI for viral genome downloads in the INDEX workflow. Downloads are parallelized across child taxa for better fault tolerance against network errors and spot instance reclamation.

**Changes:**
- New `ENUMERATE_CHILD_TAXA` module: reads `nodes.dmp` to split downloads by child taxid (emits self for leaf taxa)
- New `DOWNLOAD_VIRAL_GENOMES` module: uses `datasets`/`dataformat` CLI per child taxon, with optional API key support
- New `PREPARE_VIRAL_METADATA` module + Python script: merges per-taxon metadata, adds `species_taxid` from virus taxonomy DB, and maps genome files to `local_filename` — maintaining full downstream compatibility with `filter-viral-genbank-metadata.py` and `ADD_GENBANK_GENOME_IDS`
- New `ncbi_datasets` container (`ncbi-datasets-cli=18.20.0`, `unzip=6.0` from conda-forge)
- Removed `params.ncbi_viral_params`; replaced with `params.assembly_source` (`"genbank"`/`"refseq"`), optional `params.ncbi_api_key`, and optional `params.download_virus_taxid` (defaults to `params.virus_taxid`)
- Updated `makeVirusGenomeDB` subworkflow wiring and `INDEX` workflow caller
- Updated all config files (`index.config`, `index-for-run-test.config`, `tests/configs/index.config`) and `docs/config.md`
- Deleted `modules/local/downloadViralNCBI/` and `containers/ncbi_genome_download.yml`
- 11 pytest tests for `prepare-viral-metadata.py` (all passing)
- INDEX nf-test passes end-to-end with Hepatitis D virus test case


willbradshaw (commented):

## Correctness

- **C1** (**issue**): `DOWNLOAD_VIRAL_GENOMES` will fail for child taxa that have no assemblies in the chosen `assembly_source`. When `datasets download` returns no results, it exits non-zero or produces no `output.zip`, causing `unzip` to fail. Since `metadata.tsv` is a non-optional output, the entire process errors out and blocks the pipeline. Consider adding a guard that produces an empty `metadata.tsv` (header only) and skips the unzip/flatten steps when `datasets` finds nothing — or use `errorStrategy 'ignore'` at the process level (though that's less precise).

- **C2** (**issue**): CHANGELOG formatting: line 7 is missing a blank line before the `# v3.2.0.2` heading. The current text runs the last bullet directly into the next version header, which will render incorrectly in markdown.

- **C3** (**suggestion**): `test-data/tiny-index/output/input/index-params.json` still contains `"ncbi_viral_params"` (line 18) and lacks the new params (`assembly_source`, `ncbi_api_key`, `download_virus_taxid`). This is a snapshot file — if the INDEX nf-test generates a new snapshot that replaces it, this is fine. But if the snapshot comparison is against this file, it will fail. Please confirm the INDEX nf-test passes with updated snapshots, or update this file in the PR.

- **C4** (**suggestion**): The `api_key` parameter is interpolated directly into the shell script via Nextflow's `${api_key}` without quoting. While NCBI API keys are typically alphanumeric, this is a shell injection vector in principle. Consider passing it via an environment variable:
  ```groovy
  script:
      def env_block = api_key ? "export NCBI_API_KEY='${api_key}'" : ""
      """
      ${env_block}
      datasets download genome taxon ${taxid} \
          --assembly-source ${assembly_source} \
          --include genome \
          ${api_key ? '--api-key "$NCBI_API_KEY"' : ''} \
          ...
      """
  ```
  (The same technically applies to `taxid` and `assembly_source`, but those are system-derived values.)

- **C5** (**nit**): In `ENUMERATE_CHILD_TAXA`, the `nodes.dmp` parsing uses `split("\\t|\\t")` followed by `.rstrip("\\t|")`. The NCBI `nodes.dmp` format uses `\t|\t` as delimiter (including a trailing `\t|\n`), so this works — but `fields[0]` is the raw taxid with no surrounding `|`, so the `rstrip` on `child_id` is a no-op. Not a bug, just unnecessary.

## Testing

- **T1** (**suggestion**): `ENUMERATE_CHILD_TAXA` has inline Python logic (parsing `nodes.dmp`, the leaf-taxon fallback) but no unit tests. Consider extracting the logic into a standalone Python script (consistent with the pattern used for `prepare-viral-metadata.py`) so it can be tested with pytest — or at minimum add a comment explaining that the INDEX nf-test covers it end-to-end.

- **T2** (**nit**): `test_gzipped_metadata_input` (line 171 of `test_prepare_viral_metadata.py`) creates a `.tsv.gz` file using `df.to_csv(meta_path, ...)`, which writes uncompressed text despite the `.gz` extension. Pandas infers compression from the suffix, so this actually works — but it's worth a quick comment to clarify that pandas handles the gzip transparently, since the code reads with `pd.read_csv()` which also auto-detects.

## Documentation

**No issues found.** `docs/config.md` is updated to document the new params (`assembly_source`, `ncbi_api_key`, `download_virus_taxid`) and remove `ncbi_viral_params`. The CHANGELOG entries clearly describe the changes.

## Quality

- **Q1** (**suggestion**): `ENUMERATE_CHILD_TAXA` only enumerates **direct** children (depth 1) of the given taxid. For the production case where `virus_taxid = 10239` (all Viruses), direct children are high-level taxa like *Duplodnaviria*, *Riboviria*, etc. — roughly 10–15 taxa. Each `DOWNLOAD_VIRAL_GENOMES` call then downloads **all** assemblies under that entire subtree (since `datasets download genome taxon` is recursive). This means the parallelization is quite coarse-grained — a single child like *Riboviria* would still contain the vast majority of viral genomes. Is this the intended granularity? If the goal is better fault tolerance via smaller chunks, consider enumerating at a deeper level (e.g., depth 2 or 3) or using `--assembly-level` filters.

- **Q2** (**nit**): The old `DOWNLOAD_VIRAL_NCBI` process used `label "max"` (highest resource allocation), while the new `DOWNLOAD_VIRAL_GENOMES` uses `label "small"`. This makes sense for leaf-taxon downloads, but for the production config where each child of 10239 could contain hundreds of thousands of genomes, the `small` label may be insufficient for disk/memory. Worth verifying against production data sizes.

## Risks

- **R1** (**suggestion**): The old test configs used `--assembly-level complete` to restrict downloads to complete assemblies. The new implementation drops this filter entirely — `datasets download genome taxon` will now return all assembly levels (complete, chromosome, scaffold, contig). For the test case (Hepatitis D virus 1, taxid 2847173) this is probably fine, but for the production config this is a behavioral change that could significantly increase the number of downloaded genomes. The PR description doesn't mention this change. If intentional, please document the rationale; if not, consider adding `--assembly-level complete` to the `datasets` command or making it configurable.

- **R2** (**suggestion**): The `DOWNLOAD_VIRAL_GENOMES` process has no explicit `errorStrategy` or `maxRetries`. The profile-level defaults (`retry` with `maxRetries = 1`) will apply, but given that network downloads to NCBI are the primary failure mode this PR aims to address, consider whether `maxRetries = 1` is sufficient or whether this process should have a higher retry count.

willbradshaw:
[Claude Code] Responding to all review comments:

**G1** (gemini: `--no-progress`): Done — added `--no-progress` to the `datasets download` command.

**G2** (gemini: assembly count): Done — changed to `$(( $(wc -l < metadata.tsv) - 1 ))` to subtract the header row.

**C1** (empty download guard): Done — wrapped the download/unzip/flatten in an `if datasets download ...; then ... else ...` block that produces a header-only `metadata.tsv` and empty `genomes/` directory on failure.

**C2** (CHANGELOG blank line): Done — added missing blank line before `# v3.2.0.2`.

**C3** (index-params.json): Done — replaced `"ncbi_viral_params": "--section refseq --assembly-level complete --taxids 2847173"` with the three new params: `"assembly_source": "refseq"`, `"ncbi_api_key": ""`, `"download_virus_taxid": "2847173"`.

**C4** (API key via env var): Done — now exports `NCBI_API_KEY` env var and references it as `"$NCBI_API_KEY"` in the command, avoiding direct interpolation.

**C5** (no-op rstrip): Done — removed the no-op `.rstrip()` as part of T1 extraction.

**R1** (`--assembly-level complete`): Done — re-added `--assembly-level complete` to the `datasets download` command to match previous behavior.

**T1** (extract ENUMERATE_CHILD_TAXA to standalone script): Done — extracted inline Python into `enumerate-child-taxa.py` with full pytest suite (9 tests passing) and nf-test (2 tests passing). Module refactored from `script:` to `shell:` block.

**T2** (pandas gzip comment): Done — added clarifying comment that pandas handles gzip compression/decompression transparently based on file suffix.

**Q1** (deeper parallelization): Created issue #691 to track this for future work.

**Q2** (`small` label): Confirmed — `small` label plus defaults are fine for the per-child-taxon granularity.

**R2** (retry settings): Confirmed — default retry settings (`maxRetries = 1`) are fine for now.

willbradshaw (commented):

## Correctness

- **C1** (**issue**): In `modules/local/downloadViralGenomes/main.nf`, the `if datasets download ...` pattern treats **all** non-zero exit codes as "no assemblies found", silently producing a header-only metadata file. This means transient failures (network errors, API rate limits, NCBI outages) are indistinguishable from taxa that genuinely have no assemblies. Because the process exits 0 in the `else` branch, Nextflow's `errorStrategy "retry"` will not retry these failures. In production, this could silently produce an incomplete genome database. Consider checking `datasets`' exit code or stderr to distinguish "no data" from actual errors — for example, `datasets` returns exit code 0 with an empty zip when there are no results, vs. a non-zero exit code for network/API failures. If the intent is to tolerate missing data for some taxa, the `else` branch should only fire on a known "no data" condition, and genuine errors should propagate.

- **C2** (**suggestion**): `--assembly-level complete` is hardcoded in `downloadViralGenomes/main.nf:20`. The old `ncbi_viral_params` passthrough allowed configuring this per deployment. If complete-only is the intended production behavior, this is fine — but it's a behavioral change worth calling out explicitly in the PR description, since the production `index.config` previously passed only `--section genbank` to `ncbi-genome-download` (whose default assembly level may have differed).

## Testing

- **T1** (**suggestion**): There are no nf-tests for `DOWNLOAD_VIRAL_GENOMES` or `PREPARE_VIRAL_METADATA` as individual modules. The INDEX workflow nf-test provides end-to-end coverage, and the pytest suite for `prepare-viral-metadata.py` is solid (11 tests covering mapping, missing files, gzipped input, symlinks). However, a module-level nf-test for `PREPARE_VIRAL_METADATA` (with canned inputs) would make it easier to catch regressions without running the full INDEX workflow. `DOWNLOAD_VIRAL_GENOMES` is harder to unit-test (NCBI dependency), so end-to-end coverage is reasonable there.

- **T2** (**nit**): In `test_enumerate_child_taxa.py`, the `test_parametrized_hierarchy` parametrized test (lines 93–106) duplicates the cases already covered by `test_parent_with_multiple_children`, `test_parent_with_one_child`, and `test_leaf_taxon_returns_self`. Consider removing the parametrized test or the individual tests to avoid redundancy.

## Documentation

**No issues found.** `docs/config.md` is updated to document the new `assembly_source`, `ncbi_api_key`, and `download_virus_taxid` parameters. The CHANGELOG accurately describes the changes. Inline comments in the Nextflow modules explain the pipeline step numbering.

## Quality

- **Q1** (**nit**): In `prepare-viral-metadata.py` lines 125–129, the `local_filename` lambda captures `accession_to_file` and `output_genomes_dir` from the enclosing scope. A more explicit approach would be to use `metadata["assembly_accession"].map(accession_to_file)` to create the filename column, then prepend the directory — this avoids the `KeyError` risk if the lambda is called on an accession not in the dict (though `dropna` on line 133 handles it, the current code would raise `KeyError` before reaching `dropna` for any row where `acc in accession_to_file` is `True` but the key was somehow removed). Minor readability preference.

## Risks

- **R1** (**suggestion**): In `downloadViralGenomes/main.nf:13`, the NCBI API key is embedded in the generated `.command.sh` via Groovy string interpolation (`"export NCBI_API_KEY='${api_key}'"`). This means the key is visible in plain text in the Nextflow work directory's `.command.sh` file and potentially in task trace logs. NCBI API keys are low-sensitivity (rate-limit tokens, not auth credentials), so this is acceptable for now, but consider using Nextflow's `secret` directive or passing the key solely via environment variable in the process definition (`env NCBI_API_KEY`) if stricter credential handling is desired in the future.

- **R2** (**suggestion**): The `taxid` and `assembly_source` values are interpolated directly into shell commands in `downloadViralGenomes/main.nf:18–19`. These come from config files and taxonomy data (trusted sources), so the injection risk is low. However, for defense-in-depth, consider validating that `assembly_source` is one of `genbank`/`refseq` and that `taxid` is numeric, either in the workflow caller or in the module script.

willbradshaw (commented):

## Correctness

- **C1** (**suggestion**): In `DOWNLOAD_VIRAL_GENOMES` (line 14), the API key is embedded in the script via `export NCBI_API_KEY='${api_key}'`. The `${api_key}` is Nextflow string interpolation, so the literal key is baked into `.command.sh` before bash ever runs. If the key contains a single quote character, this produces a bash syntax error (`export NCBI_API_KEY='abc'def'`). Consider using Nextflow's `env` process directive or `secret` directive instead of string interpolation to avoid both the quoting fragility and the key appearing in `.command.sh`. (Acknowledged as low-priority in #693 since NCBI API keys are rate-limit tokens, not auth credentials — but the quoting bug is a separate correctness concern.)

- **C2** (**suggestion**): In `DOWNLOAD_VIRAL_GENOMES`, `extra_args` (line 36) is interpolated directly into the shell command with no quoting or validation. While the other inputs (`taxid`, `assembly_source`) are validated, `extra_args` is a free-form string. If a config value inadvertently includes shell metacharacters (e.g. unmatched quotes, semicolons), the command will behave unexpectedly. This is low-risk since it's set in trusted config files, but consider adding a comment documenting that this parameter must contain only valid CLI flags, or wrapping it in validation similar to the other inputs.

- **C3** (**nit**): In `DOWNLOAD_VIRAL_GENOMES`, the empty-taxon path (lines 46–47, 54–55) creates `genomes/` with `mkdir -p` but never populates it. The `optional: true` on the output handles this correctly — just noting that the `mkdir -p genomes` on the empty-result paths is dead code since the `genomes/*.fna.gz` glob won't match regardless.

## Testing

- **T1** (**suggestion**): The `DOWNLOAD_VIRAL_GENOMES` nf-test makes a live NCBI API call (`datasets download genome taxon 2847173`). If this test is part of the regular CI suite, it introduces a flaky external dependency — NCBI rate limits, transient outages, or network issues can cause spurious failures. Consider whether this should be tagged to run only in a subset of CI jobs, or whether a stub/mock test could complement the live test. (If the test already runs only in the `test-workflows-index` job that's expected to hit external services, this is fine — just verify that expectation.)

- **T2** (**nit**): `test_enumerate_child_taxa.py` covers the core logic well (6 test cases). One minor gap: there's no test for a malformed `nodes.dmp` line (e.g. fewer than 2 tab-delimited fields). The code handles this with `if len(fields) < 2: continue`, but it's untested. Low priority since the input comes from NCBI's own taxonomy dump.

## Documentation

**No issues found.** Config docs (`docs/config.md`) are updated for all new/changed parameters. The CHANGELOG is comprehensive and describes both the user-facing changes and the architectural motivation. The PR description clearly explains the design (parallel per-child-taxon downloads for fault tolerance).

## Quality

- **Q1** (**suggestion**): `enumerate_children()` reads the entire `nodes.dmp` file line-by-line for a single parent lookup. In production, `nodes.dmp` is ~250MB with ~2.5M lines. This is fine for a one-shot call, but if the function were ever reused, it would benefit from building a parent→children index in one pass. Not actionable now — just noting the O(n) scan design.

- **Q2** (**nit**): `prepare-viral-metadata.py` uses `match_genomes_to_accessions` which does a linear scan of `genome_files` for each accession (`O(accessions × files)`). For the full Viruses taxon this could be ~200K accessions × ~200K files. Consider building a `{prefix: filename}` dict in one pass instead of filtering the full list per accession. This would change `O(n×m)` to `O(n+m)`.

## Risks

- **R1** (**suggestion**): This PR changes the parallelism model from a single monolithic `ncbi-genome-download` process (label `"max"`) to many small per-child-taxon `DOWNLOAD_VIRAL_GENOMES` processes (label `"small"`). For the full Viruses taxon (10239), `enumerate_children` finds only *direct* children (one level deep), which is likely ~15–20 major virus orders. This is a reasonable number. However, if `download_virus_taxid` is set to a taxid with hundreds of direct children, or if the enumeration strategy is later changed to multi-level, the number of parallel NCBI API calls could become very large. Consider documenting the expected scale (direct children only, typically <100 taxa) in the config parameter description or module comment.

- **R2** (**suggestion**): The output format of `PREPARE_VIRAL_METADATA` must exactly match what `FILTER_VIRAL_GENBANK_METADATA` and `ADD_GENBANK_GENOME_IDS` expect — specifically the `local_filename`, `species_taxid`, `assembly_accession`, and `taxid` columns. I've verified these columns are present and the names match. However, the `local_filename` values now point through a chain of symlinks (`ncbi_genomes/<file> → input_genomes/<file> → Nextflow-staged symlink → actual file`). This works correctly but is more fragile than the previous single-directory approach. If any intermediate work directory is cleaned up (e.g. aggressive `-w` cleanup or manual deletion), downstream processes could see broken symlinks. The PR description should note this symlink chain as a design trade-off.

PR #690, opened 2026-03-11
Migrate all modules from shell: to script: blocks

willbradshaw:
Nextflow's `shell:` block syntax (`!{var}` for Nextflow interpolation, `$var` for Bash) is deprecated in favor of `script:` blocks (`${var}` for Nextflow, `\$var` for Bash). This PR migrates all 57 modules in `modules/local/` to use `script:` blocks, completing the transition. This is a pure refactoring with no intentional behavior change, plus bugfixes for escaping issues caught during the migration.

**Changes:**
- Converted all Nextflow modules from `shell:` (`'''`/`!{var}`) to `script:` (`"""`/`${var}`) blocks.
- Fixed a `$RANDOM` escaping bug in `subsetFastn` where triple-backslash (`"\\\$RANDOM"`) would have prevented Bash variable expansion; corrected to `"\$RANDOM"`
- Fixed an awk debug message in `extractViralHitsToFastq` where `$colIndex` was being evaluated as a field reference instead of printing the numeric column index


willbradshaw (commented):

## Correctness

- **C1** (**suggestion**): In `modules/local/extractViralHitsToFastq/main.nf`, the awk debug message on the success path (`colIndex != -1`) has a subtle semantic change. The original code had `print "'seq_id' column index: $colIndex"` where `$colIndex` was inside an awk double-quoted string (printed as literal text `$colIndex`). The new code has `print "'\\''seq_id'\\'' column index: " \$colIndex` where `$colIndex` is outside the string, so awk evaluates it as a field reference (printing the field value, e.g. `seq_id`). Neither version prints the numeric column index, and this is only a diagnostic message, but it's technically a behavioral change in a PR described as "no behavior change." Consider either using `colIndex` (without `$`) to print the numeric index, or documenting this as a minor intentional change.

- **C2** (**nit**): The `test-modules` CI check failed due to an nf-test plugin update infrastructure error, not due to the PR's changes. The module tests never actually ran. Consider re-running the check to get actual test coverage of the converted modules — this is the most important test suite for validating the escaping changes.

## Testing

**No issues found.** All workflow and subworkflow test suites pass (test-subworkflows, test-workflows-run, test-workflows-downstream, test-workflows-index). The test-modules failure is an infrastructure issue unrelated to the code changes. Since this is a pure refactoring, no new tests are needed — the existing snapshot tests cover the converted modules.

## Documentation

**No issues found.** The CHANGELOG entry is clear and accurate. The PR description thoroughly categorizes the changes by complexity and documents the `$RANDOM` bugfix.

## Quality

- **Q1** (**nit**): The PR description states "This is a pure refactoring with no behavior change," but it also documents the `$RANDOM` escaping bugfix in `subsetFastn` (which _is_ a behavior change, albeit a bugfix). Consider rewording to something like "This is a pure refactoring with no intentional behavior change, plus a bugfix for `$RANDOM` escaping in `subsetFastn`."

## Risks

**No issues found.** The conversion is mechanical and well-categorized by complexity. The highest-risk modules (awk/sed with field references and embedded Python/R) use correct escaping patterns. The bracken awk over-escaping was already caught and fixed (commit 57bf4d4e). All escaping patterns I verified produce identical bash output to the originals.

willbradshaw:
Note for reviewers: this is a wide but shallow PR that makes similar small changes to lots of scripts. Since...

- We've gone through several rounds of AI code review and AI response-to-review;
- I've reviewed the changes manually myself;
- All CI passes;

...I think it's fine to skip detailed third-party review in this case. As such, I recommend reviewing the new CI, spot-checking a couple of scripts, and then calling it done without detailed review.

PR #717, opened 2026-04-08
Make run_nf_test_parallel.py repo root detection symlink-safe

jo-delia-faraguna:
## Summary
- Use `Path.absolute()` instead of `Path.resolve()` in `run_nf_test_parallel.py` so symlinked copies of the script compute the repo root relative to the symlink location, not the symlink target

Ref: [COMP-1586](https://linear.app/securebio-nao/issue/COMP-1586/switch-to-use-parallelized-nf-test-ci-implementation)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # v3.2.1.3-dev
 
-- Made repo root detection in `run_nf_test_parallel.py` symlink-safe, allowing the script to be symlinked from other repositories.
+- Make repo root detection in `run_nf_test_parallel.py` symlink-safe, allowing the script to be symlinked from other repositories.
 
 # v3.2.1.2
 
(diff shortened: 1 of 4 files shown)


jo-delia-faraguna:
Closing in favor of copying the scripts directly into consuming repos. Symlinking into bin/ will always break Nextflow module bundling, and the --repo-root chaining adds unnecessary complexity.

willbradshaw:
> Closing in favor of copying the scripts directly into consuming repos. Symlinking into bin/ will always break Nextflow module bundling, and the --repo-root chaining adds unnecessary complexity.

Hmm, is this true even if we use the tricks I was using for cross-module code in the old overhaul repo? ([example](https://github.com/naobservatory/chimera-detection-v2/blob/main/modules/local/filterSplitStrand/resources/usr/bin/filter_split_strand.py#L26-L31))

(Not necessarily recommending this, those tricks were pretty ugly, but it was convenient)

Issue #716, opened 2026-04-06
Extract hardcoded BBDuk masking params into shared config


jo-delia-faraguna:
## Context

In `workflows/index.nf:38`, the BBDuk masking parameters for viral genome filtering are hardcoded inline:

```groovy
virus_genome_params.putAll([k: "20", hdist: "3", entropy: "0.5", polyx_len: "10"])
```

The chimera detection repo (`nao-chimera-detection`) duplicates these values in its `cd_index.nf` workflow, which calls the same `MAKE_VIRUS_GENOME_DB` subworkflow. If these values change in mgs-workflow, the CD pipeline's copy will silently diverge.

## Suggestion

Extract these BBDuk defaults into a shared location (e.g., a config file or a `params` block in a dedicated config) so both mgs-workflow and downstream consumers like chimera-detection can import them from a single source of truth.

Generated with [Claude Code](https://claude.com/claude-code)

PR #723, opened 2026-04-14
Add mark_duplicates_similarity to rust-tools workspace

willbradshaw:
Add the similarity-based duplicate marking tool to the `rust-tools/` Cargo workspace, making it available in the `nao-rust-tools` Docker container for pipeline integration in a follow-up PR.

**Changes:**
- Create `rust-tools/mark_duplicates_similarity/` crate: copy source from `post-processing/rust_dedup/`, refactoring CLI from positional args to clap (-i/-o flags) for consistency with other workspace tools
- Use pinned git dependency for `nao_dedup` library (rev = "404c708...") instead of the `submodule` path dependency, so `cargo build --workspace` fetches it at build time
- Convert all 8 Python integration tests to Rust (`tests/integration.rs`), following the `process_vsearch_cluster_output` testing pattern, and add two new tests suggested by the AI reviewer
- Update `nao-rust-tools.Dockerfile` to copy and verify the new binary

Note that, while this PR is large in terms of changed lines, the two major files are both closely based on pre-existing code:
- `rust-tools/mark_duplicates_similarity/src/main.rs` is a direct copy from `post-processing/rust_dedup/` (with a minor change in CLI arguments)
- `rust-tools/mark_duplicates_similarity/tests/integration.rs` is a one-to-one port of the Python tests from `post-processing/rust_dedup/` (plus two new tests suggested by the AI reviewer)

The existing code in `post-processing/rust_dedup` is left untouched in this PR; it should be removed once `securebio/nao-mgs-partner-reports` is updated to use the new files.

### DIFF
--- a/rust-tools/README.md
+++ b/rust-tools/README.md
@@ -6,6 +6,7 @@ Rust utilities that run as part of the Nextflow pipeline. These are compiled int
 ## Workspace Tools
 
 - **mark_duplicates** — Marks duplicate alignments in SAM/BAM data
+- **mark_duplicates_similarity** — Marks similarity-based duplicates among alignment-unique reads using [nao-dedup](https://github.com/securebio/nao-dedup)
 - **process_vsearch_cluster_output** — Processes tabular output from VSEARCH clustering
 
 ## External Tools
@@ -22,6 +23,7 @@ development workflow, and how to add new tools.
 
 ## Note on post-processing/rust_dedup/
 
-A separate Rust tool exists in `post-processing/rust_dedup/` for similarity-based
-duplicate marking. That tool is currently a standalone post-processing utility (not
-part of the Nextflow pipeline) and may be consolidated here in a future PR.
+The similarity-based duplicate marking tool has been copied into this workspace as
+`mark_duplicates_similarity`. The original source in `post-processing/rust_dedup/`
+is retained for backwards compatibility with `securebio/nao-mgs-partner-reports`
+and will be removed once that dependency is migrated.
--- a/rust-tools/mark_duplicates_similarity/src/main.rs
+++ b/rust-tools/mark_duplicates_similarity/src/main.rs
@@ -1,0 +1,240 @@
+use anyhow::{bail, Context, Result};
+use clap::Parser;
+use flate2::read::GzDecoder;
+use flate2::write::GzEncoder;
+use flate2::Compression;
+use nao_dedup::{DedupContext, DedupParams, MinimizerParams, ReadPair};
+use std::collections::HashMap;
+use std::fs::File;
+use std::io::{BufRead, BufReader, BufWriter, Write};
+use std::time::Instant;
+
+// ------------------------------------------------------------------------------------------------
+// ARGUMENT PARSING
+// ------------------------------------------------------------------------------------------------
+
+/// Mark similarity-based duplicates among alignment-unique reads
+#[derive(Parser)]
+#[command(author, version, about, long_about = None)]
+struct Args {
+    /// Input gzipped TSV file path
+    #[arg(short, long)]
+    input: String,
+    /// Output gzipped TSV file path
+    #[arg(short, long)]
+    output: String,
+}
+
+// ------------------------------------------------------------------------------------------------
+// HELPER FUNCTIONS
+// ------------------------------------------------------------------------------------------------
+
+fn find_column(header_fields: &[&str], name: &str) -> Result<usize> {
+    header_fields
+        .iter()
+        .position(|&f| f == name)
+        .with_context(|| format!("Missing required column: {}", name))
+}
+
+// ------------------------------------------------------------------------------------------------
+// MAIN
+// ------------------------------------------------------------------------------------------------
+
+fn main() -> Result<()> {
+    let args = Args::parse();
+    let input_path = &args.input;
+    let output_path = &args.output;
+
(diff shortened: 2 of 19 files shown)


willbradshaw (commented):

## Correctness

**No issues found.** The Rust source in `mark_duplicates_similarity/src/main.rs` is a faithful port of the original `post-processing/rust_dedup/src/similarity_duplicate_marking.rs`, with the only structural change being the migration from positional CLI args to clap `--input`/`--output` flags. The two-pass algorithm (pass 1: build dedup context and alignment-dup counts; pass 2: write output with similarity exemplar and group size columns) correctly handles alignment duplicates (NA), similarity duplicates (exemplar pointer with NA group size), and surviving exemplars (exemplar = self with aggregated group size). Header parsing, column validation, and malformed-line detection all look correct.

## Testing

- **T1** (**suggestion**): The integration tests cover no-duplicates, alignment-only duplicates, similarity duplicates, empty input, single read, column order preservation, multi-level group sizes, missing columns, and malformed lines -- good breadth. However, there is no test for a read whose `prim_align_dup_exemplar` value does not appear as any row's `seq_id` (i.e. an orphaned alignment duplicate pointing to a nonexistent exemplar). This is arguably a data integrity violation in the input, but the current code would silently process it: the orphaned exemplar ID would be passed to `ctx.get_cluster_id()` in the group-sizes loop even though it was never fed to `ctx.process_read()`. Consider adding a test (or at minimum a comment) documenting the expected behavior for malformed alignment-duplicate references.

## Documentation

**No issues found.** The PR updates `docs/output.md` with the new `experimental/` directory and its guarantees, `docs/versioning.md` with the experimental-output exemption, `rust-tools/README.md` with the new tool listing and the note about `post-processing/rust_dedup/` coexistence, and `CHANGELOG.md` with the new entry. The Rust source has clear section headers and inline comments explaining the two-pass algorithm.

## Quality

- **Q1** (**nit**): In `mark_duplicates_similarity/src/main.rs`, the variable name `au_id` in the group-sizes loop (line 153: `for (au_id, count) in &align_dup_counts`) is slightly misleading -- `align_dup_counts` is keyed by `prim_align_dup_exemplar` values, which are alignment-unique exemplar IDs, but the variable name `au_id` does not make this mapping relationship obvious. Consider renaming to `align_exemplar` or `prim_align_exemplar` for clarity.

## Risks

**No issues found.** The PR adds a new binary to the rust-tools container but does not wire it into any Nextflow process -- the experimental output channels are all `Channel.empty()`, making this a no-op addition. The pinned git dependency (`rev = "404c708..."`) is appropriately immutable. The nf-test exclude patterns for `experimental`/`experimental_downstream` correctly prevent the empty experimental directories from breaking expected-output assertions. The stacked-PR relationship with #716 is documented in the PR description.

_Reviewed after 3 iteration(s) of automated review-and-fix. 8 findings were automatically resolved._

willbradshaw:
[Claude Code] Responses to final review findings:

**T1** (orphaned alignment dup test): The tool's input is always produced by `MARK_ALIGNMENT_DUPLICATES`, which guarantees every `prim_align_dup_exemplar` value corresponds to a valid `seq_id`. Testing behavior on referentially-invalid input would document undefined behavior rather than a contract — not worth codifying.

[Will note: I don't find Claude's reasoning here convincing, but I also don't think T1 is worth doing, so I let it be.]

**Q1** (rename `au_id`): Intentionally kept as-is. The variable name matches the original source in `post-processing/rust_dedup/`, and the two copies need to stay in sync until the old one is removed. Diverging names would make future maintenance harder.

[Will note: I agree with Claude here that it's better to keep these in sync.]

rust-tools/mark_duplicates_similarity/src/main.rs
willbradshaw, line 1: This is almost entirely unchanged from `post-processing/rust_dedup`; the only change is a slight change in argument parsing to bring it into consistency with the other tools in `rust-tools` (with `--input` and `--output` instead of positional args.)

rust-tools/mark_duplicates_similarity/tests/integration.rs
willbradshaw, line 1: These are mostly just Rust versions of the Python tests currently in `post-processing/rust_dedup`, plus a couple of extra expected-failure tests suggested by the AI reviewer. I don't think it's necessary to review these in depth.

rust-tools/README.md
```diff
@@ -22,6 +23,7 @@ development workflow, and how to add new tools.
 
 ## Note on post-processing/rust_dedup/
 
-A separate Rust tool exists in `post-processing/rust_dedup/` for similarity-based
-duplicate marking. That tool is currently a standalone post-processing utility (not
-part of the Nextflow pipeline) and may be consolidated here in a future PR.
+The similarity-based duplicate marking tool has been copied into this workspace as
+`mark_duplicates_similarity`. The original source in `post-processing/rust_dedup/`
+is retained for backwards compatibility with `securebio/nao-mgs-partner-reports`
+and will be removed once that dependency is migrated.
```
dp-rice, line 29: Could you make a migration issue in partner reports?

dp-rice (approved):
Just skimmed, but seems fine.

rust-tools/README.md
willbradshaw: Filed https://github.com/securebio/nao-mgs-partner-reports/issues/102?reload=1

PR #724, opened 2026-04-14
Add MARK_SIMILARITY_DUPLICATES process and subworkflow integration

willbradshaw:
This PR integrates the similarity-based duplicate-marking Rust binary from #723 into the workflow via a new `MARK_SIMILARITY_DUPLICATES` Nextflow process and modified `MARK_VIRAL_DUPLICATES` subworkflow, plus their associated `nf-test` tests. Wiring the output into the DOWNSTREAM experimental channel is deferred to a follow-up PR.

PR #725, opened 2026-04-14
Wire similarity dedup into DOWNSTREAM experimental output

willbradshaw:
Small PR To wire the similarity-based duplicate marking output from #724 into the DOWNSTREAM workflow's `experimental_downstream` channel.

I've run the workflow-level test manually and confirmed it creates the expected file in the expected place with the expected content. It's not included in `expected-outputs` in `pyproject.toml` so no changes needed there.


willbradshaw (commented):

## Correctness

**No issues found.** The wiring logic correctly assigns `Channel.empty()` for ONT and the subworkflow output for short-read, matching the existing branching pattern in `downstream.nf`. The `experimental_downstream` emit replaces the previous `Channel.empty()` with `sim_dup_ch`, which correctly carries the `[label, file]` tuple structure expected by the Nextflow output block.

## Testing

- **T1** (**suggestion**): In the short-read test (line 60), `experimentalFiles` collects relative path strings via `experimentalDir.relativize(file)`, while in the header-only test (line 106), `experimentalFiles` collects the raw file objects. This inconsistency means the short-read test asserts on a path string (`== "tt1_duplicate_reads_similarity.tsv.gz"`) while the header-only test calls `.linesGzip` on a file object. Both work, but the collection logic should be consistent -- either collect strings in both (and assert on the filename in the header-only test) or collect file objects in both (and compare `.toString()` in the short-read test). This would make the tests easier to maintain.

## Documentation

**No issues found.** The parent PR (#724) already added documentation for similarity-based duplicate marking in `docs/downstream.md`, including its experimental status. The CHANGELOG entry accurately describes the change. The `docs/output.md` experimental outputs section already covers `experimental_downstream/`.

## Quality

**No issues found.** The change is minimal and follows the established patterns in the workflow: the ONT/short-read branching mirrors the existing `dup_output_ch` and `clade_counts_ch` assignments, and the channel is wired through cleanly to the emit block.

## Risks

**No issues found.** The PR description explains this is a stacked PR wiring an experimental output. The output goes into `experimental_downstream/`, which is explicitly documented as unstable and excluded from expected-output tracking. No stable outputs, interfaces, or defaults are changed.

_Reviewed after 1 iteration(s) of automated review-and-fix. 1 finding was automatically resolved._

willbradshaw:
[Claude Code] **T1**: The inconsistency is intentional — the short-read test needs relativized strings to assert on the filename, while the header-only test needs file objects to call `.linesGzip` for header inspection. Consolidating would require collecting file objects in both and then calling `.toString()` for the filename check, which would be less readable than the current approach.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 - Add `experimental/` and `experimental_downstream/` output directories for staging new outputs that are not yet guaranteed to be stable across point releases.
 - Add `mark_duplicates_similarity` to the rust-tools workspace and Docker container for similarity-based duplicate marking.
 - Add `MARK_SIMILARITY_DUPLICATES` Nextflow process and integrate into `MARK_VIRAL_DUPLICATES` subworkflow.
+- Wire similarity dedup output into `experimental_downstream` channel for short-read platforms.
```
dp-rice: Nit: Assuming this completes the feature, it would be good consolidate/reorganize these entries to be cleaner and more user-focused.

dp-rice (approved):

CHANGELOG.md
willbradshaw: Consolidated

PR #726, opened 2026-04-14
Add sentinel file to RUN workflow

willbradshaw:
- Adds self-certification of completion in RUN via writing a `sentinel.json` file to `logging/` after validating that all expected output files have been published. External systems can atomically check for this file instead of probing individual outputs.
- The sentinel process uses a Nextflow `exec:` block to check published files against a list of expected output filepaths via `file().exists()`, which handles both local and S3 paths natively. It retries with exponential backoff (15s, 30s, 1m, ..., 16m) up to a configurable timeout (`params.sentinel_max_wait_mins`, default 32). The `exec:` block consumes minimal resources — no container is launched; it just sleeps and checks file existence.
- The sentinel file contains `runStartedAt` & `runCompletedAt` timestamps. The former of these replaces the old `logging/time.txt`, which is removed.
- The part of the RUN workflow that handles outputs and publishing is now quite long: longer than the part that actually runs processes. This should get cleaned up in a future PR, probably via some modularization into subworkflows. However, I didn't want to add that complexity to this PR.


willbradshaw (commented):

## Correctness

**No issues found.** The sentinel process correctly collects only the channels whose files appear in the `expected-outputs-run` list, validates their existence with retry/backoff to handle asynchronous publishing, and writes the JSON marker to `task.workDir`. The regex for parsing pyproject.toml uses `(?s)` DOTALL mode, which is functionally equivalent to the test helper's `[\s\S]*?` approach. The deliberate exclusion of `sentinel.json` from the expected-outputs list in `pyproject.toml` avoids a circular dependency.

## Testing

**No issues found.** Module tests cover the key axes: illumina (with shortread-extra patterns), ONT (without shortread-extra), S3 output directories, and the timeout path when files are missing. All three workflow-level test cases (empty input, shortread, ONT) assert sentinel validity via the `assertSentinelValid` helper. The `getObservedOutputs` exclusion of "sentinel" keeps the expected-vs-observed output comparison stable while the dedicated helper verifies sentinel content separately.

## Documentation

**No issues found.** `docs/output.md` documents `sentinel.json` in the logging section with a clear description of its purpose and contents. The `sentinel_max_wait_mins` parameter is discoverable via commented-out entries in both `configs/run.config` and `configs/run_ont.config`. The CHANGELOG entry covers both the sentinel file and the new parameter.

## Quality

- **Q1** (**nit**): In `tests/workflows/run.nf.test`, the `getObservedOutputs` exclusion uses a substring match (`relPath.contains("sentinel")`), which is broader than necessary -- it would exclude any future file whose path happens to contain the substring "sentinel". Consider matching the specific filename (`relPath.endsWith("sentinel.json")`) for precision, consistent with how you would want to handle this if the output set grows.

## Risks

**No issues found.** The PR description thoroughly addresses the key design decisions: the `exec:` block running on the head node, the exponential backoff strategy and default timeout, and the rationale for verifying published files rather than relying on channel completion alone. The new `sentinel_max_wait_mins` parameter defaults to 32 in-code (via null check in `workflows/run.nf`) rather than in config files, which is documented via the commented-out config entries.

_Reviewed after 4 iteration(s) of automated review-and-fix. 12 findings were automatically resolved._

willbradshaw:
(Dismissing Q1, this is consistent with the other exclusions)

willbradshaw (commented):
@jo-delia-faraguna Claude got confused and posted a review here for a different PR, please ignore.

configs/run.config
```diff
@@ -25,6 +25,9 @@ params {
     random_seed = "17310" // Random seed for non-deterministic processes. Empty string -> random seed.
     taxid_artificial = "81077" // Parent taxid for artificial sequences in NCBI taxonomy
 
+    // Sentinel file settings
+    // sentinel_max_wait_mins = 32 // Max minutes to wait for output files to be published before writing sentinel
```
jo-delia-faraguna: **[nit]** Do you want to specify this should reasonably be `15*(2**x)`seconds for a positive natural number _x_?

modules/local/writeSentinel/main.nf
jo-delia-faraguna, line 1: **[nit]** Do you want to rename this something `RUN`-specific? Or do you plan to reuse this function for `DOWNSTREAM` as well?

modules/local/writeSentinel/main.nf
```diff
@@ -0,0 +1,67 @@
...
+        }
+        def maxWaitMs = (config.max_wait_mins as long) * 60 * 1000
+        def intervalMs = 15000L
+        def totalWaitedMs = 0L
+        def missing = expected.findAll { !file("${config.output_dir}/${it}").exists() }
+        while (!missing.isEmpty()) {
+            if (totalWaitedMs >= maxWaitMs) {
+                error("Timed out after ${config.max_wait_mins} minutes waiting for " +
+                      "${missing.size()} published output file(s):\n  " +
+                      missing.join("\n  "))
```
jo-delia-faraguna: **[nit]** Consider a more informative error message

```suggestion
                error("Timed out after ${config.max_wait_mins} minutes waiting for " +
                      "${missing.size()}/${expected.size()} remaining published output file(s):\n  " +
                      missing.join("\n  "))
```

test-data/writeSentinelRun/pyproject_s3.toml
```diff
@@ -0,0 +1,11 @@
+# Fixture for S3 test: patterns match files that already exist in s3://nao-testing/tiny-test/
```
jo-delia-faraguna, line 1: **[nit]** This seems fine but it's a little strange to use this test set, which represents a small `RUN` _input_. I guess we don't have any stable `RUN` outputs in s3 and it's probably not worth it to add one just for this.

test-data/writeSentinelRun/pyproject_s3.toml
```diff
@@ -0,0 +1,11 @@
...
+[project]
+version = "0.0.0"
+
+[tool.mgs-workflow]
+expected-outputs-run = [
+    "raw/tiny-test_R1.fastq.gz",
+    "raw/tiny-test_R2.fastq.gz",
+]
+
+expected-outputs-run-shortread-extra = []
```
jo-delia-faraguna, line 11: **[nit]** Do you want to exercise this check as well for the s3 case by adding e.g. `raw/tiny-test_interleaved.fastq.gz`

tests/modules/local/writeSentinelRun/main.nf.test
```diff
@@ -0,0 +1,124 @@
...
+            assert process.success
+            def sentinel = new groovy.json.JsonSlurper().parse(
+                path(process.out.sentinel[0]).toFile()
+            )
+            assert sentinel.runStartedAt == "2026-01-01 00:00:00 UTC (+0000)"
+            assert sentinel.runCompletedAt != null
+        }
+    }
+
+    test("Should fail when expected outputs are missing") {
```
jo-delia-faraguna, line 99: **[suggestion]** Do you want to add a test that fails when the extra `illumina` output is missing?
```
input[3] = [
    output_dir: "${projectDir}/test-data/writeSentinel/output_ont",
    pyproject_path: "${projectDir}/test-data/writeSentinel/pyproject.toml",
    platform: "illumina",
    max_wait_mins: 1
]
```

modules/local/writeSentinel/main.nf
```diff
@@ -0,0 +1,67 @@
...
+process WRITE_SENTINEL {
+    executor 'local'
+    input:
+        val(ready)           // Dependency signal: collected items from all output channels
+        val(sample_names)    // List of sample names from samplesheet
+        val(start_time)      // Start time string
+        val(config)          // Map: output_dir, pyproject_path, platform, max_wait_mins
+    output:
+        path("sentinel.json"), emit: sentinel
+    exec:
```
jo-delia-faraguna: **[nit]** Can we reuse [getRunOutputSuffixes](https://github.com/securebio/nao-mgs-workflow/tree/main/modules/local/getRunOutputSuffixes) or [discoverRunOutput](https://github.com/securebio/nao-mgs-workflow/blob/main/subworkflows/local/discoverRunOutput/main.nf) at all? Probably not because this needs to run `exec` right?

modules/local/writeSentinel/main.nf
```diff
@@ -0,0 +1,67 @@
...
+        // Uses the same regex approach as the test helper getExpectedOutputs
+        // Assumes array values in pyproject.toml do not contain literal ] characters
+        def pyprojectText = file(config.pyproject_path).text
+        def keys = ["run"]
+        if (config.platform == "illumina") keys.add("run-shortread-extra")
+        def expected = []
+        for (wfKey in keys) {
+            def key = "expected-outputs-${wfKey}"
+            def sectionMatch = (pyprojectText =~ /(?s)${key} = \[(.*?)\]/)
+            if (sectionMatch) {
```
jo-delia-faraguna: **[suggestion]** Do you want to fail if the section is never matched? And exercise that path in a test if so?

jo-delia-faraguna: Probably not to match the behavior in [getRunOutputSuffixes#L40](https://github.com/securebio/nao-mgs-workflow/blob/main/modules/local/getRunOutputSuffixes/resources/usr/bin/get_run_output_suffixes.py#L40)

jo-delia-faraguna (approved):
LGTM, no blocking comments. I appreciate that reusing the existing toml parsing and expansion code may not be possible.

configs/run.config
willbradshaw: [Claude Code] Done — added a brief note in configs/run.config and configs/run_ont.config that reasonable values follow 15*(2**x)/60 min.

modules/local/writeSentinel/main.nf
willbradshaw: [Claude Code] Done — renamed to WRITE_SENTINEL_RUN (dir writeSentinelRun) to leave room for the parallel WRITE_SENTINEL_DOWNSTREAM on the stacked branch.

modules/local/writeSentinel/main.nf
willbradshaw: [Claude Code] Done — applied the suggestion inside the new shared SentinelUtils.waitForFiles helper.

test-data/writeSentinelRun/pyproject_s3.toml
willbradshaw: [Claude Code] On reflection, going from 2 to 3 expected files doesn't exercise a new code path in the existence-check loop — same iteration, just one more item. Skipping to avoid coupling the test to an extra S3 object.

tests/modules/local/writeSentinelRun/main.nf.test
willbradshaw: [Claude Code] Skipping to keep nf-test runtime down — the existing expect_fail case already exercises the timeout + missing-file path. Happy to revisit if we hit a missed regression.

modules/local/writeSentinel/main.nf
willbradshaw: [Claude Code] Right — those wrap a Python script in a process container and only expose filename suffixes, whereas this needs full expected paths (with wildcard expansion) evaluated synchronously in an exec: block so file().exists() can hit S3 from the head node. The new SentinelUtils refactor centralises the parser so WRITE_SENTINEL_DOWNSTREAM can reuse it.

modules/local/writeSentinel/main.nf
willbradshaw: [Claude Code] Agreed — staying consistent with get_run_output_suffixes.py:40 (silent empty list on missing key).

PR #737, opened 2026-04-23
Fix MASK_FASTQ_READS OOM by scaling memory with input size (#736)

### LINKED ISSUE #736: Resource re-tiering for ONT viral-read extraction on merged multi-GB libraries
Tracks a resource-tier adjustment in `EXTRACT_VIRAL_READS_ONT` surfaced by stress-testing `mgs-workflow` v3.2.1 on merged multi-GB ONT libraries (one fastq.gz per library instead of many small divs).

### Problem

- `MASK_FASTQ_READS` at `label "large"` (32 GiB) OOMs on merged ONT libraries larger than ~6 GB of gzipped FASTQ. Fails reliably on ~29 GB libraries.

### Proposed fix

Point-level PR with one change:
1. `MASK_FASTQ_READS`: replace `label "large"` with an input-size-aware memory closure (32 / 64 / 128 GiB based on gzipped FASTQ size). CPU allocation unchanged at 16.

No schema / output / results changes. Empirical evidence and threshold choice live in the PR description.

### Out of scope

Adjacent observations that are NOT part of this PR:
- **`EXTRACT_VIRAL_READS_ONT:FILTLONG` CPU reservation**: reserves 8 CPUs via `label "small"` but `filtlong` itself is single-threaded. The process pipes through `gzip` though, so filtlong+gzip can use ~2 cores in parallel; observed 5–40% of 8 cores reflects that. Dropping to `cpus = 1` would serialize them and could extend walltime. Deferred pending walltime measurement or a move to a middle tier (e.g. `xsmall`, 4 CPUs).
- `SORT_MINIMAP2_VIRAL` peaks at 0.84 ratio on 16 GiB, but `sort_tsv.py` wraps GNU sort which spills to disk when the buffer fills — can't OOM. A memory bump is a speed/disk-I/O optimization, not a correctness fix.
- `PROCESS_VIRAL_MINIMAP2_SAM` peaks at 0.55 ratio on 4 GiB, but memory is dominated by static reference DBs with a streaming merge-join — input-invariant and safe.

### Labels

enhancement, priority_2

dp-rice:
We're planning on modifying the ONT basecalling workflow to merge the hundreds of small "div" fastq files per sample into a single large one. In testing, I found that the very largest Zephyr sample (by total fastq filesize) ran into an OOM error in `EXTRACT_VIRAL_READS_ONT:MASK_FASTQ_READS`. This PR fixes that error by allocating memory based on the input file size.

I done a full end-to-end RUN workflow with the largest input files and some large-but-smaller ones. I've confirmed that:
1. The closure works so that the process is allocated the expected amount of memory
2. The scaling is sufficient all along the curve of input sizes, with plenty of headroom.

Note that the smallest value in the memory scaling is the current default, so we never allocate less memory than currently do.

---
[Claude from here]

Closes #736.

**Changes:**
- `configs/resources.config`: add a `ResourceTierUtils` class at the top of the file and a new `withLabel: bbmask_resources` block with `cpus = 16` and a dynamic `memory = { ResourceTierUtils.pickMemoryTier(reads, ['2 GB', '10 GB'], ['32 GB', '64 GB', '128 GB']) }`. Tier thresholds are gzipped-FASTQ byte sizes; the three memory values come from the empirical data below. The helper is generic (`pickMemoryTier(input, thresholds, memories)`) — accepts a `Path` or `List<Path>` (sums sizes across lists), takes `MemoryUnit`-parseable strings, and enforces the invariant `memories.size() == thresholds.size() + 1`.
- `modules/local/maskRead/main.nf`: change `label "large"` to `label "bbmask_resources"`. Module is label-only.
- `tests/nextflow.config`: add `withLabel: bbmask_resources` cap (4 CPUs, 15 GB) so the maskRead nf-tests fit GitHub runner limits. Structurally identical to the `small`, `large`, `kraken_resources`, and `blast_resources` caps in the same file.
- `docs/developer.md`: document closure-based `memory` directives alongside static resource labels as two variants of the same "labels own resources" pattern, including the caveat that any helper classes such closures depend on must live in `configs/resources.config`, not `lib/*.groovy`.

**Why the helper lives in the config rather than `lib/`**

Nextflow's `lib/*.groovy` auto-loading adds classes to the *script* classpath only — `main.nf`, included modules, `exec:`/`script:` blocks. It does not add them to the *config* classpath. A `memory = { ResourceTierUtils.pickMemoryTier(...) }` closure inside `configs/resources.config` fails at task-launch with `No signature of method: groovy.util.ConfigObject.pickMemoryTier(...)` because the config scope resolves `ResourceTierUtils` as an empty `ConfigObject` rather than the Groovy class. Declaring the class at the top of `configs/resources.config` puts it in the config classpath, where the `withLabel: bbmask_resources` closure can reach it.

The tier-picking logic therefore isn't covered by a `nextflow_function` test — nf-test targets classes on the script classpath, and ours is declared inside a config file. Tier behaviour is covered by manual reproducible runs against sparse fixtures (see below).

**Manual verification — tier selection at each boundary**

Reproducible on any machine with Nextflow installed:

```
truncate -s 1M /tmp/tiny.fastq.gz     # 1 MiB, expects 32 GiB
truncate -s 5G /tmp/medium.fastq.gz   # 5 GiB, expects 64 GiB
truncate -s 20G /tmp/big.fastq.gz     # 20 GiB, expects 128 GiB
```

Write a minimal driver (not committed) that attaches the real `bbmask_resources` label to a one-line `echo "memory=${task.memory}"` process, run it against each fixture with `-c configs/resources.config`, and read the resolved memory from `.command.sh` or stdout. On a machine smaller than 128 GiB the local executor rejects the task on CPU/memory before execution, but Nextflow still logs the requested memory — which is what the closure resolved to. With the committed config, the three fixtures resolve to 32 / 64 / 128 GiB respectively.

**Evidence — MASK_FASTQ_READS memory scaling**

18 empirical MASK_FASTQ_READS data points across three regimes: prod un-merged deliveries (existing baselines), synth-1 (8 merged libraries, 2–29 GB pre-filtlong), and synth-2 (5 merged libraries, 8–23 GB pre-filtlong, targeted at the 128 GiB tier and the 64/128 GiB boundary). Ordered below by the size actually seen by MASK (post-`FILTLONG`), which is what the closure evaluates. All sizes in GiB (binary, consistent with `MemoryUnit`).

image

| Source | Delivery | Scope | Pre-filtlong GiB¹ | MASK-input GiB² | Bucket | Peak RSS GiB | Peak/bucket | Runtime |
|:---:|---|---|---:|---:|:---:|---:|---:|---:|
| prod | 2026-02-13-b | unclassified pool | 1.70 | ≲ 1.7 ³ | 32 | 3.2 | ≤ 0.10 | 2m 21s |
| synth-1 | 2026-02-27-a | PZ-260210-Copl-NAS-P2-R1-A2 | 1.97 | 1.86 | 32 | 8.5 | 0.27 | 7m 11s |
| prod | 2026-04-03-a | unclassified pool | 2.18 | ≲ 2.2 ³ | 32–64 | 4.5 | ≤ 0.14 | 2m 42s |
| prod | 2026-01-28-b | unclassified pool | 2.24 | ≲ 2.2 ³ | 32–64 | 5.0 | ≤ 0.16 | 3m 18s |
| synth-1 | 2026-04-03-a | PZ-251208-BoDT-NAS-P2-A2 | 2.79 | 2.59 | 64 | 14.3 | 0.22 | 8m 24s |
| synth-1 | 2026-04-03-a | PZ-251208-BoDT-NAS-P2-A3 | 3.70 | 3.49 | 64 | 14.8 | 0.23 | 12m 6s |
| prod | 2026-04-10-a | unclassified pool | 3.93 | 1.20 ⁴ | 32 | 6.2 | 0.19 | 4m 14s |
| synth-1 | 2026-04-10-a | PZ-260331-SoS-NAS-P2-R1-A1 | 4.29 | 3.95 | 64 | 23.3 | 0.36 | 10m 9s |
| synth-1 | 2026-04-10-a | PZ-260330-NoS-NAS-P1-R1-A1 | 4.35 | 4.04 | 64 | 23.3 | 0.36 | 10m 35s |
| synth-1 | 2026-01-28-b | PZ-251208-BoDT-NAS-P2-A1 | 4.48 | 4.26 | 64 | 23.3 | 0.36 | 10m 31s |
| synth-1 | 2026-02-13-b | PZ-260206-MBTA_Ha-NAS-P2-R1-A1 | 5.33 | 5.02 | 64 | 27.8 | 0.43 | 13m 3s |
| prod | 2026-02-27-a | unclassified pool | 5.23 | ≲ 5.2 ³ | 64 | 8.7 | ≤ 0.14 | 9m 39s |
| synth-2 | 2026-03-04-a | PZ-260219-BoDT-NAS-P2-R1-A1 | 8.02 | 7.82 | 64 | 28.5 | 0.45 | 35m 42s |
| synth-2 | 2026-01-28-b | PZ-251210-BoDT-NAS-P2-R1-A1 | 9.84 | 9.62 | 64 | 34.3 | 0.54 | 51m 21s |
| synth-2 | 2026-01-13-a | PZ-260109-BoDT-NAS-P1-R1-A1 | 12.13 | 11.77 | 128 | 40.1 | 0.31 | 51m 24s |
| synth-2 | 2026-03-04-a | PZ-260216-Copl-NAS-P2-R1-A1 | 13.13 | 12.84 | 128 | 44.6 | 0.35 | 59m 26s |
| synth-2 | 2026-02-13-a | PZ-260131-MBTA_Ha-NAS-P1-R1-A1 | 21.64 | 21.20 | 128 | 71.4 | 0.56 | 1h 37m |
| synth-1 | 2026-02-27-a | PZ-260210-Copl-NAS-P3-R1-A2 | 27.35 | 26.54 | 128 | 84.2 | 0.66 | 1h 58m |

¹ **Pre-filtlong**: merged fastq (synth) or upstream `unclassified.fastq.gz` (prod) — the input handed to `FILTLONG`.
² **MASK-input = post-`FILTLONG`**, the value the closure evaluates. Filtlong is aggressive on prod unclassified pools (30% retained, from the one work dir that survived garbage collection) and mild on merged sample libraries (95–97% retained). Unclassified pools carry short and low-quality reads that `filtlong --min_length 50 --max_length 15000 --min_mean_q 90` trims out; merged sample libraries preserve the long-read distribution and pass through nearly intact.
³ The filtlong work dir for 4 of 5 prod unclassified tasks has been garbage-collected, so post-filtlong is shown as an upper bound at the pre-filtlong size. Scaling by the 04-10-a ratio would move the 2.18 GiB and 2.24 GiB rows into the 32 GiB bucket.
⁴ Directly measured via surviving work dir: 3.93 GiB pre-`FILTLONG` → 1.20 GiB post.

Observations:

- Under the proposed closure, peak/bucket ratio is ≤ 0.66 across all 18 points. The single 0.66 is the 29 GB synth-1 library at the top of the 128 GiB bucket; everything else sits ≤ 0.56.
- **The 10 GiB / 64→128 GiB step is validated.** `synth-2/PZ-251210-BoDT-NAS-P2-R1-A1` at 9.62 GiB post-`FILTLONG` exercised the tightest point in the closure's design — an input just under 10 GiB landing in the 64 GiB bucket — and peaked at 34.3 GiB (ratio 0.54, 1.87× headroom). A neighboring point at 7.82 GiB peaked at 28.5 GiB (ratio 0.45).
- **128 GiB is safe for the biggest merged inputs observed.** `synth-2/PZ-260131-MBTA_Ha-NAS-P1-R1-A1` at 21.20 GiB post-`FILTLONG` peaked at 71.4 GiB (ratio 0.56). The prior 84.2 GiB peak on synth-1's 26.54 GiB input (ratio 0.66) remains the largest observation.
- Peak_rss scales sub-linearly with MASK-input above ~6 GiB: 11.77 → 21.20 GiB input (+80%) gave 40.1 → 71.4 GiB peak (+78%); 21.20 → 26.54 GiB input (+25%) gave 71.4 → 84.2 GiB peak (+18%). Consistent with bbmask's working set saturating once input exceeds its hashing dictionary capacity. 128 GiB should cover merged libraries up to ~35–40 GiB before getting tight.
- Merged sample libraries have ~3× the peak/input ratio of prod unclassified pools at the same input size: unclassified pools are sparse per-byte (short, low-quality reads filtered heavily by filtlong); merged sample libraries preserve the long-read distribution, giving more bases per compressed byte and a bigger bbmask working set. The closure has to fit the merged case because that's the upcoming regime as the basecall workflow starts shipping merged outputs.

If a real-world production library ever breaches headroom in any bucket, the tier thresholds/memories can be widened in the `bbmask_resources` label in `configs/resources.config` without touching the module.


dp-rice (commented):

## Correctness

**No issues found.** The memory closure in `modules/local/maskRead/main.nf:9` evaluates `reads.size()` on a single staged `Path` (the process input is a single `path(reads)` and the only caller in `subworkflows/local/extractViralReadsONT/main.nf:51` feeds it a single file emitted by FILTLONG), so `.size()` returns file bytes as intended. The `withName: 'MASK_FASTQ_READS'` cap in `tests/nextflow.config` correctly overrides the module-level closure for CI.

## Testing

- **T1** (**suggestion**): The PR description acknowledges there is no automated coverage of the closure branches, and explains why the prior stub test was worse than no test. Consider a lightweight module-level nf-test that stages sparse fixtures at representative sizes and asserts against the real `process.memory` directive (not a copy of the closure logic) — this would catch regressions like an accidental threshold flip or unit swap without re-duplicating the closure. If `process.memory` turns out to be hard to introspect cleanly in nf-test, a one-line note on that limitation next to the closure (or in the new `docs/developer.md` exception entry) would save future maintainers from repeating the same stub-test experiment.

## Documentation

- **D1** (**suggestion**): The PR description frames the FILTLONG re-tier in terms of `EXTRACT_VIRAL_READS_ONT`, but the label change applies to every caller — `subworkflows/local/subsetTrim/main.nf` also invokes FILTLONG twice (`FILTLONG_STRINGENT`, `FILTLONG_LOOSE`) for ONT taxonomic subset-trimming. Worth one line in the PR body acknowledging the broader blast radius. The subset case runs on `n_reads_profile` reads so the risk is almost certainly fine, but a reviewer with ops context shouldn't have to discover the extra callers themselves.

## Quality

**No issues found.** The inline comment on `modules/local/maskRead/main.nf:2-5` explains the scaling model and clarifies that input sizes refer to the gzipped file, and the `tests/nextflow.config` comment block gives future maintainers the context for why the `withName` cap exists. Moving `cpus 16` into the process body alongside the `memory` closure is consistent with the new exception the PR is opening in `docs/developer.md`.

## Risks

- **R1** (**suggestion**): The FILTLONG CPU reduction (8 → 1) may shift Batch bin-packing on ONT run jobs beyond the viral-extraction path (see **D1**). Not a correctness risk, but flagging it in the PR body would let a human reviewer sign off on the wider impact with organizational context.
- **R2** (**nit**): The `memory` closure assumes `reads` is always a single `Path`. If a future change to the `MASK_FASTQ_READS` input interface ever admits multiple files (e.g., a list channel bound to `path(reads)`), `reads.size()` silently returns the list length instead of file bytes and always falls through to the 32 GB branch. A one-line comment documenting the single-file invariant — or defensively summing `reads*.size()` — would make the trap explicit.

dp-rice:
- D1/R1: Decided to drop this change altogether and focus the PR on only the actual problem process.
- R2: Added a comment.

dp-rice:
T1: addressed by extracting closure to a pure function

docs/developer.md
```diff
@@ -28,6 +28,7 @@ These guidelines represent best practices to implement in new code, though some
     - Each workflow should have a `<workflow_name>.md` document in `docs/`.
 - Process conventions (see `modules/local/vsearch/main.nf` for an example of a well-written process that follows these conventions):
     - All processes should have a label specifying needed resources (e.g. `label "small"`). Resources are then specified in `configs/resources.config`.
+        - Exception: when peak memory scales strongly with input size such that a single label would either under-provision the largest inputs (causing OOMs) or over-provision the smallest (wasting Batch capacity), a process may declare an input-size-aware `memory` closure directly in the module. Keep the tier-selection logic in a `lib/*.groovy` helper so it can be unit-tested via `nextflow_function` (see `lib/ResourceTierUtils.groovy` and `tests/lib/resourceTierUtils/main.nf.test`); the module's `memory` closure should be a one-line call into the helper. See `modules/local/maskRead/main.nf` for an example.
```
willbradshaw: For consistency and configurability, I would prefer to define the closure within a label definition in `resources.config` rather than directly in the process.

modules/local/maskRead/main.nf
```diff
@@ -1,7 +1,16 @@
...
+// lib/ResourceTierUtils.groovy) keeps headroom sane across the range of
+// ONT merged-library inputs while avoiding large reservations for small
+// inputs. The closure assumes `reads` is a single Path; if the input is
+// ever refactored to accept multiple files, `reads.size()` silently returns
+// the list length rather than file bytes, and every input falls through to
+// the smallest (32 GB) tier.
 process MASK_FASTQ_READS {
-    label "large"
     label "BBTools"
+    cpus 16
```
willbradshaw: As above, not sold on hardcoding this in the process, versus specifying in a label definition

tests/lib/resourceTierUtils/main.nf.test
```diff
@@ -0,0 +1,72 @@
+nextflow_function {
```
willbradshaw: Oh, this is neat, I didn't know about this.

lib/ResourceTierUtils.groovy
```diff
@@ -0,0 +1,16 @@
...
+// Files in lib/ are automatically loaded by Nextflow and callable from
+// `memory` closures and `exec:` blocks.
+
+import nextflow.util.MemoryUnit
+
+class ResourceTierUtils {
+
+    // Pick a memory tier for MASK_FASTQ_READS from the gzipped FASTQ size in bytes.
+    // Empirical basis and threshold choice: see PR #737.
+    static MemoryUnit maskFastqReadsMemory(long gzippedFastqBytes) {
```
willbradshaw: Following on on my suggestion above to move the resource definition into a config label definition, I suggest making this configurable with arguments; currently it's a bit too single-use for a global lib definition.

tests/nextflow.config
```diff
@@ -46,4 +46,13 @@ process {
         cpus = 4
         memory = 15.GB
     }
+
+    // MASK_FASTQ_READS dropped its `label "large"` in favor of an input-size-aware
```
willbradshaw: To me this is another argument for avoiding process-specific resource allocations

willbradshaw (changes requested):
I think the closure approach is good (much better than bumping a fixed allocation) but the current approach of putting the closure in the process definition has some bad side effects. Propose moving it into `resources.config` as a reusable resource definition

dp-rice:
Unfortunately, you can't import groovy objects from `lib/` in a config file. Claude:

>   The closure in `configs/resources.config` calls `ResourceTierUtils.pickMemoryTier(...)`, but Nextflow resolves `ResourceTierUtils` as a `groovy.util.ConfigObject`, not as the class in `lib/ResourceTierUtils.groovy`. Nextflow's `lib/*.groovy` auto-loading adds classes to the script classpath but not the config classpath, so the closure can't find the helper.

I think the options are:
1. Go back to the original setup, where we import the function into the process definition. Keep the test coverage, but break the label pattern.
2. Move the function to the config file. Lose the test coverage, but preserve the label pattern.

Up to you which is more important. I'm leaning toward 2 since we've already tested the function (doesn't help against regressions though).

willbradshaw:
I agree with (2).

docs/developer.md
```diff
@@ -28,6 +28,8 @@ These guidelines represent best practices to implement in new code, though some
     - Each workflow should have a `<workflow_name>.md` document in `docs/`.
 - Process conventions (see `modules/local/vsearch/main.nf` for an example of a well-written process that follows these conventions):
     - All processes should have a label specifying needed resources (e.g. `label "small"`). Resources are then specified in `configs/resources.config`.
+        - Most labels declare static resources (`cpus = 8; memory = 16.GB`).
+        - When a process's peak memory scales strongly with input size, the label's `memory` directive may be a closure over the process inputs — e.g. `memory = { ResourceTierUtils.pickMemoryTier(reads, ...) }`. Any helper classes such a closure depends on must be declared at the top of `configs/resources.config`, not in `lib/*.groovy`: Nextflow's `lib/` auto-loading adds classes to the script classpath but not the config classpath, so a config-scope closure cannot resolve `lib/` classes. See `bbmask_resources` and `ResourceTierUtils` in `configs/resources.config` for an example.
```
willbradshaw: Could you remove stale references to `lib`? I notice Claude often leaves these correction comments that make sense at the time but get stale quickly once the corrected-away-from state no longer exists.

willbradshaw (approved):

docs/developer.md
dp-rice: Haha, yes, I've been fighting with Claude to stop doing this all over the place and this is somehow where it landed. Sorry I missed that one.

PR #738, opened 2026-04-28
Tolerate empty viral taxa in DOWNLOAD_VIRAL_GENOMES

jo-delia-faraguna:
Make `DOWNLOAD_VIRAL_GENOMES` robust to taxa rooted at Viruses (10239) without assemblies, emitting a header-only `${taxid}_metadata.tsv` and exiting 0. This requires declaring `${taxid}_genomes/` as an optional output.

Chained off #740 

Closes [COMP-1896](https://linear.app/securebio-nao/issue/COMP-1896/cd-index-fails-when-a-newly-published-viral-taxon-has-no-assemblies)

## Context
- In [makeVirusGenomeDB](https://github.com/securebio/nao-mgs-workflow/blob/main/subworkflows/local/makeVirusGenomeDB/main.nf), we enumerate all child taxa of the Viruses taxon (10239) and download all genomes for each in parallel. If any one of these child taxa has no assemblies, the process errors out.
- NCBI publishes new viral taxa to its taxonomy before assemblies are linked to them, causing `datasets download genome taxon <taxid>` to exit 1 with `Error: ... no genome data is currently available for this taxon.`
- Adding these taxa to [`viral_taxids_exclude_hard`](https://github.com/securebio/nao-mgs-workflow/blob/main/configs/index.config#L40) would not work since this parameter has no effect on which taxa get downloaded and is only passed to `MAKE_VIRUS_TAXONOMY_DB`.

## Reproducing example
- `datasets download genome taxon 3700704` fails with `Error: The taxonomy ID '3700704' is valid for 'Yamazakiviridae', but no genome data is currently available for this taxon.`
  - This taxon is rooted at Viruses.
  - As of 2026-04-29, this taxon and its children have no assemblies (but there are some GenBank nucleotide records like [PQ111734.1](https://www.ncbi.nlm.nih.gov/nuccore/PQ111734.1)).

## Detailed changes
- Added a new test for the empty-taxon path using a stub `datasets` that exits 1 with a hardcoded NCBI error string.
  - Spoofing the error message risks the test passing even if the `datasets` CLI message drifts, but it has been stable across all releases and a public nf-core repo [here](https://github.com/nf-core/pathogensurveillance/blob/master/modules/local/find_assemblies/main.nf#L28) greps on the same string.
  - If the `datasets` path changes in the container we're using and the stub no longer overrides it, the `empty_taxon`-tagged test should fail, catching any drift there.

### DIFF
--- a/tests/configs/empty_taxon.config
+++ b/tests/configs/empty_taxon.config
@@ -1,13 +1,13 @@
 /*
 ========================================================================================
     Test config for the empty-taxon success path of DOWNLOAD_VIRAL_GENOMES.
-    
+
     Mounts a stubbed `datasets` CLI into the container based on the `ncbi_datasets` image
     in `configs/containers.config`, replacing the real binary at /opt/conda/bin/datasets.
-    
+
     The stub exits non-zero with the same error that NCBI returns for a valid taxon with
     no linked assemblies.
-    
+
     If a future image changes the binary path, the `empty_taxon` test would fail because
     the stub would not replace the real `datasets` binary, which would fail with an error
     like `Error: The taxonomy name 'taxid-with-no-assemblies' is not recognized.`
--- a/tests/configs/empty_taxon.config
+++ b/tests/configs/empty_taxon.config
@@ -1,0 +1,22 @@
+/*
+========================================================================================
+    Test config for the empty-taxon path of DOWNLOAD_VIRAL_GENOMES.
+    Bind-mounts a stubbed `datasets` CLI into the container and prepends it
+    to PATH. The stub exits 1 with the same stderr signature NCBI emits
+    when a taxon has no assemblies linked, exercising the module's
+    "emit empty outputs" branch without depending on which taxa happen to
+    be empty in the live NCBI taxonomy.
+========================================================================================
+*/
+
+includeConfig "index.config"
+
+process {
+    withName: DOWNLOAD_VIRAL_GENOMES {
+        // Bind-mount the stub directly over the real `datasets` binary inside
+        // the container. We can't use beforeScript + PATH because beforeScript
+        // runs on the host (in the docker-launching wrapper), not inside the
+        // container.
+        containerOptions = "-v ${projectDir}/tests/modules/local/downloadViralGenomes/stub_bin/datasets:/opt/conda/bin/datasets:ro"
+    }
+}
--- a/tests/configs/empty_taxon.config
+++ b/tests/configs/empty_taxon.config
@@ -1,10 +1,9 @@
 /*
 ========================================================================================
-    Test config for failure paths of DOWNLOAD_VIRAL_GENOMES.
+    Test config for the empty-taxon success path of DOWNLOAD_VIRAL_GENOMES.
     Mounts a stubbed `datasets` CLI into the container, replacing the real
-    binary at /opt/conda/bin/datasets. The stub exists non-zero to simulate
-    a true process error as well as the handled special case where no assemblies
-    exist for a given taxon.
+    binary at /opt/conda/bin/datasets. The stub exits non-zero the same error
+    that NCBI returns for a valid taxon with no linked assemblies.
 ========================================================================================
 */
 
@@ -13,9 +12,9 @@ includeConfig "index.config"
 process {
     withName: DOWNLOAD_VIRAL_GENOMES {
         // Replace the real `datasets` binary in the container with the stub,
-        // based on the binary path in the `ncbi_datasets` image pinned in 
+        // based on the binary path in the `ncbi_datasets` image pinned in
         // `configs/containers.config`.
-        // If a future image changes the binary path, both `empty_taxon` tests
+        // If a future image changes the binary path, the `empty_taxon` test
         // would fail because the `datasets` command would fail with an error like
         // `Error: The taxonomy ID '99999999' does not match any existing taxids for 'genome'`
         containerOptions = "-v ${projectDir}/tests/modules/local/downloadViralGenomes/stub_bin/datasets:/opt/conda/bin/datasets:ro"
--- a/tests/configs/empty_taxon.config
+++ b/tests/configs/empty_taxon.config
@@ -14,11 +14,10 @@ process {
     withName: DOWNLOAD_VIRAL_GENOMES {
         // Replace the real `datasets` binary in the container with the stub,
         // based on the binary path in the `ncbi_datasets` image pinned in 
-        // `configs/containers.config`. If a future image changes the binary path,
-        // the real` datasets` binary will instead be used. However, this
-        // regression would cause the `Should emit empty outputs when taxon has
-        // no assemblies` test to fail with an error like `Error: The taxonomy ID
-        // '99999999' does not match any existing taxids for 'genome`
+        // `configs/containers.config`.
+        // If a future image changes the binary path, both `empty_taxon` tests
+        // would fail because the `datasets` command would fail with an error like
+        // `Error: The taxonomy ID '99999999' does not match any existing taxids for 'genome'`
         containerOptions = "-v ${projectDir}/tests/modules/local/downloadViralGenomes/stub_bin/datasets:/opt/conda/bin/datasets:ro"
     }
 }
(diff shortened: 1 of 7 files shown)


jo-delia-faraguna (commented):

## Correctness

**No issues found.** The sentinel grep at `modules/local/downloadViralGenomes/main.nf:31` is anchored to `^Error:.*no genome data is currently available for this taxon\.$`, tightly matching NCBI's emitted line and avoiding spurious matches elsewhere in stderr. The `if ! datasets ...; then ...; fi` control flow correctly separates the swallow-empty-taxon and propagate-failure branches, and `optional: true` on the `genomes` output is the standard Nextflow idiom for the empty case.

## Testing

**No issues found.** Both the empty-taxon swallow path and the negative (non-sentinel error propagation) path are exercised by stubbed `datasets` invocations in `tests/modules/local/downloadViralGenomes/main.nf.test:41-94`, and the stub's default `*)` arm at `tests/modules/local/downloadViralGenomes/stub_bin/datasets:34-39` exits non-zero with an explicit error so unmapped test taxids surface as failures rather than silently exercising the empty-taxon branch.

## Documentation

**No issues found.** The new behavior is captured in the inline comments at `modules/local/downloadViralGenomes/main.nf:14-21`, the CHANGELOG entry, and the PR description. The bind-mount coupling is documented in `tests/configs/empty_taxon.config:16-24`, and the taxid-to-branch contract between the test and stub is documented at `tests/modules/local/downloadViralGenomes/main.nf.test:34-40` and `tests/modules/local/downloadViralGenomes/stub_bin/datasets:2-11`. No user-facing references to `DOWNLOAD_VIRAL_GENOMES` failure semantics exist in `docs/` or `README.md` that would need updating.

## Quality

**No issues found.** The header schema is extracted to a `metadata_header` Groovy variable at `modules/local/downloadViralGenomes/main.nf:16` and reused in both the empty-taxon branch (line 33) and the happy-path header rewrite (line 67). The `if ! ...; then ...; fi` control flow is idiomatic and avoids manual `set +e`/`rc=$?` bookkeeping.

## Risks

**No issues found.** The PR description explains the motivation (NCBI taxonomy lag), provides a concrete reproducer (taxid 3700704), names the design choice (sentinel-string detection), and acknowledges its brittleness with an external precedent (nf-core/pathogensurveillance grepping the same string). Downstream impact is bounded: `optional: true` on the `genomes` output is standard Nextflow, and `PREPARE_VIRAL_METADATA` already tolerates header-only metadata input. No human-judgment risks are missing from the description.

_Reviewed after 4 iteration(s) of automated review-and-fix. 9 findings were automatically resolved._

tests/modules/local/downloadViralGenomes/main.nf.test
```diff
@@ -31,4 +31,66 @@ nextflow_process {
...
+                """
+                input[0] = "88888888"
+                input[1] = "refseq"
+                input[2] = ""
+                input[3] = 5
+                """
+            }
+        }
+        then {
+            assert process.failed
```
willbradshaw: You need to also check that the error message is as expected, otherwise the test could pass if it fails for some unrelated reason like a syntax error.

tests/configs/empty_taxon.config
willbradshaw, line 1: This is very opaque to me; it reads like uncurated AI verbiage. It's great that you've found a way to test this but work is needed to make this more comprehensible.

willbradshaw (changes requested):
I'm on board with the approach but I don't understand the dark magic you're doing for the tests; clarification needed. (Also, the PR description needs a human touch.)

tests/modules/local/downloadViralGenomes/main.nf.test
jo-delia-faraguna: Thanks - added a check for the error message.

tests/configs/empty_taxon.config
jo-delia-faraguna: Thanks. I spent some time reading over and tweaking the comments before sending to you, but I agree this was not comprehensible. I rewrote to make it more concise and hopefully more clear.

jo-delia-faraguna (commented):

## Correctness

**No issues found.** The empty-taxon path captures stderr to `dl_err.txt`, replays it to the parent stderr for visibility, dispatches on an anchored regex matching the documented NCBI error, and emits a header-only metadata TSV with no genome files (consistent with the now-`optional` `genomes` output). Non-empty failures still propagate via the explicit `exit 1`, and the `trap` cleans up the temp file on all exits.

## Testing

- **T1** (**suggestion**): `tests/modules/local/downloadViralGenomes/main.nf.test:78` only asserts `process.out.genomes.size() == 0`. Because `genomes` is `optional: true`, this passes both when the optional output is correctly absent and (hypothetically) when the glob silently matches nothing in a future regression. Consider also asserting that no `*.fna.gz` files exist in the work directory, or that the success-path stderr contains the "emitting empty outputs" diagnostic, to pin the test to the empty-taxon code path specifically rather than just to "no genomes were emitted."
- **T2** (**suggestion**): There is no integration test confirming that `MAKE_VIRUS_GENOME_DB` (`subworkflows/local/makeVirusGenomeDB/main.nf:39-48`) handles a header-only metadata TSV from `CONCATENATE_TSVS` and an empty `download_ch.genomes.collect()` from one child taxon without breaking downstream consumers (`PREPARE_VIRAL_METADATA`, `FILTER_VIRAL_GENBANK_METADATA`, etc.). The module-level test verifies the new branch in isolation, but the bug being fixed is a subworkflow-level failure; a subworkflow test (or a note in the PR description about manual verification at that level) would close the loop on the actual failure mode.

## Documentation

- **D1** (**nit**): The new `tests/configs/empty_taxon.config` and the `stub_bin/datasets` pattern (mounting a stubbed binary into the container via `containerOptions`) is novel in this repo — no prior `stub_bin/` or `containerOptions` mounts exist under `tests/`. Consider a brief mention in `docs/testing.md` so future contributors know the convention exists and where to place similar stubs; otherwise the next person needing to stub a containerized CLI is likely to reinvent it.

## Quality

- **Q1** (**nit**): `modules/local/downloadViralGenomes/main.nf:14` defines `metadata_header` as a Groovy `def` and interpolates it into both the empty-taxon branch and the normal-path header rewrite (line 74). This is a clean dedup, but note that the four column names are also asserted in the test at line 83. If a future change adds a column, all locations must move together — a brief comment on the `def` line pointing to the test would make the coupling explicit.
- **Q2** (**nit**): `tests/modules/local/downloadViralGenomes/stub_bin/datasets:11-18` parses argv with a manual `prev`/`for` loop to find the value after `taxon`. A simpler `while [[ $# -gt 0 ]]; do case "$1" in taxon) taxid="$2"; shift 2;; *) shift;; esac; done` would be more idiomatic, but the current implementation works correctly given the fixed argument order produced by the module.

## Risks

**No issues found.** The PR description explicitly calls out the trade-off of grepping on the NCBI error string (with a comment in `main.nf` and a public nf-core precedent), the deferred-stderr behavior is documented inline at `modules/local/downloadViralGenomes/main.nf:22-29`, and the stub-binary mounting strategy and its failure mode (image path drift) are documented in `tests/configs/empty_taxon.config`. The behavioral change (empty viral taxa no longer fail the index build) is summarized in the CHANGELOG entry with sufficient context for human reviewers.

_Reviewed after 4 iteration(s) of automated review-and-fix. 10 findings were automatically resolved._

jo-delia-faraguna:
- T1: **Addressed**: added an additional `aasert`
- T2: **Declined**: I thought about implementing this, but it would require mocking `datasets download, datasets rehydrate, dataformat tsv genome` and `CONCATENATE_TSVS` is already tested to handle empty inputs
- D1: **Addressed**: added a brief note to `docs/testing.md`
- Q1: **Declined**: The tests will fail if this changes, so it is already self-documenting
- Q2: **Addressed**: Switched to use the proposed expression

jo-delia-faraguna:
@willbradshaw back to you. I reworked the comments and documentation by hand. I also added a test that verifies unrelated errors are raised using the real, non-stubbed `datasets` binary. I wrote the PR description by hand but I reworked it slightly based on your feedback.

willbradshaw (approved):
Much better, thanks!

PR #743, opened 2026-04-29
Add --batch_job_role parameter to Batch profiles

### LINKED ISSUE #742: Add support for aws.batch.jobRole in Batch profiles
Add a new optional `--batch_job_role` pipeline parameter that, when supplied, configures `aws.batch.jobRole` on the `standard`, `batch`, and `test_run` profiles and disables `fusion.exportStorageCredentials`. When omitted, profile behavior is unchanged.

This lets operators run the pipeline using an IAM role attached to the Batch job rather than relying on the caller's exported AWS credentials, which is the standard pattern for AWS Batch workloads.

**Scope:**
- `configs/profiles.config`: new `batch_job_role` param; conditional `fusion.exportStorageCredentials` and `aws.batch.jobRole` in the three Batch profiles above.
- `docs/batch.md`: document the new parameter.

The `ec2_s3` profile is intentionally unchanged — it does not use the `awsbatch` executor, so `aws.batch.jobRole` does not apply. The `ec2_local` and `rust_dev` profiles do not export credentials.

Default behavior (no `--batch_job_role`) is fully preserved; no breaking change.

dp-rice:
Adds a new optional `--batch_job_role` pipeline parameter so operators can run mgs-workflow against AWS Batch with credentials sourced from an IAM role attached to the job, rather than from the caller's exported AWS credentials. When the parameter is omitted, profile behavior is fully unchanged — no breaking change for existing users.

Closes #742.

I've confirmed that this actually attaches the IAM role to the jobs.

**Changes:**
- `configs/profiles.config`: declared a new `batch_job_role = ""` param; in the `standard`, `batch`, and `test_run` profiles, replaced the hard-coded `fusion.exportStorageCredentials = true` with `fusion.exportStorageCredentials = !params.batch_job_role` and added `aws.batch.jobRole = params.batch_job_role ?: null`. The `ec2_s3` profile is intentionally untouched because it does not use the `awsbatch` executor.
- `docs/batch.md`: added a new "Optional: use a Batch job role" section documenting when to use the parameter and the IAM trust/PassRole prerequisites.
- `CHANGELOG.md`: added an entry under the existing `v3.2.1.4-dev` heading. No version bump (point-level work continues in the open dev cycle); no change to index/pipeline compatibility trackers.

**Validation:**

Verified end-to-end with `nextflow run` (which exercises real CLI param processing) against a no-op workflow that prints the resolved values of `params.batch_job_role`, `fusion.exportStorageCredentials`, and `aws.batch.jobRole`. Each of the three modified profiles was tested with and without `--batch_job_role`:

| Profile | Flag | `fusion.exportStorageCredentials` | `aws.batch.jobRole` |
|---|---|---|---|
| `standard` | (omitted) | `true` | `null` |
| `standard` | `--batch_job_role <ARN>` | `false` | `<ARN>` |
| `batch` | (omitted) | `true` | `null` |
| `batch` | `--batch_job_role <ARN>` | `false` | `<ARN>` |
| `test_run` | (omitted) | `true` | `null` |
| `test_run` | `--batch_job_role <ARN>` | `false` | `<ARN>` |

The three "no flag" rows match today's behavior exactly, confirming backwards compatibility. The three "flag set" rows confirm the conditional flips both settings together when the parameter is provided.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # v3.2.1.4-dev
 
+- Add `--batch_job_role` parameter to allow Batch profiles to use an IAM job role instead of exported AWS credentials (#742).
 - Prevent `MASK_FASTQ_READS` from running out of memory on merged ONT libraries larger than ~6 GB of gzipped FASTQ (#737).
     - Replaces `label "large"` with a new `label "bbmask_resources"` whose `memory` directive is an input-size-aware closure (32 / 64 / 128 GiB tiers, chosen from input byte size). The label and its closure live in `configs/resources.config` alongside other resource labels; the module itself stays label-only. CPU allocation unchanged at 16.
     - Tier-selection logic is a generic `ResourceTierUtils.pickMemoryTier(input, thresholds, memories)` helper declared at the top of `configs/resources.config`. The class lives in the config file (rather than `lib/*.groovy`) because Nextflow's `lib/` auto-loading puts classes on the script classpath but not the config classpath; config-scope `memory = { … }` closures can't resolve `lib/` classes.
--- a/configs/profiles.config
+++ b/configs/profiles.config
@@ -15,7 +15,7 @@ nextflow.enable.moduleBinaries = true
 
 params {
     db_download_timeout = 1200 // Timeout in seconds for database downloads (default: 20 minutes)
-    batch_job_role = ""        // Optional IAM role ARN for Batch jobs; when set, fusion credential export is disabled and the role is attached via aws.batch.jobRole
+    batch_job_role = ""        // Optional IAM role ARN for Batch jobs (see docs/batch.md)
 }
 
 // Workflow run profiles
--- a/configs/profiles.config
+++ b/configs/profiles.config
@@ -15,13 +15,15 @@ nextflow.enable.moduleBinaries = true
 
 params {
     db_download_timeout = 1200 // Timeout in seconds for database downloads (default: 20 minutes)
+    batch_job_role = ""        // Optional IAM role ARN for Batch jobs; when set, fusion credential export is disabled and the role is attached via aws.batch.jobRole
 }
 
 // Workflow run profiles
 profiles {
     standard { // Run on AWS Batch
         fusion.enabled = true
-        fusion.exportStorageCredentials = true
+        fusion.exportStorageCredentials = !params.batch_job_role
+        aws.batch.jobRole = params.batch_job_role ?: null
         process.executor = "awsbatch"
         process.errorStrategy = "retry"
         process.maxRetries = 1
@@ -29,7 +31,8 @@ profiles {
     }
     batch { // Run on AWS Batch
         fusion.enabled = true
-        fusion.exportStorageCredentials = true
+        fusion.exportStorageCredentials = !params.batch_job_role
+        aws.batch.jobRole = params.batch_job_role ?: null
         process.executor = "awsbatch"
         process.errorStrategy = "retry"
         process.maxRetries = 1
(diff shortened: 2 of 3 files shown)


dp-rice (commented):

## Correctness

**No issues found.** The Groovy truthiness logic is correct for the documented contract: with the default empty string, `!params.batch_job_role` evaluates to `true` (preserving today's `fusion.exportStorageCredentials = true` behavior) and `params.batch_job_role ?: null` resolves to `null` (equivalent to leaving `aws.batch.jobRole` unset). The PR's six-row validation matrix exercises both states across all three modified profiles.

## Testing

- **T1** (**suggestion**): The validation matrix in the PR description confirms config *resolution* (param → `fusion.exportStorageCredentials` / `aws.batch.jobRole`), but does not confirm end-to-end Batch execution with an actual job role assuming credentials inside a container. That's the failure mode users will hit if the wiring is wrong (e.g., a Nextflow version that handles `aws.batch.jobRole = null` differently than expected, or a Fusion edge case). Worth either explicitly noting in the PR description that end-to-end Batch execution was not validated, or running one real Batch job through the `batch` profile with `--batch_job_role` to confirm the role is actually attached and credentials flow through.

## Documentation

- **D1** (**suggestion**): `docs/batch.md:115` lists the three modified profiles by name (`standard`, `batch`, `test_run`) but doesn't mention that `ec2_s3` is intentionally not covered. A reader scanning for a job-role option who is using `ec2_s3` may waste time before realizing it doesn't apply. Consider a single sentence noting that `ec2_s3` is excluded because it doesn't use the `awsbatch` executor.
- **D2** (**nit**): `configs/profiles.config:18` — the inline comment is on a long line (~140 chars). Many of the surrounding param comments are shorter; consider breaking it across two lines or trimming for visual consistency.

## Quality

- **Q1** (**nit**): The two-line block (`fusion.exportStorageCredentials = !params.batch_job_role` + `aws.batch.jobRole = params.batch_job_role ?: null`) is duplicated across three profiles. Nextflow profiles don't have clean inheritance, so duplication is hard to avoid here, but if a fourth Batch profile is added later, the duplication becomes a real correctness risk (drift between profiles). Not worth refactoring for this PR; flagging as a future maintenance concern.

## Risks

**No issues found.** The PR description already covers the relevant risks: the parameter defaults to a falsey value so existing profile behavior is preserved, no schema/dependency changes, and the IAM prerequisites (trust policy with `ecs-tasks.amazonaws.com` and caller `iam:PassRole`) are documented. The two CI failures (`check-nextflow-version`, `scan-containers`) are unrelated dev-branch issues, not caused by this PR.

dp-rice:
- D1: not worth addressing. This profile doesn't run batch.
- D2: Fixed
- Q1: nothing to do here.

dp-rice:
[Claude Code]

Re T1 — the pending CI checks don't address it. All the nf-test jobs (`test-modules`, `test-subworkflows`, `test-workflows-{run,index,downstream}`) run on Ubuntu runners with `tests/nextflow.config` and the local executor; none of them load `configs/profiles.config` or activate a Batch profile, so a regression in the modified `fusion.exportStorageCredentials` / `aws.batch.jobRole` resolution would pass green.

A pure-CI fix (e.g. a `nextflow config -profile <name>` matrix) would only re-check what the PR description already validates — config resolution — without actually running anything on Batch, so it isn't worth the wiring.

The cheapest meaningful end-to-end check is to run the existing tiny test dataset against the `test_run` profile twice: once with `--batch_job_role <ARN>` and once without. That single pair exercises everything that matters:

- Batch rejects submission if `aws.batch.jobRole` isn't attached or the caller lacks `iam:PassRole`.
- Fusion containers fail to read S3 inputs if `fusion.exportStorageCredentials = false` and the role isn't being assumed inside the container.
- The control run (no flag) confirms no regression in the default path.

Plan: run that pair once before merge using a role with S3 access to the test bucket and the `ecs-tasks.amazonaws.com` trust policy already documented in `docs/batch.md`, and note the outcome in the PR description. Not worth wiring into CI as a standing job.

dp-rice:
@willbradshaw I haven't tested this with a real job role yet. I'll set up the infra in the terraform repo separately and run a quick test (see previous comment) before merging. But I think it's fine to review now.

willbradshaw (approved):

PR #745, opened 2026-04-29
Bump minimum Nextflow version to 26.04.0

jo-delia-faraguna:
Bumps the pipeline's minimum required Nextflow version from 25.10.4 to 26.04.0.

Closes ([COMP-1910](https://linear.app/securebio-nao/issue/COMP-1910/bump-nextflow-version-to-26040))


jo-delia-faraguna:
I did some work on this but it seems like the latest nf-test bundles Groovy 4.0.30 [here](https://github.com/askimed/nf-test/blob/v0.9.5/pom.xml#L32) while nextflow bundles Groovy 4.0.31 [here](https://github.com/nextflow-io/nextflow/blob/v26.04.0/modules/nextflow/build.gradle#L44-L49), so this might be blocked on a new nf-test release.

PR #753, opened 2026-05-01
COMP-1177: Partition taxon subtree for viral genome downloads

jo-delia-faraguna:
The dehydrated `datasets download genome taxon` step in `DOWNLOAD_VIRAL_GENOMES` fails reliably for taxon 2559587 (Riboviria, ~198k assemblies) because the long-lived HTTP/2 stream NCBI uses to deliver the manifest zip becomes unstable at that size — every retry over the past day has hit the same `stream error: stream ID …; INTERNAL_ERROR; received from peer`. Splitting at depth 1 is insufficient: the taxonomy under Riboviria is highly skewed, with ~98% of assemblies pinned to a single child at every level down through Alphainfluenzavirus.

This PR replaces the depth-1 enumeration with a top-down threshold-based subtree partition. Starting at the requested root, the algorithm emits the shallowest set of subtree roots whose subtree sizes (taxa count in `nodes.dmp`, used as a conservative proxy for assembly count) do not exceed a configurable threshold. At the default of `50000`, the 10239 subtree partitions into ~180 shards with the largest holding ~19k actual assemblies — well inside the band proven safe by the largest known-good child of 10239 (2731341, ~41k assemblies).

**Changes:**
- New `modules/local/partitionTaxonSubtree/` (replaces `modules/local/enumerateChildTaxa/`): Python script and `PARTITION_TAXON_SUBTREE` process implementing the top-down partition with iterative DFS for subtree sizing.
- New `params.virus_taxid_max_partition_size` (default `50000`), threaded through `workflows/index.nf` -> `MAKE_VIRUS_GENOME_DB` -> the new module: tuning knob for shard size, calibrated against demonstrated-safe and demonstrated-failing taxa.
- `subworkflows/local/makeVirusGenomeDB/main.nf`: import + call updated; new positional argument for the threshold.
- Renamed nf-test directory and rewrote tests to cover root-fits, splits-at-children, and leaf-returns-self at the new module signature.
- Pytest covers `build_child_map`, `compute_subtree_sizes`, `partition_subtree`, and the partition entrypoint with parametrized threshold cases (14 tests).
- `docs/config.md`: documented the new parameter; clarified the existing `download_virus_taxid` description.
- Version bump `3.2.1.4` -> `3.2.1.5-dev` (Point: internal sharding refactor; output structure of `MAKE_VIRUS_GENOME_DB` is unchanged).


jo-delia-faraguna:
Closing since assemblies are not guaranteed to be on leaf nodes and this approach becomes a lot more complicated given that.

PR #757, opened 2026-05-02
Bump BBDuk resources

evanfields:
* Adds a "medium" resource label with 12 CPUs and bumps the interleaved BBDuk processes from small (8 CPU) to medium;
* Tweaks BBDuk command line invocation to reserve threads for data pipelines and JVM overhead, reducing starvation and thrashing

In recent benchmarking experiments I found that overall throughput of the zcat -> paste -> BBDuk pipeline was degraded when BBDuk had `t=<number of CPUs on the machine>`, likely due to mild thrashing: no full cores left for zcat, paste, BBDuk parsing, and JVM overhead. Reducing the cores allocated to BBDuk actually improved overall throughput somewhat.

Rather than just reduce the cores, I've also added a new medium resource level and used that for the BBDuk process. This should approximately max out the data throughput of two single threaded zcat processes. In simple benchmarks, this change increases throughput 1.5x, so a pure wall time win without additional cost.

Asterisks on the above: (A) most EC2 families don't have a 12-core size, so if Batch isn't nicely packing 12-core containers onto a larger instance, this could end up being a modest cost increase; (B) this is assuming streaming data via Fusion isn't already a bottleneck. I think (B) is true; I mimicked by testing the streaming speed from a mounted S3 bucket (same region) which was very fast.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# v3.2.1.5-dev
+
+- Reserve CPUs for streaming pipeline processes in BBDuk to prevent thread thrashing, and add a new `medium` resource label (12 CPU, 24 GB).
+
 # v3.2.1.4
 
 - Tolerate viral taxa with no linked NCBI assemblies in `DOWNLOAD_VIRAL_GENOMES`, emitting a header-only `${taxid}_metadata.tsv` and an empty `${taxid}_genomes/` (now declared `optional`).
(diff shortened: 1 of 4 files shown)


willbradshaw (approved):
Code changes seem fine, but some required tests are failing; will need to fix those before we can merge.

evanfields:
I don't think I'll have time to clean this up in the immediate future, and with the new Nucleaze PR this matters less. Probably better to spend the time on Nucleaze anyway!

PR #758, opened 2026-05-05
Drop non-current assemblies and duplicate viral accessions during makeVirusGenomeDB

jo-delia-faraguna:
Prevent duplicate viral sequences from ended up in indexes by dropping non-current assemblies from NCBI in `FILTER_VIRAL_GENBANK_METADATA` and de-duplicating accessions in `CONCATENATE_GENOME_FASTA`.

Closes [COMP-1943](https://linear.app/securebio-nao/issue/COMP-1943/update-make-virus-genome-db-to-drop-duplicate-sequences)

## Context

- NCBI `datasets` returns non-current assembly versions alongside current ones for some viral taxa (see [ncbi/datasets#576](https://github.com/ncbi/datasets/issues/576)), which can introduce duplicate sequence accessions
- `CONCATENATE_GENOME_FASTA` naively concatenates and preserves these duplicate sequence accessions
- `bowtie2-build` does not error in `INDEX` but `samtools view`does in `RUN:EXTRACT_VIRAL_READS:EXTRACT_VIRAL_READS_SHORT:BOWTIE2_VIRUS` with `[E::sam_hrecs_update_hashes] Duplicate entry`

## Changes

- `DOWNLOAD_VIRAL_GENOMES` writes `assembly_status` to per-taxid metadata TSV
- `FILTER_VIRAL_GENBANK_METADATA` drops rows where `assembly_status != "current"`
- `CONCATENATE_GENOME_FASTA` runs `seqkit rmdup --by-name` after concatenation to drop any lingering duplicates defensively

## Performance

- I ran the `seqkit rmdup --by-name` command against `s3://nao-testing/mgs-workflow-test/index-latest/output/results/virus-genomes-masked.fasta.gz` and it completed in ~1min, so I do not expect future production index builds to be significantly flakier or slower as a result of this change

## Reproducing example

The `RUN:EXTRACT_VIRAL_READS:EXTRACT_VIRAL_READS_SHORT:BOWTIE2_VIRUS`  error is reproducible using the latest SB internal index and the public `Illumina_100M` benchmark.

```bash
nextflow run main.nf \
  -c configs/run.config \
  -profile batch \                                              
  --platform illumina \                              
  --base_dir <base-dir> \
  --ref_dir <latest-SB-internal-index> \ 
  --sample_sheet s3://nao-testing/benchmarks/Illumina_100M/metadata/samplesheet.csv \                                         
  --queue <queue>
```


modules/local/concatenateGenomeFasta/main.nf
```diff
@@ -1,7 +1,7 @@
 // Concatenate downloaded genomes from ncbi-genome-download according to a file of genome IDs
 process CONCATENATE_GENOME_FASTA {
     label "single"
-    label "seqtk"
```
jo-delia-faraguna, line 4: We didn't actually use `seqtk` in the original implementation.

jo-delia-faraguna (commented):

## Correctness

**No issues found.** The schema guard in `filter-viral-genbank-metadata.py:57-61` raises before the `assembly_status == "current"` filter would silently match nothing, the `seqkit rmdup --by-name` invocation in `concatenateGenomeFasta/main.nf:27-29` runs under `set -euo pipefail` so a failure on either side of the pipe aborts the process, and the `[[ -s genomes-duplicates.tsv ]]` guard handles the no-duplicates case where seqkit may not create the file.

## Testing

- **T1** (**suggestion**): `test_filter` in `test_filter_viral_genbank_metadata.py:68` writes the script's `output_db` to a `.tsv.gz` path and then reads it via `gzip.open`, which exercises pandas's auto-detected gzip output but does not actually verify the deduplication-relevant behavior the rest of the PR introduces — the parametrized cases all hinge on `assembly_status` filtering and never feed two rows with the same `assembly_accession`. Consider adding a case where two rows share an accession and only one is `current`, asserting the surviving row is the `current` one, to lock in the de-duplication intent stated in the PR description.
- **T2** (**nit**): The `concatenateGenomeFasta` nf-test at `tests/modules/local/concatenateGenomeFasta/main.nf.test:40` asserts `fasta["SHARED_NAME_DIFF_SEQ"] == "AAAAAAAA"` to verify "first occurrence" semantics, but this depends on `xargs` preserving the order of `paths.csv`. Adding a brief comment that `paths.csv` lists `GCA_002` before `GCA_003` (and that this ordering is what makes `AAAAAAAA` the expected first occurrence) would make the assertion easier to follow for future readers.

## Documentation

- **D1** (**suggestion**): `docs/output.md:108` describes `virus-genome-metadata-gid.tsv.gz` but does not enumerate its columns, so the silent addition of `assembly_status` is not user-visible there. If users are expected to consume this file, consider noting the new column (or, if the columns are intentionally undocumented, leaving as-is). No action required if the file is treated as internal.

## Quality

- **Q1** (**nit**): The comment block at `modules/local/concatenateGenomeFasta/main.nf:22-26` is missing a trailing period on the last sentence ("...this guards against remaining duplicates") — minor consistency nit relative to the other comment sentences in the same block.

## Risks

**No issues found.** The PR description covers the behavioral change (dropping non-current assemblies, defensive de-duplication), the container swap from `seqtk` to the existing `seqkit` image, and a performance measurement against the production-scale concatenated FASTA. The schema addition (`assembly_status`) is internal to the index build and the one downstream consumer (`process_viral_bowtie2_sam.py:130-135`) reads only named columns, so the extra column is non-breaking.

_Reviewed after 4 iteration(s) of automated review-and-fix. 9 findings were automatically resolved._

jo-delia-faraguna:
[Claude Code] All four findings addressed in abdcd982:

- **T1 (test dedup intent):** Refactored `filter_viral_genbank_metadata.py` to expose `filter_metadata()` and rewrote the test to import directly. The existing parametrized cases now assert `(accession, taxid, assembly_status)` tuples (not just accession sets), which locks in that the surviving rows are the `current` ones with the right taxid.
- **T2 (paths.csv ordering):** Updated the comment at the assertion to note that `paths.csv` lists `GCA_002.fna.gz` before `GCA_003.fna.gz`, which is what makes `AAAAAAAA` the expected first occurrence.
- **D1 (docs/output.md column list):** Not changed. `docs/output.md` does not enumerate columns of any output TSV — it only describes the file's purpose. Listing only `assembly_status` would be inconsistent.
- **Q1 (trailing period):** Added.

Drive-by improvements while in there:
- Drop the explicit `KeyError` schema-guard — pandas raises naturally on the missing column.
- Commit `concatenateGenomeFasta` fixtures uncompressed (`.fna`) for legibility; gzip them at test time via a new `gzipFixtureDir` helper in `tests/helpers/FileHelpers.groovy`.
- Rename the filter script to use underscores so the test can import it directly (matches the `enumerateChildTaxa` pattern); drops the `subprocess` + `tsv_factory` plumbing.

_Posted by Claude Code on Jo's behalf._

modules/local/filterViralGenbankMetadata/resources/usr/bin/filter-viral-genbank-metadata.py
jo-delia-faraguna, line 1: Moved to `filter_viral_genbank_metadata.py` so we can import it for `pytest`. Also refactored so we can test just the filtering to avoid having to write files during tests.

willbradshaw (approved):
I haven't read the code in great detail, but what I did read looks good, the CI is passing, and the description/goal sgtm.

PR #760, opened 2026-05-06
Improve Nextflow version checking CI

willbradshaw:
Makes 3 changes to Nextflow version checking CI:
- Changes from prioritizing versions by strict chronology to semantic versioning; patches to older LTEs no longer cause the test to fail.
- Adds a `.nextflowignore` file (analogous to `.trivyignore`) to allow new Nextflow versions to be ignored, either permanently or with a defined expiry date.
- Updates the pinned Nextflow version to 25.10.5 so the updated test now passes.


willbradshaw (commented):

## Correctness

- **C1** (**suggestion**): `parse_nextflowignore` raises immediately on the first malformed line, but does not validate uniqueness. Two entries for the same version (e.g. one permanent and one expirable, or two `exp:` lines with different dates) silently collapse into one due to the `set` semantics, with the later entry's expiration winning or losing arbitrarily depending on order. Consider detecting duplicates in `.nextflowignore` and raising `ValueError`, so a maintainer who appends a new `exp:` line for an already-present version sees an explicit error instead of an undefined merge.
- **C2** (**nit**): `parse_nextflowignore` strips inline comments via `raw.split("#", 1)[0]`, which means a `#` inside the version field — although unreachable for valid semver — would be silently truncated rather than reported. Not a real bug for the X.Y.Z grammar, but if the format ever grows (e.g. pre-release tags), this becomes a footgun. Splitting only on `<whitespace>#` would be more defensive.

## Testing

- **T1** (**suggestion**): There is no test exercising the `select_target_version` path where `releases` is empty (vs. all-ignored). `select_target_version([], set())` raises `"No Nextflow release candidates supplied."`, but no test covers this branch — `bin/test_check_nextflow_version.py:TestSelectTargetVersion` only tests the all-ignored case. Add a parametrised case or dedicated test for the empty-releases branch so the distinct error message is locked in.
- **T2** (**nit**): `TestParseNextflowignore.test_malformed_entries_raise` uses a single shared `match="Malformed entry"`, which is fine, but the `25.10.3 exp:30-01-01` and `25.10.3 exp:2030/01/01` cases also exercise the `IGNORE_ENTRY_PATTERN` regex. If you ever loosen the regex, these would silently move into the `Invalid expiration date` branch and the test would still pass. Pinning the expected message per case (e.g. via `pytest.param(..., id=...)` with a per-case match) would make regressions louder.

## Documentation

**No issues found.** `docs/ci.md` is updated with a clear description of the new `.nextflowignore` format, expiration semantics, and the highest-semver-among-non-ignored selection rule; the CHANGELOG entry summarises the behaviour change; and the `.nextflowignore` file itself carries an in-file format reference. No other docs reference the removed `EXCLUDED_VERSIONS` constant.

## Quality

- **Q1** (**suggestion**): `check_pinned_against_target` is now a one-line wrapper around an equality check whose docstring is longer than the body. After the refactor it adds little over inlining `if pinned != target: raise ValueError(...)` directly in `main()`. Either inline it, or — if the abstraction is being kept for testability — drop the parametrisation in `TestCheckPinnedAgainstTarget.test_mismatch_fails` (the three rows all exercise identical logic; the comments documenting "below/above/typo" are valuable but the cases are redundant from a coverage perspective).
- **Q2** (**nit**): `DEFAULT_RELEASES_URL` is named for a default, but `parse_args` exposes `--releases-url` as a free-form string. Callers can pass arbitrary URLs (used in tests via mocked `urlopen`, so the value does not actually matter), but no validation is performed. Consider either dropping the CLI flag (tests already mock `urlopen`, so the URL is never read in practice) or noting in the help text that it is intended for testing only.
- **Q3** (**nit**): The two new top-level docstring sentences in `bin/check_nextflow_version.py` describe the algorithm ("highest semver-sorted… after filtering out prereleases, drafts, and any versions listed in .nextflowignore") which duplicates the docstrings of `fetch_releases` / `select_target_version` / `parse_nextflowignore`. A shorter top-level summary would reduce drift risk.

## Risks

**No issues found.** The PR description already covers the key risk (semver-max vs. chronological-max, illustrated by the 25.10.5 vs. 26.04.0 case) and explains both the permanent and time-limited ignore entries with their justifications. The `packaging>=24.0` dependency is mainstream and is installed via the existing `setup-python` composite action, so CI behaviour is preserved. No interface or output-format changes outside the version-check workflow itself.

_Reviewed after 4 iteration(s) of automated review-and-fix. 6 findings were automatically resolved._

willbradshaw:
[Claude Code] Responses to the automated review:

**T1, Q2, Q3** — Done in a73832dd:
- T1: Added `TestSelectTargetVersion.test_raises_when_no_releases` covering `select_target_version([], set())`, locking in the distinct "No Nextflow release candidates supplied." message.
- Q2: Added `help=` to `--releases-url` documenting that the override is intended for tests.
- Q3: Trimmed the top-level module docstring to a one-line summary; algorithm details remain in the per-function docstrings (single source of truth).

**C1** (duplicate-entry detection) — Dismissed: out of scope for this PR. The current parser validates format and dates; adding uniqueness validation is a defensible follow-up but expands scope beyond "introduce the `.nextflowignore` mechanism." Not tracking as an issue for now.

**C2** (`#` vs. `<whitespace>#` splitter) — Dismissed: the X.Y.Z grammar enforced by `IGNORE_ENTRY_PATTERN` makes a `#` inside the version field unreachable. Defensive code for hypothetical future grammar extensions is over-engineering by project convention; if the grammar evolves, the splitter is one of several places that would need revisiting together.

**T2** (per-case `match=` strings in `test_malformed_entries_raise`) — Dismissed: every line in the parametrize set is a malformed format that fails at the regex layer, so they all legitimately share the same error. Per-case `match=` strings would tighten coupling to the current message text without strengthening the assertion's semantic content.

**Q1** (inline `check_pinned_against_target` or drop redundancy in `test_mismatch_fails`) — Dismissed: the helper exists for direct testability without going through `main()`, and the parametrize cases (below/above/typo) document distinct *semantic* situations a maintainer would encounter — that documentation value is what the inline comments call out, even though the underlying check is the same equality test. Inlining or trimming would lose more than it saves.

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # v3.2.1.5-dev
 
 - Add CVE-2026-4878 (libcap2), CVE-2026-33845 (libgnutls30), CVE-2026-33846 (libgnutls30), CVE-2026-42311 (Pillow in multiqc), and GHSA-82j2-j2ch-gfr8 (rustls-webpki embedded in Polars in multiqc) to `.trivyignore`; no fix currently available for any, and none is exercised by our pipeline.
+- Replace the hardcoded `EXCLUDED_VERSIONS` constant in `bin/check_nextflow_version.py` with a `.nextflowignore` config file supporting permanent and time-limited (`exp:YYYY-MM-DD`) ignores; switch target selection to highest-semver-among-non-ignored. Ignore `26.04.0` until 2026-06-01 to suppress the upgrade until nf-test 0.9.6 ships.
```
jo-delia-faraguna: **[nit]** `until nf-test 0.9.6 ships` is no longer accurate. Better to say we're deferring until we're ready to bump stable so we can keep main and stable on the same major version?

.nextflowignore
```diff
@@ -0,0 +1,24 @@
...
+# Format (one entry per line):
+#     <X.Y.Z>                    -> permanent ignore
+#     <X.Y.Z> exp:YYYY-MM-DD     -> ignore until YYYY-MM-DD, then warn-and-unignore
+#
+# Lines beginning with '#' and blank lines are ignored; trailing `# ...`
+# comments on entry lines are also stripped. Entries with a past `exp:` date
+# print a warning and are treated as absent, so stale temporary ignores cannot
+# accumulate silently.
+#
+# When the highest-semver upstream Nextflow release matches an active entry,
```
jo-delia-faraguna: **[nit]** Not sure what `an active entry` means.

.nextflowignore
```diff
@@ -0,0 +1,24 @@
...
+# the version check falls back to the next-highest non-ignored release.
+
+# Broken release: AWS Batch job submission issues. Permanently excluded; this
+# version was superseded almost immediately and we will never pin to it.
+25.10.3
+
+# nf-test 0.9.6 (compatible with Nextflow 26.04.x) is not yet released. Once
```
jo-delia-faraguna: **[nit]** This is not accurate, was a mistaken diagnosis on my part.

configs/profiles.config
```diff
@@ -1,7 +1,7 @@
 // Enforce minimum Nextflow version required to run this pipeline
 // The '!' prefix makes this a hard requirement (fails instead of warning)
 manifest {
-    nextflowVersion = '!>=25.10.4'
+    nextflowVersion = '!>=25.10.5'
```
jo-delia-faraguna, line 4: **[fyi]** This will break our nightly (once we merge this to `dev`) and production infrastructure (once we merge this to `main`) until we redeploy with this version. I opened a PR to do this, but looks like `25.10.5` isn't out on [bioconda](https://anaconda.org/channels/bioconda/packages/nextflow/overview) yet so we can't build the container.

bin/check_nextflow_version.py
```diff
@@ -11,140 +11,199 @@
...
-    Fetch the latest Nextflow release version from GitHub API.
+    Parse a .nextflowignore file and return the active ignore set.
+
+    Each non-comment, non-blank line must match `<X.Y.Z>` optionally followed
+    by ` exp:YYYY-MM-DD`. Entries with an expiration in the past are dropped
+    (with a warning to stderr) so stale ignores cannot accumulate silently.
 
     Args:
-        api_url (str): The GitHub API URL to fetch the latest release.
+        ignore_path (Path): Path to the .nextflowignore file. A missing file
```
jo-delia-faraguna, line 98: **[nit]** Why allow this?

bin/check_nextflow_version.py
```diff
@@ -11,140 +11,199 @@
...
-    )
-    with urllib.request.urlopen(request, timeout=30) as response:
-        data = json.loads(response.read().decode())
-    tag = data["tag_name"].removeprefix("v")
-    return validate_semver(tag, api_url)
-
+    if not ignore_path.exists():
+        return set()
+    today = today if today is not None else date.today()
+    active: set[str] = set()
```
jo-delia-faraguna: **[nit]** I find `active` to be a bit confusing. Maybe `currently_ignored` or something?

bin/test_check_nextflow_version.py
```diff
@@ -68,181 +69,324 @@ def test_invalid(self, tmp_path: Path, content: str) -> None:
...
+
+    def _write(self, tmp_path: Path, body: str) -> Path:
+        path = tmp_path / ".nextflowignore"
+        path.write_text(body)
+        return path
+
+    def test_missing_file_returns_empty(self, tmp_path: Path) -> None:
+        assert parse_nextflowignore(tmp_path / "nope") == set()
+
+    def test_empty_file_returns_empty(self, tmp_path: Path) -> None:
```
jo-delia-faraguna: **[nit]** Probably easy to parametrize these?

bin/test_check_nextflow_version.py
```diff
@@ -68,181 +69,324 @@ def test_invalid(self, tmp_path: Path, content: str) -> None:
...
-        releases_data = [
-            {"tag_name": "v25.10.2", "prerelease": False, "draft": False},
-            {"tag_name": "v25.10.1", "prerelease": False, "draft": False},
-        ]
-        with patch("urllib.request.urlopen", return_value=self._mock_releases_response(releases_data)):
-            with pytest.raises(ValueError, match="Could not find any non-excluded release"):
-                get_latest_non_excluded_version({"25.10.2", "25.10.1"})
 
+class TestSelectTargetVersion:
+    def test_picks_highest_semver_not_chronological(self) -> None:
```
jo-delia-faraguna: **[nit]** Could also be parametrized.

bin/test_check_nextflow_version.py
```diff
@@ -68,181 +69,324 @@ def test_invalid(self, tmp_path: Path, content: str) -> None:
...
-                    else:
-                        with pytest.raises(ValueError, match=expected_error):
-                            main()
+        ignore_file = tmp_path / ".nextflowignore"
+        with patch("sys.argv", self._argv(config, ignore_file)), pytest.raises(
+            ValueError, match="Could not find nextflowVersion",
+        ):
+            main()
+
+    def test_ignored_target_falls_back_to_next(
```
jo-delia-faraguna: **[nit]** These could be parametrized as well.

bin/check_nextflow_version.py
```diff
@@ -11,140 +11,199 @@
...
+    if not eligible:
+        if not releases:
+            raise ValueError("No Nextflow release candidates supplied.")
+        raise ValueError(
+            f"No eligible Nextflow release: all {len(releases)} candidates "
+            f"are in the ignore set ({sorted(ignored)}).",
+        )
+    return max(eligible, key=Version)
+
+def check_pinned_against_target(pinned: str, target: str) -> None:
```
jo-delia-faraguna: **[nit]** I find `target` a bit confusing, especially for the error messages. It might be more legible to say `latest` or something more specific. I found myself having to double check which is `pinned` and which is `target`.

jo-delia-faraguna (approved):
LGTM, nits only.

Flagging [this](https://github.com/securebio/nao-mgs-workflow/pull/760/changes#r3204383107) comment in particular about downstream infra.

CHANGELOG.md
willbradshaw: [Claude Code] Done in db586f70 — dropped the "until nf-test 0.9.6 ships" rationale; entry now just says "Ignore `26.04.0` until 2026-06-01."

.nextflowignore
willbradshaw: [Claude Code] Done in db586f70 — reworded to "If the highest-semver upstream release is ignored here, the check falls back to the next-highest non-ignored release."

.nextflowignore
willbradshaw: [Claude Code] Done in db586f70 — replaced the multi-line nf-test rationale with "Defer 26.04.0; revisit on expiry."

bin/check_nextflow_version.py
willbradshaw: [Claude Code] Kept the behavior. A missing `.nextflowignore` is treated as no-ignores so the script stays usable on (a) branches/forks that haven't created the file yet and (b) tests that pass a non-existent path as the "no ignores" scenario. Tightening to fail-on-missing would be a real behavior change we'd want to do separately if at all.

bin/check_nextflow_version.py
willbradshaw: [Claude Code] Done in db586f70 — renamed `active` → `currently_ignored` throughout `parse_nextflowignore` and the main-flow print line.

bin/check_nextflow_version.py
willbradshaw: [Claude Code] Done in db586f70 — renamed `target` → `latest_eligible` in the `check_pinned_against_target` parameter, the main-flow variable, the print line ("Latest eligible Nextflow version"), and the error message ("pinned X != latest eligible Y").

bin/test_check_nextflow_version.py
willbradshaw: [Claude Code] Done in b30ca5f5 — collapsed the empty-file / comments-only / permanent / active-expirable / expiring-today / trailing-comment cases into one parametrized `test_returns_currently_ignored_set` with `pytest.param(..., id=...)`.

bin/test_check_nextflow_version.py
willbradshaw: [Claude Code] Done in b30ca5f5 — parametrized into `test_returns_highest_semver_non_ignored` over `(ignored, expected)` with the shared `RELEASES` factored to a class attribute.

bin/test_check_nextflow_version.py
willbradshaw: [Claude Code] Done in b30ca5f5 — parametrized into `test_main_happy_path` over `(pinned, ignore_body, releases, expected_latest_eligible)`. Raises-tests stayed separate because they assert a different shape.

PR #761, opened 2026-05-07
Rework MAKE_VIRUS_GENOME_DB to filter viral accessions before download (COMP-1980)

jo-delia-faraguna:
Reworks `MAKE_VIRUS_GENOME_DB` to enumerate every viral assembly up-front via a single `datasets summary` call, filter the kept accessions by host infection and assembly status, then chunk them into N-line files for parallel download. Caps the long-tail wallclock — Riboviria's 198k-assembly shard ran ~11h end-to-end on the prior taxon-mode flow ([COMP-1980](https://linear.app/securebio-nao/issue/COMP-1980/rework-make-virus-genome-db-to-enumerate-all-viral-accessions-filter)) — and avoids Fusion metadata corruption during rehydrate ([COMP-1680](https://linear.app/securebio-nao/issue/COMP-1680/make-downloadviralgenomes-robust-to-fusion-errors)) by running each download task on `/scratch` with a single batched `mv` instead of the prior per-file `find -exec mv`.

**Changes:**
- New `ENUMERATE_VIRAL_ACCESSIONS` module: single up-front `datasets summary` call yielding one combined metadata.tsv. No special-case empty-taxon handling — failures are loud.
- `FILTER_VIRAL_GENBANK_METADATA` now computes `species_taxid` rollup internally from `virus_db` (no longer requires `species_taxid` on the input metadata, since `PREPARE_VIRAL_METADATA` runs downstream of filtering now). Emits `chunk_NNNN.txt` files of accessions instead of a single `accessions.csv`; drops the `paths.csv` emit.
- `DOWNLOAD_VIRAL_GENOMES`: input changed from `val(taxid)` to `path(accession_chunk)`; uses `datasets download genome accession --inputfile`; adds Nextflow `scratch '/scratch'` directive; replaces `find ... -exec mv {} \;` with batched `xargs -0 mv -t`; adds an explicit count-verification check; drops the per-chunk metadata output.
- `PREPARE_VIRAL_METADATA` adds a new `ncbi_paths.txt` emit (mirroring `local_filename`) so `CONCATENATE_GENOME_FASTA` keeps its current paths-file interface.
- `MAKE_VIRUS_GENOME_DB` subworkflow rewired: drops `ENUMERATE_CHILD_TAXA` + `CONCATENATE_TSVS` calls and the `taxonomy_nodes` input; adds `chunk_size` input; `download_ch.genomes.collect()` flattens across chunks before passing to `PREPARE_VIRAL_METADATA`.
- New `params.viral_accession_chunk_size`: 10000 in `configs/index.config`, 1 in `configs/index-for-run-test.config` (the test only downloads HDV, so `chunk_size = 1` still exercises the multi-chunk fan-out path).
- Removed orphaned `tests/configs/empty_taxon.config` and the taxon-mode datasets stub at `tests/modules/local/downloadViralGenomes/stub_bin/datasets` — both unreachable after the input-mode swap.
- Tests: new `enumerateViralAccessions/main.nf.test` (HDV success + non-existent-taxid failure); updated `downloadViralGenomes/main.nf.test` to use a chunk file derived from a setup `ENUMERATE_VIRAL_ACCESSIONS` call (avoids hardcoding a refseq assembly accession that could go stale); updated `prepareViralMetadata/main.nf.test` to assert the new `paths` emit. Python tests: 23/23 pass for `filter_viral_genbank_metadata.py` (added `write_accession_chunks` coverage + species-rollup case) and `prepare_viral_metadata.py` (updated for new `output_paths` arg).
- Version bumped to `3.2.2.0-dev` (Results-level): output schemas preserved but the accession-mode swap could in principle yield a different genome set in edge cases (e.g. taxid reassignment between summary and download), so results aren't guaranteed directly comparable across the boundary.


jo-delia-faraguna:
Superseded by the COMP-1680 chain: #779 (CONCATENATE_GENOME_FASTA SIGPIPE) → #780 (DOWNLOAD_VIRAL_GENOMES scratch + xargs) → #781 (chunked viral accession refactor). This PR's content lives in #781, minus the stale-rebase reverts to nextflowVersion (now 25.10.5 on dev via #765) and fusion.exportStorageCredentials (now defaulted to false on dev via #764). Closing.

PR #764, opened 2026-05-08
Default fusion.exportStorageCredentials to false in Batch profiles

dp-rice:
## Summary

In the `standard`, `batch`, and `test_run` profiles, `fusion.exportStorageCredentials` is now `false`. Containers obtain AWS credentials from the EC2 instance role by default, or from a job role attached via `--batch_job_role <ARN>` when one is provided. The `ec2_s3` profile is unchanged.

This returns mgs-workflow to Nextflow's framework default for this setting; the explicit `true` was a workflow-side override.

I've tested that this doesn't break the CI (by changing target branch to ci-test) and that I can run nextflow in batch mode targeting `nao-mgs-staging-test`.

## Mechanics

- The conditional `fusion.exportStorageCredentials = !params.batch_job_role` becomes the constant `false`. The `aws.batch.jobRole = params.batch_job_role ?: null` line is unchanged, so passing `--batch_job_role <ARN>` continues to work exactly as before.
- Containers without a job role rely on the EC2 instance role via IMDS. `docs/batch.md` Step 1 already requires `AmazonS3FullAccess` on the instance role for this to be sufficient.

## Compat

- **No change** for users who pass `--batch_job_role <ARN>` (already gets `false`).
- **No change** for users running on AWS Batch with the documented instance-role setup (`AmazonS3FullAccess`).
- **Breaking** for users on AWS Batch with a hardened instance role lacking S3 access who don't pass `--batch_job_role`. Such users either need to grant S3 to the instance role or pass a job role. CHANGELOG flags this.
- `ec2_s3` (single-EC2 fusion runs, not Batch) is intentionally untouched.

## Test plan

- [x] Target this PR (or a successor) at `ci-test` to exercise `test-chained.yml` end-to-end and confirm the chained run completes.
- [x] Verified locally: a routine-orchestrator run with this setting completes via the EC2 instance role.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # v3.2.1.5-dev
 
+- Default `fusion.exportStorageCredentials = false` in the `standard`, `batch`, and `test_run` profiles. Users who pass `--batch_job_role <ARN>` see no change. Users running on AWS Batch without a job role now rely on the EC2 instance role for S3 access. The `ec2_s3` profile is unchanged.
 - Fix `samtools view` failure on duplicate sequence IDs in the concatenated viral genome FASTA produced by `CONCATENATE_GENOME_FASTA`. NCBI `datasets` sometimes returns superseded ("previous") assembly versions alongside current ones, causing duplicate accessions.
     - `DOWNLOAD_VIRAL_GENOMES` now emits an `assembly_status` column in the per-taxid metadata TSV (via `dataformat tsv genome --fields ...assminfo-status`), which `PREPARE_VIRAL_METADATA` flows through to `FILTER_VIRAL_GENBANK_METADATA`, which keeps only rows with `assembly_status == 'current'` (dropping `previous`, `replaced`, `suppressed`, etc.).
     - `CONCATENATE_GENOME_FASTA` now runs `seqkit rmdup --by-name` after concatenation as a defense-in-depth check. Switched the process from the `seqtk` container to the existing `seqkit` container (no `seqtk` invocation in the script).
--- a/configs/profiles.config
+++ b/configs/profiles.config
@@ -22,7 +22,7 @@ params {
 profiles {
     standard { // Run on AWS Batch
         fusion.enabled = true
-        fusion.exportStorageCredentials = !params.batch_job_role
+        fusion.exportStorageCredentials = false
         aws.batch.jobRole = params.batch_job_role ?: null
         process.executor = "awsbatch"
         process.errorStrategy = "retry"
@@ -31,7 +31,7 @@ profiles {
     }
     batch { // Run on AWS Batch
         fusion.enabled = true
-        fusion.exportStorageCredentials = !params.batch_job_role
+        fusion.exportStorageCredentials = false
         aws.batch.jobRole = params.batch_job_role ?: null
         process.executor = "awsbatch"
         process.errorStrategy = "retry"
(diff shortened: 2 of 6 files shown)


dp-rice (commented):

## Correctness
**No issues found.** The three identical edits in `configs/profiles.config` correctly replace `fusion.exportStorageCredentials = !params.batch_job_role` with the constant `false`. When `--batch_job_role` is set, the resulting value matches the prior behavior; when it is not, the value changes from `true` to `false`, which is the intended new default. (The `check-nextflow-version` CI failure is a pre-existing failure on `dev`, unrelated to these changes.)

## Testing
**No issues found.** No automated test applies to a Nextflow profile constant change. The PR description documents manual validation: a chained CI run via the `ci-test` base, plus a routine-orchestrator run on `nao-mgs-staging-test`.

## Documentation
- **D1** (**issue**): Removing the `eval "$(aws configure export-credentials --format env)"` step from `installation.md` and `troubleshooting.md` leaves the docs internally inconsistent. Three places still depend on those env vars:
  - `docs/testing.md:102-106` still shows the eval block and says it is "described in the installation doc" — that cross-reference is now broken.
  - `bin/run-nf-test.sh:6-11` still hard-requires `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY`, which `installation.md` step 6 (`nf-test test`) implicitly relies on.
  - The `ec2_local` profile (`docker.runOptions = '... -e AWS_ACCESS_KEY_ID=\$AWS_ACCESS_KEY_ID ...'`) and the `ec2_s3` profile (`fusion.exportStorageCredentials = true`, unchanged) still need the env vars.

  As written, a user following `installation.md` to run tests, or using `ec2_local` / `ec2_s3`, will hit `AccessDenied` with no documented recovery. Either retain the eval step in `installation.md` (scoped to the contexts that still need it: tests, `ec2_local`, `ec2_s3`) or update `testing.md` and the running-tests step in `installation.md` to inline the eval and drop the broken cross-reference.
- **D2** (**nit**): `docs/batch.md:117` says "Step 3 above configures this with `AmazonS3FullAccess`". Step 3 attaches the instance role; the `AmazonS3FullAccess` requirement is specified in Step 1 (item 4). The PR description itself says "Step 1 already requires `AmazonS3FullAccess`", which is the more accurate phrasing.

## Quality
**No issues found.** The change is direct and minimal. Replacing the conditional with the constant matches the new default policy, and `aws.batch.jobRole = params.batch_job_role ?: null` is preserved so `--batch_job_role` still works as before.

## Risks
**No issues found.** The "Compat" section enumerates each user category (Batch users with a job role, Batch users on the documented instance-role setup, Batch users with a hardened instance role and no job role, `ec2_s3` users), and the CHANGELOG entry reiterates the breaking case. A human reviewer has the context needed to evaluate impact.

dp-rice:
[Claude Code] Addressed both documentation comments in ecdcef70.

**D1**: Verified the actual scope of the requirement before patching:

- `bin/run-nf-test.sh:6-11` hard-requires `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY` (explicit pre-flight check); since `testing.md` recommends running tests via this script, the eval is needed there. Re-added it in `installation.md` step 6 ahead of `nf-test test`, and dropped the now-broken "as described in the installation doc" wording from `testing.md` (the eval block is shown inline there).
- `ec2_local` and `ec2_s3` do **not** strictly require the eval if `~/.aws/credentials` is set up (per installation step 2): `ec2_local` mounts `$HOME/.aws` into containers so in-container AWS tools find credentials in the file, and `ec2_s3` relies on Nextflow's standard credentials chain to pick up the file before `fusion.exportStorageCredentials = true` propagates them. So no doc additions for those profiles.

**D2**: `docs/batch.md` now attributes the `AmazonS3FullAccess` requirement to Step 1 (where it's specified) rather than Step 3 (which only attaches the role).

willbradshaw (approved):

PR #763, opened 2026-05-07
Speed up DOWNLOAD_VIRAL_GENOMES using scratch and parallelized mv

jo-delia-faraguna:
Speed up `DOWNLOAD_VIRAL_GENOMES` using a non-Fusion [scratch](https://docs.seqera.io/nextflow/reference/process#scratch) directory and parallelized `mv` command.

Chained on #762 

Closes [COMP-1680](https://linear.app/securebio-nao/issue/COMP-1680/make-downloadviralgenomes-robust-to-fusion-errors)

## Context
- Previous production index builds took ~15hrs to complete for Riboviria
  - ~10hrs of that was spent on the `find output/ncbi_dataset/data -name '*.fna.gz' -exec mv {} ${taxid}_genomes/ \;` step
  - ~3hrs on the `rm -rf output/ output.zip raw_metadata.tsv` step.
  - Fusion treats each of these operations as a `CopyObject` and/or `DeleteObject` since all files are mounted and tracked in S3
  - [Logs](https://us-east-1.console.aws.amazon.com/batch/home?region=us-east-1#/jobs/ec2/detail/863e4753-3414-4ea5-9a56-7095c78a2699) are filled with errors like this, suggesting the Fusion system is choking
    ```
    ERR [seqera-fusion] Failed to load file metadata.
    This indicates an internal state corruption that should not occur.
    Check the `error` field for more details and report this issue.
    error="entry not found" goid=3417571 operation=Release path=/s3/<bucket>/<path>/work/02/ecb123424ed797aed65412ee09fdf5/output/ncbi_dataset/data/<accession>/<genome>.fna.temp
    ```
- I am preparing a more major refactor of the `MAKE_VIRUS_GENOME_DB` subworkflow in #761, but the approach here already reduces runtime to <3hr.

## Module interface change
- `DOWNLOAD_VIRAL_GENOMES.genomes` now emits a single `${taxid}_genomes/` directory path instead of the previous channel of per-file `*.fna.gz` paths. Updated `PREPARE_VIRAL_METADATA` to handle this

## Performance
- I ran the **old** download module for Riboviria on Batch and it took [**15hrs**](https://us-east-1.console.aws.amazon.com/batch/home?region=us-east-1#/jobs/ec2/detail/863e4753-3414-4ea5-9a56-7095c78a2699)
[...]

PR #770, opened 2026-05-10
Fix CONCATENATE_GENOME_FASTA exit 141 + bump memory

jo-delia-faraguna:
- The addition of `set -euo pipefail` in #758 to `CONCATENATE_GENOME_FASTA` fails with a `SIGPIPE` when run on large input directories, which are not represented by our current tests.
  - Added `|| true` to handle the pipe error in the diagnostic printout.
- Separately, this process also hit an OOM when trying to build a new production index, with and without the `seqkit rmdup` command.
  - Bumped resources to `single_cpu_16GB_memory` and confirmed it resolved the issue.
  - I think this is caused by the pending #763 change to emit downloaded directories instead of individual files, which seems to increase the Fusion memory overhead.

### DIFF
--- a/.trivyignore
+++ b/.trivyignore
@@ -35,10 +35,10 @@ CVE-2025-68973 exp:2026-06-30
 # surface for these CVEs is minimal in our pipeline context.
 #
 # TODO: Remove these ignores when a newer ncbi-datasets-cli is available on
-# conda-forge built with Go >= 1.25.9. Check with:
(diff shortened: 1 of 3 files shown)


jo-delia-faraguna:
Superseded by the COMP-1680 chain: #779 → #780 → #781. The `|| true` SIGPIPE swallow landed in #779. The CONCATENATE_GENOME_FASTA memory bump to `single_cpu_16GB_memory` was dropped — that OOM was driven by #763's directory emit, which is not being shipped. Closing.

PR #778, opened 2026-05-11
Temporarily ignore 5 HIGH Go stdlib CVEs in ncbi_datasets container

jo-delia-faraguna:
Temporarily ignore five new HIGH-severity CVEs (CVE-2026-33811, CVE-2026-33814, CVE-2026-39820, CVE-2026-39836, CVE-2026-42499) in the Go stdlib v1.23.4 shipped inside `ncbi-datasets-cli` (the `ncbi_datasets` container).

All five require Go ~= 1.25.10 ~= 1.26.3, similarly to the other already-tracked vulnerabilities in `.trivyignore`. The latest `ncbi-datasets-cli 18.25.1` on conda-forge still ships Go 1.23.4, so no upstream fix is available.

Generated with [Claude Code](https://claude.com/claude-code)

PR #779, opened 2026-05-11
Fix CONCATENATE_GENOME_FASTA SIGPIPE on large genome directories (COMP-1680)

jo-delia-faraguna:
The addition of `set -euo pipefail` in #758 to `CONCATENATE_GENOME_FASTA` fails with a `SIGPIPE` when run on large input directories, which are not represented by our current tests. Add `|| true` to handle the pipe error in the diagnostic printout.

Chained #778 --> **#779** --> #780 --> #781

### DIFF
--- a/.trivyignore
+++ b/.trivyignore
@@ -23,9 +23,10 @@ CVE-2025-6020 exp:2026-06-30
 # GnuPG gpgv: information disclosure / potential code execution (no Debian fix available as of 2026-03-14)
 CVE-2025-68973 exp:2026-06-30
 
-# ── Go stdlib v1.23.4 in ncbi-datasets-cli (ncbi_datasets container) ──
+# ── Go stdlib v1.23.4 in ncbi-datasets-cli 18.21.0 (ncbi_datasets container) ──
 # The NCBI datasets/dataformat binaries are pre-compiled Go executables distributed
-# via conda-forge. They ship Go stdlib v1.23.4; fixes require Go >= 1.25.10+.
+# via conda-forge. They ship Go stdlib v1.23.4.
+# Go ~= 1.25.10 or ~= 1.26.3 is required to address all vulnerabilities.
 # We cannot rebuild these — only NCBI can, by updating their Go toolchain.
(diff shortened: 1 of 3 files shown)


jo-delia-faraguna (commented):

## Correctness

- **C1** (**suggestion**): The `|| true` on line 16 of `modules/local/concatenateGenomeFasta/main.nf` fixes the `head` pipe but the same pattern exists one line above: `ls ${genome_dir} | wc -l` (line 14). If `ls` produces enough output to fill the pipe buffer before `wc -l` finishes, the same SIGPIPE can occur. In practice `wc -l` consumes all input so this is unlikely, but for consistency consider adding `|| true` there as well, or switching to `find ... | wc -l` which avoids the issue entirely.

## Testing

**No issues found.** The existing nf-test in `tests/modules/local/concatenateGenomeFasta/main.nf.test` covers the deduplication logic. The SIGPIPE addressed here only manifests with very large directories that exceed the kernel pipe buffer, which is impractical to reproduce in a unit test fixture.

## Documentation

**No issues found.** This is an internal process fix with no user-facing interface changes. The CHANGELOG entry and inline comment adequately explain the change and its motivation.

## Quality

**No issues found.** The fix is minimal and targeted. The inline comment on line 15 explains *why* `|| true` is needed, which is appropriate for a non-obvious shell idiom.

## Risks

**No issues found.** The `|| true` is scoped to a diagnostic `ls | head` statement and cannot mask failures in the actual data-processing logic, which remains under `set -euo pipefail`.

_Reviewed after 1 iteration(s) of automated review-and-fix. 0 findings were automatically resolved._

jo-delia-faraguna:
[Claude Code] `wc -l` fully consumes its input (reads until EOF), so `ls` will never write to a closed pipe regardless of directory size. SIGPIPE only occurs with `head` because it exits after reading N lines while `ls` is still writing. Adding `|| true` here would be unnecessary and could mask actual errors.

willbradshaw (approved):

PR #780, opened 2026-05-11
Speed up DOWNLOAD_VIRAL_GENOMES via /scratch + xargs mv

jo-delia-faraguna:
Speed up `DOWNLOAD_VIRAL_GENOMES` using a non-Fusion [scratch](https://docs.seqera.io/nextflow/reference/process#scratch) directory and parallelized `mv` command.

Chained #807 --> **this PR**

## Context
- Previous production index builds using `datasets` took ~15hrs to complete for Riboviria ([trace](https://us-east-1.console.aws.amazon.com/s3/object/nao-mgs-index?region=us-east-1&prefix=20260307/output/logging/trace_2026-03-30_11-11-32.tsv))
  - ~10hrs of that was spent on the `find output/ncbi_dataset/data -name '*.fna.gz' -exec mv {} ${taxid}_genomes/ \;` step
  - ~3hrs on the `rm -rf output/ output.zip raw_metadata.tsv` step.
  - Fusion treats each of these operations as a `CopyObject` and/or `DeleteObject` since all files are mounted and tracked in S3. These operations are very slow when working with many files at once.
  - [Logs](https://us-east-1.console.aws.amazon.com/batch/home?region=us-east-1#/jobs/ec2/detail/863e4753-3414-4ea5-9a56-7095c78a2699) are filled with errors like this, suggesting the Fusion system is choking
    ```
    ERR [seqera-fusion] Failed to load file metadata.
    This indicates an internal state corruption that should not occur.
    Check the `error` field for more details and report this issue.
    error="entry not found" goid=3417571 operation=Release path=/s3/<bucket>/<path>/work/02/ecb123424ed797aed65412ee09fdf5/output/ncbi_dataset/data/<accession>/<genome>.fna.temp
    ```

## Performance
- The old `DOWNLOAD_VIRAL_GENOMES` module for Riboviria takes **15hrs on its own**.
- This new implementation takes 5min for `ENUMERATE_VIRAL_ACCESSIONS`, 1min for `FILTER_VIRAL_GENBANK_METADATA`, and **~10min walltime** (in parallel with 5 tasks) for `DOWNLOAD_VIRAL_GENOMES`.
- With `scratch = true`, the per-task working dir for `DOWNLOAD_VIRAL_GENOMES` now lives on the instance EBS volume rather than on Fusion/S3 for the duration of the task.
  - The Riboviria download is only on the order of a few GBs, and SB's internal Batch launch templates typically mount 1TB root EBS volumes, so I don't think we'll run into storage issues.

## Index equality
I compared the most recent SB private index (`s3://nao-mgs-index/20260307/`, built using `3.2.1.2-dev`) against a full production index built with the code in this PR chain (`s3://nao-jo/workflow-index-test-781/`). The indexes are structurally equivalent — total content is within 1%, and the assembly set overlaps 98%.

_Below is Claude Code with human review and spot checking_

### FASTA (`virus-genomes-masked.fasta.gz`)

| Metric | This PR chain | SB private index |
|---|---|---|
| Compressed bytes | 673,251,762 | 678,022,738 |
| Sequence count | 76,090 | 75,358 |

### Metadata (`virus-genome-metadata-gid.tsv.gz`)

| Metric | This PR chain | SB private index |
|---|---|---|
| Total rows | 77,964 | 76,380 |
| Unique assemblies | 45,379 | 43,807 |
| Unique taxids | 5,721 | 5,601 |
| GenBank (`SOURCE_DATABASE_GENBANK`) | 73,867 | 71,555 |
| RefSeq (`SOURCE_DATABASE_REFSEQ`) | 4,097 | 4,825 |

### Assembly overlap

The number of unique assemblies is lower than the number of unique sequences since many are multi-segment assemblies with multiple sequences.

| Category | Count | Explanation | Examples |
|---|---|---|---|
| In both indexes | ~43k (98.0% of old / 94.6% of new) | Core set — same assemblies kept by both builds | e.g. `GCA_027938485.1` |
| Only in new | ~2.5k | Existing assemblies the per-taxon approach missed | e.g. `datasets summary genome accession GCA_000858185.1 --as-json-lines` |
| Only in old (GenBank) | ~150 | NCBI marked these `assembly_status != 'current'` since Mar 7; the #758 filter now drops them | e.g. `datasets summary genome accession GCA_000854765.1 --as-json-lines` |
| Only in old (RefSeq) | ~700 | Suppressed GCF assemblies; their paired GCA versions remain `current` and are present in the new index | e.g. `datasets summary genome accession GCF_000848585.1 --as-json-lines` |


jo-delia-faraguna (commented):

## Correctness

**No issues found.** The `find -print0 | xargs -0 -r mv -t` replacement is a correct and safe equivalent of the original per-file `find -exec mv` approach. The `-r` flag prevents `xargs` from invoking `mv` when `find` returns no results, and `-0`/`-print0` handles filenames with whitespace or special characters. The `scratch = '/scratch'` directive is applied only to Batch profiles where `/scratch` is already volume-mounted, which is consistent with the existing infrastructure pattern.

## Testing

**No issues found.** The new "Should download genomes with scratch enabled" test case (`tests/modules/local/downloadViralGenomes/main.nf.test:58-80`) exercises Nextflow's scratch stage-out path via a dedicated config (`tests/configs/scratch_download.config`) that sets `scratch = true`. This verifies that file staging works correctly with the `xargs`-based move. The test uses `scratch = true` (Nextflow temp directory) rather than `scratch = '/scratch'` (absolute path), which is the right choice for CI portability. The existing happy-path, failure, and empty-taxon tests are preserved.

## Documentation

**No issues found.** Both `docs/usage.md` and `docs/troubleshooting.md` are updated with user-facing guidance on the new `scratch` directive -- including how to disable it and what infrastructure assumptions it relies on. The process-level comment in `modules/local/downloadViralGenomes/main.nf:2` and the per-profile comments in `configs/profiles.config` explain why the directive is set and where it is configured. The trailing-newline issue in `docs/troubleshooting.md` from the prior review has been fixed.

## Quality

**No issues found.** The profile config changes are consistent across all three Batch profiles (`standard`, `batch`, `test_run`), with identical comments. The CHANGELOG entry is well-written and correctly placed at the top of the current version section. The test config cleanly extends `index.config` with a single override, following the existing pattern in `tests/configs/`.

## Risks

**No issues found.** The PR description thoroughly addresses the key risks: the `/scratch` space requirement is discussed with reference to SecureBio's standard 1TB EBS volumes, the performance improvement is quantified (15hrs to 2.5hrs), and the known limitation at production scale (197k files crashing Nextflow stage-out) is explicitly called out as addressed by downstream chained PRs. The `scratch` directive is scoped to Batch profiles only, so `ec2_local` and `ec2_s3` users are unaffected.

_Reviewed after 2 iteration(s) of automated review-and-fix. 2 findings were automatically resolved._

willbradshaw:
@jo-delia-faraguna I'm confused about

> This PR's implementation fails at production scale because the 197k Riboviria genome files crash Nextflow's stage out behavior.

Is this even something we should be merging before we merge #781? I'm not sure we should be merging code that we know breaks at scale.

configs/profiles.config
willbradshaw, line 1: I don't love repeating this verbatim across a bunch of different profiles. Could we not handle this with a closure, perhaps in the process definition itself? (e.g. if fusion.enabled = true -> scratch = '/scratch')

willbradshaw: Or, if we want to extend this to other processes in the future, create a `use_scratch` label and put the closure there.

jo-delia-faraguna: I was unable to do a `profiles.config`-level closure for this: let me know if this is acceptable.

willbradshaw (approved):

PR #783, opened 2026-05-11
Configure fastp to retain mates only at least 35bp

evanfields:
This ensures we don't end up with reads too short for the kmer methods we use to identify viral reads or classify taxonomically.


.trivyignore
```diff
@@ -25,20 +25,20 @@ CVE-2025-68973 exp:2026-06-30
 
 # ── Go stdlib v1.23.4 in ncbi-datasets-cli 18.21.0 (ncbi_datasets container) ──
 # The NCBI datasets/dataformat binaries are pre-compiled Go executables distributed
-# via conda-forge. They ship Go stdlib v1.23.4; fixes require Go >= 1.24.6+.
```
jo-delia-faraguna: **[nit]** Rebase/merge the remote `dev` since these changes already landed I think?

modules/local/fastp/main.nf
jo-delia-faraguna, line 1: **[nit]** Is it worth it to add a test to `tests/modules/local/fastp/main.nf.test` that verifies we do filter out e.g. 30bp reads?

jo-delia-faraguna (approved):
LGTM.

mikemc (approved):
Looks good, but I suggest that you reference https://github.com/securebio/nao-mgs-workflow/issues/739 somewhere that makes it into the final squash commit message as that has more context than existing commit messages.

I agree that @jo-delia-faraguna 's comment about adding a test to verify 30bp reads are filtered would be nice but isn't a blocker

PR #799, opened 2026-05-12
Downgrade Nextflow to 25.10.4; permanently ignore 25.10.5 (bioconda gap)

willbradshaw:
`25.10.5` was skipped over in bioconda, so our conda-based provisioning automations can't install it. Downgrade the pinned Nextflow version to `25.10.4` and add a permanent `25.10.5` ignore to `.nextflowignore` so the version-currency check stops demanding we move to it. Updates the in-flight v3.2.1.5 CHANGELOG entry (release PR #798 hasn't merged yet) to reflect what actually ships, rather than appending a self-reverting bullet.


jo-delia-faraguna (approved):
LGTM, thanks.

PR #805, opened 2026-05-14
Fix manual-reset workflow's Show version info step

willbradshaw:
The `Show version info` step in `manual-reset.yml` was failing with `fatal: invalid reference: stable` because `actions/checkout@v4` only creates a local branch for the checked-out ref — other branches are only present as `origin/<name>` remote-tracking refs. Replace the brittle `git checkout <ref> -- pyproject.toml` / revert pattern with `git show origin/<ref>:pyproject.toml`, which reads file content without touching the working tree.

**Verification:** `manual-reset.yml` is gated on the `stable-reset` environment and only runs via `workflow_dispatch` on `main`, so the fix can't be exercised in a normal PR run. End-to-end verification will happen at the next stable reset.


willbradshaw (commented):

## Correctness

**No issues found.** `git show origin/<ref>:pyproject.toml` correctly avoids the bug in the original code: `actions/checkout@v4` only materializes the requested ref as a local branch, so `git checkout stable -- pyproject.toml` fails with `invalid reference: stable`. The replacement reads from the remote-tracking ref (always present after fetch-depth: 0) and leaves the working tree untouched, so the subsequent `Reset stable branch to main` step starts from a clean state. The pipeline's error behavior is also acceptable: if `git show` fails, `tomllib.load` on empty stdin raises and python exits non-zero, propagating the failure.

## Testing

- **T1** (**nit**): This is a CI workflow change, so the fix can only be verified by an end-to-end dispatch of `manual-reset.yml`. The PR description doesn't say whether the author has run a dry-run or verified the new step succeeds in a sandbox environment — consider noting verification status (or planned manual verification at the next stable reset) for the reviewer's benefit.

## Documentation

**No issues found.** The change is purely an internal fix to a diagnostic step; no user-facing behavior or interface changes. `docs/ci.md`'s "Manual stable reset" section describes the workflow at a level that does not need updating.

## Quality

- **Q1** (**nit**): The `read_version()` pipeline isn't protected by `set -o pipefail`. The step still fails when `git show` fails (because `tomllib.load` will then raise on empty input), but this is incidental rather than guaranteed. Adding `set -o pipefail` at the top of the `run:` block would make the failure semantics explicit and protect against future changes to the python invocation that might mask the upstream error.

## Risks

**No issues found.** Behavior change is limited to a diagnostic display step in a manually-dispatched workflow already gated by the `stable-reset` environment. The new code reads `pyproject.toml` from `origin/main` / `origin/stable` instead of the (formerly checked-out) local refs — functionally equivalent given `fetch-depth: 0`.

willbradshaw:
[Claude Code] Addressed both nits:

- **T1:** Updated PR description with a "Verification" note. The step can't be exercised in a normal PR run (gated on `stable-reset` environment, `workflow_dispatch`-only), so end-to-end verification will happen at the next stable reset.
- **Q1:** Added `set -euo pipefail` to the `Show version info` `run:` block in d5643936 to make failure semantics explicit.

dp-rice (approved):
Looks good

PR #807, opened 2026-05-14
Chunk viral accession downloads in MAKE_VIRUS_GENOME_DB

jo-delia-faraguna:
Replaces the prior per-child-taxon download flow in `MAKE_VIRUS_GENOME_DB` with an upfront accession enumeration, filter, chunk, and download pipeline.

- The prior implementation's load was very imbalanced: the Riboviria taxon previously accounted for ~197/277k total genomes downloaded and took 15hrs while other downloads completed in minutes.
- The prior implementation also downloaded genomes only to later filter out ~85% of them using infection status, [hard excludes](https://github.com/securebio/nao-mgs-workflow/blob/main/configs/index.config#L40), and [genome patterns to exclude](https://github.com/securebio/nao-mgs-workflow/blob/main/configs/index.config#L31).
- This implementation uses `datasets summary` to enumerate all viral accessions upfront, then filters them using infection status and hard excludes, and then evenly chunks them for parallel downloads.
- This implementation still filters on genome patterns to exclude after downloading genomes since that relies on the genome sequence headers themselves. This is a relatively small fraction of the total filtering.
- The single-call `datasets summary genome taxon 10239` enumeration (all ~277k viral assemblies) has been validated at production scale: the NCBI API returns results in seconds and the response fits comfortably in memory.

Chained **this PR** --> #780

### Configs

- New `params.viral_accession_chunk_size`: 10000 in `configs/index.config` to control download-level parallelization
- Split `datasets_extra_args` into `datasets_summary_extra_args` and `datasets_download_extra_args`
- Test config `assembly_source` changed from `"refseq"` to `"all"` so taxid 2847173 returns both RefSeq + GenBank accessions, exercising the multi-source download path

### New modules

- `ENUMERATE_VIRAL_ACCESSIONS`: single up-front `datasets summary` call to enumerate all viral accessions. Emits one combined `metadata.tsv` file.

### Changed modules

- `FILTER_VIRAL_GENBANK_METADATA`: now computes `species_taxid` internally from `virus_db`. Emits `chunk_NNNN.txt` files of accessions.
- `DOWNLOAD_VIRAL_GENOMES`: now uses the accession chunk files and `datasets download genome accession --inputfile`.
- `PREPARE_VIRAL_METADATA`: now emits `ncbi_paths.txt`.

### Removed modules from `MAKE_VIRUS_GENOME_DB`

- `ENUMERATE_CHILD_TAXA`: replaced by `ENUMERATE_VIRAL_ACCESSIONS` and **completely removed from `mgs-workflow` codebase**
- `CONCATENATE_TSVS`: removed from subworkflow: no longer needed since we generate a single `metadata.tsv` file. Still used elsewhere in `mgs-workflow`, however.

## Benchmarking

I did not benchmark this PR's production index performance alone since Fusion errors still cause the download steps to take hours. I did benchmark the chained #780 PR's performance, however, which includes these changes + the local scratch dir changes and runs in ~15min total walltime across the enumeration and download steps (compared to ~15hrs previously for just the Riboviria download).

The changes introduced in this PR would generate an identical index to the current `main` version. See "Confirming old and new method download the same assemblies for Alphainfluenzavirus influenzae" details section below for my justification for this claim. I more directly compare the index I built in #780 as additional validation.

See #780 for more detailed performance and equality validations.

<details>
<summary>Querying number of accessions in NCBI</summary>

I wanted to figure out how many accessions we are downloading in the old method only to throw them away during filtering. I looked at the [hard excluded taxids](https://github.com/securebio/nao-mgs-workflow/blob/main/configs/index.config#L40C34-L40C57) for a naive underestimate: we also throw away viral genomes not annotated as vertebrate- or human-infecting later.

```bash
# first three are hard excludes
# fourth is all viruses
# fifth is Virobiria
for t in 2731619 2732413 2732411 10239 2559587; do
    n=$(datasets summary genome taxon "$t" --assembly-source all --as-json-lines | wc -l)
    echo "$t: $n"
done
```

This yields

```bash
2731619: 38614
2732413: 0
2732411: 494
10239: 277330
2559587: 198203
```

**So we are throwing away 13% of the viral genomes using hard exclusions.**

Our filtered metadata file for the downloaded accessions used in the [public index](https://us-east-1.console.aws.amazon.com/s3/object/nao-testing?region=us-east-1&prefix=mgs-workflow-test/index-latest/output/results/virus-genome-metadata-gid.tsv.gz) also contains only roughly 40k unique assemblies (some split into multiple entries if they are multi-segment genomes).

```bash
# histogram of number of occurrences of unique assemblies in the latest genome's metadata file
aws s3 cp s3://nao-testing/mgs-workflow-test/index-latest/output/results/virus-genome-metadata-gid.tsv.gz - \
    | zcat \
    | awk -F'\t' 'NR>1 {c[$1]++} END {
            for (k in c) hist[c[k]]++
            print "segments_per_assembly\tn_assemblies"
            for (n in hist) print n "\t" hist[n]
        }' \
    | sort -k1,1n
```

**Therefore, we throwing away ~85% of all downloaded genomes during filtering. In the new implementation, we should actually expect to download on the order of 50k genomes, then filter down to a few less once we filter out excluded sequence header patterns.**

</details>

<details>
<summary>Confirming old and new method download the same assemblies for Alphainfluenzavirus influenzae</summary>

The old method for downloading Alphainfluenzavirus influenzae can be reduced to:

```bash
# download all genomes for Alphainfluenzavirus influenzae
datasets download genome taxon 2955291 --assembly-source all --include genome --dehydrated --filename all.zip
# unzip and rehydrate
unzip -o all.zip -d all/
datasets rehydrate --directory all/ --max-workers 16 --gzip
# parse metadata
dataformat tsv genome --inputfile all/ncbi_dataset/data/assembly_data_report.jsonl --fields accession,organism-tax-id,organism-name,source_database,assminfo-status > all.tsv
```

The new method can be reduced to:

```bash
# enumerate all Alphainfluenzavirus influenzae accessions
datasets summary genome taxon 2955291 --assembly-source all --as-json-lines > acc.jsonl
dataformat tsv genome --inputfile acc.jsonl --fields accession,organism-tax-id,organism-name,source_database,assminfo-status > acc-summary.tsv
cut -f1 acc.tsv | tail -n +2 > acc.txt
# download genomes for enumerated accessions
datasets download genome accession --assembly-source all --include genome --inputfile acc.txt --dehydrated --filename acc.zip
# unzip and rehydrate
unzip -o acc.zip -d acc/
datasets rehydrate --directory acc/ --max-workers 16 --gzip
# parse metadata
dataformat tsv genome --inputfile acc/ncbi_dataset/data/assembly_data_report.jsonl --fields accession,organism-tax-id,organism-name,source_database,assminfo-status > acc-download.tsv
```

**The produced TSVs are identical except for row order.**

</details>


modules/local/downloadViralGenomes/main.nf
```diff
@@ -1,44 +1,33 @@
...
+        datasets download genome accession \\
             --assembly-source ${assembly_source} \\
             --include genome \\
             --no-progressbar \\
             --dehydrated \\
+            --inputfile ${accession_chunk} \\
             ${extra_args} \\
-            --filename output.zip 2> dl_err.txt
-        then
-            cat dl_err.txt >&2
```
jo-delia-faraguna, line 30: Error handling is no longer necessary since we never expect a task to try to download 0 assemblies.

modules/local/downloadViralGenomes/main.nf
```diff
@@ -53,25 +42,13 @@ process DOWNLOAD_VIRAL_GENOMES {
             BACKOFF=\$((BACKOFF * 2))
         done
 
-        # 3. Convert assembly report to TSV with standardized column names.
```
jo-delia-faraguna, line 56: No need to convert metadata since we already generated it upfront in `ENUMERATE_VIRAL_ACCESSIONS`

tests/configs/empty_taxon.config
jo-delia-faraguna, line 1: Not needed anymore: `DOWNLOAD_VIRAL_GENOMES` should never fail to find any assemblies.

tests/modules/local/downloadViralGenomes/stub_bin/datasets
jo-delia-faraguna, line 1: Not needed anymore: `DOWNLOAD_VIRAL_GENOMES` should fail to find any assemblies.

jo-delia-faraguna (commented):

## Correctness

- **C1** (**issue**): In `ENUMERATE_VIRAL_ACCESSIONS` (`modules/local/enumerateViralAccessions/main.nf:27-31`), when `datasets summary` returns zero assemblies (valid taxon, no genome data), `assembly_data_report.jsonl` is empty. The behavior of `dataformat tsv genome --inputfile` on an empty JSONL file is not clearly documented -- if it produces no output at all (no header), the `tail -n +2` in the pipe will also produce nothing, and `metadata.tsv` will contain only the `printf`-generated header with zero data rows. This would then flow into `FILTER_VIRAL_GENBANK_METADATA`, which calls `write_accession_chunks` with zero accessions, raising `ValueError("No accessions passed filter")` and failing the pipeline. This is a hard failure for any valid taxon that happens to have no assemblies. If that's the intended behavior, add a comment in `ENUMERATE_VIRAL_ACCESSIONS` documenting that zero-assembly taxons are expected to fail the pipeline (since the old code had an explicit empty-taxon success path). If it's not intended, handle the zero-row case upstream.

## Testing

- **T1** (**suggestion**): The index workflow integration test (`tests/workflows/index.nf.test:25-26`) asserts `metadata.rowCount >= 2` to validate the chunk fan-out path, with a helpful error message suggesting to broaden `download_virus_taxid` if NCBI publishes only one HDV assembly. However, the test does not exercise the single-chunk path (chunk_size >= total accessions) or the multi-chunk path (chunk_size < total accessions) distinctly. Consider adding a module-level nf-test for `FILTER_VIRAL_GENBANK_METADATA` that verifies chunk file output structure (number of chunks, accessions per chunk) with controlled inputs, rather than relying solely on the end-to-end workflow test with live NCBI data whose accession count can change.

## Documentation

**No issues found.** `docs/config.md` has been updated to reflect the split of `datasets_extra_args` into `datasets_summary_extra_args` and `datasets_download_extra_args`, the new `viral_accession_chunk_size` parameter, and the revised description of `download_virus_taxid`. Inline comments throughout the Nextflow modules and Python scripts explain the design decisions clearly.

## Quality

**No issues found.** The refactoring cleanly separates enumeration, filtering, chunking, and download into distinct steps with clear data flow. The species-level rollup logic in `filter_metadata` is well-commented and tested. The `write_accession_chunks` function has good input validation and the chunk file naming convention (`chunk_NNNN.txt`) is clear. The use of `xargs -0 -r mv -t` in `DOWNLOAD_VIRAL_GENOMES` is a good improvement over per-file `find -exec mv`.

## Risks

**No issues found.** The PR description thoroughly addresses the key risks: the change from per-child-taxon to per-accession-chunk downloads, the expected download reduction (~85% fewer genomes fetched), and the validation that the old and new methods produce identical results. The performance implications are documented with benchmarking data from the chained #780 PR. The new `viral_accession_chunk_size` parameter default (10000) is documented with its rationale.

_Reviewed after 4 iteration(s) of automated review-and-fix. 8 findings were automatically resolved._

jo-delia-faraguna:
- **C1**: We _want_ to fail hard in `ENUMERATE_VIRAL_ACCESSIONS` if there are no assemblies found. Added comment in the module.
- **T1**: The chunking behavior is already well-tested by the Python tests and by the `INDEX` workflow test. Adding an additional module-level test would add nf-test boilerplate for minimal additional coverage.
  - **(If a human reviewer feels it would be important to add this, I'm happy to).**

subworkflows/local/makeVirusGenomeDB/main.nf
```diff
@@ -18,11 +17,12 @@ include { MASK_GENOME_FASTA } from "../../../modules/local/maskGenomeFasta"
 
 workflow MAKE_VIRUS_GENOME_DB {
     take:
-        virus_taxid // Taxid to enumerate child taxa for parallel genome downloads
+        virus_taxid // Top-level taxid to enumerate viral assemblies for
         assembly_source // Assembly source: "genbank", "refseq", or "all"
-        datasets_extra_args // Additional arguments passed to `datasets download genome taxon`
+        datasets_summary_extra_args // Additional args passed to `datasets summary genome taxon` in ENUMERATE
+        datasets_download_extra_args // Additional args passed to `datasets download genome accession` in DOWNLOAD
+        chunk_size // Max accessions per parallel download chunk
```
willbradshaw: I think this should probably go in `other_params` (the more we can move away from long lists of positional arguments and towards named elements in maps, the better)

tests/modules/local/enumerateViralAccessions/main.nf.test
willbradshaw, line 1: For such a critical new process, the test suite here seems a little thin.

tests/workflows/index.nf.test
```diff
@@ -10,6 +10,35 @@ nextflow_pipeline {
         tag "main_index"
         then {
             assert workflow.success
+
```
willbradshaw, line 13: Hmm, this is a lot of new content and makes the test somewhat hard to read. How much of this really needs to be tested here (vs in modules or subworkflows)?

willbradshaw (approved):
Very exciting! A few suggestions, mainly on testing. No need for re-review.

subworkflows/local/makeVirusGenomeDB/main.nf
jo-delia-faraguna: Agree, implemented.

tests/modules/local/enumerateViralAccessions/main.nf.test
jo-delia-faraguna: Agree, I added a few more assertions and another test case.

tests/workflows/index.nf.test
jo-delia-faraguna: Agree, I moved a lot of this down to the module-level. We don't have module-level tests for `FILTER_VIRAL_GENBANK_METADATA` or `ADD_GENBANK_GENOME_IDS` or a subworkflow-level `MAKE_VIRAL_GENOME_DB` test. I'll open an issue to track these for the future.

PR #816, opened 2026-05-20
Throttle head-node S3 concurrency to prevent sentinel fan-out 503s

### LINKED ISSUE #801: DOWNSTREAM workflow fails on large batches at WRITE_SENTINEL_DOWNSTREAM due to simultaneous S3-poll fan-out
## Summary

When a DOWNSTREAM batch contains ~100 groups, the workflow can crash at the final `WRITE_SENTINEL_DOWNSTREAM` step. All per-group sentinels fan out simultaneously on the Nextflow head process and poll S3 in a burst that exceeds the per-prefix request-rate threshold, triggering an unrecoverable cascade (503 → SDK `mark/reset` retry bug → operator-fork semaphore exhaustion). Analytical outputs are safely published; only the sentinel JSONs are missing. Two fixes are proposed below: a small config workaround (Option A) and a structural channel-shape fix (Option B).

## Root cause

`workflows/downstream.nf` wires the sentinel inputs through a global barrier:

```groovy
WRITE_SENTINEL_DOWNSTREAM(
    groups_only_ch,
    input_downstream_ch.mix(logging_downstream_ch, results_downstream_ch).collect(),
    ...
)
```

`.collect()` flattens every emission across every group into one barrier value. Combined with `executor 'local'` on the sentinel module, this means *all* per-group sentinels start simultaneously when the barrier releases. Each sentinel does several `file().exists()` S3 polls. With many groups, the concurrent HEAD burst exceeds S3's per-prefix rate limit.

### Failure chain in `.nextflow.log`

1. Concurrent S3 HEAD bursts from `HashBuilder` / `S3Client` / sentinel `file().exists()`.
2. S3: `503 Please reduce your request rate`.
3. SDK retry attempts to replay the request body, fails with `IllegalStateException: Content input stream does not support mark/reset, and was already read once.`
4. Nextflow re-submits the failed sentinel, can't acquire the operator-fork semaphore (saturated by the other concurrent sentinels), throws `IllegalStateException: Cannot obtain the semaphore to fork operator's body.`
5. Workflow aborts, exit 1. Typical final stats: `succeededCount=N (analytical); failedCount=3; abortedCount=1`.

### Observed (single data points; not stress-tested across all combinations)

- Illumina DOWNSTREAM batch, 100 groups → failed as above.
- Illumina batch, 30 groups → completed cleanly.
- ONT batch, 32 groups → completed cleanly.
- A `nextflow -resume` against the failed batch *also failed before the sentinel step* — the same 503 → mark/reset chain reproduced during cache verification, where `HashBuilder`'s startup thread pool emits a similar burst against thousands of cached work-dir paths. So the natural \"resume to recover\" path is itself vulnerable.

Tested with Nextflow 25.10.4, `nf-amazon` 3.4.4, pipeline 3.2.1.4.

## Option A — config-level throttle (workaround)

Add to default `nextflow.config`:

```groovy
aws {
    client {
        maxConnections = 10
        maxErrorRetry  = 20
    }
}

process {
    withName: WRITE_SENTINEL_DOWNSTREAM {
        maxForks = 10
    }
    withName: WRITE_SENTINEL_RUN {
        maxForks = 10
    }
}
```

NOTE: We typically don't use `withName` for configs. Figure out what these processes have in common and follow the repo conventions for configuring.

Caps the SDK's concurrent S3 connections from the head process so neither burst (cache verification or sentinel fan-out) exceeds S3's rate limit. `maxForks` adds belt-and-suspenders at the sentinel step.

**Pros:**
- Small, contained, cache-safe (these settings don't participate in Nextflow's task-hash).
- Directly verified: applying this override let a failing 100-group batch complete cleanly with zero 503/mark-reset/semaphore signals.
- Also fixes the cache-verification burst on `-resume` (Option B does not).

**Cons:**
- Doesn't address the structural root cause; if a future change adds another head-node S3 burst, the failure could re-emerge.
- The numbers (10, 10) are empirical, not derived from documented S3 limits.
- Performance: lowers head-process S3 throughput at workflow startup (cache verification) and at `PublishDir`/fan-out points. Per-task S3 ops inside Batch worker containers are not affected. Net impact on the tested run was small.

## Option B — structural fix (per-group sentinel barrier)

Replace `.collect()` with a per-group join. Each upstream channel feeding `results_downstream_ch` already emits `[group, file]` tuples; `.collect()` discards the labels. With `groupTuple()`, each sentinel fires when its own group's outputs are ready:

```groovy
per_group_ready = results_downstream_ch
    .groupTuple()        // [group, [files for that group]]
    // ... combine with per-workflow channels (input_*, logging_*) as needed
WRITE_SENTINEL_DOWNSTREAM(per_group_ready, start_time_str, sentinel_params)
```

Same wiring exists in `workflows/run.nf` for `WRITE_SENTINEL_RUN`.

**Pros:**
- Eliminates the root cause; sentinels naturally stagger as groups finish. No magic numbers, scales to any batch size.
- Each group's sentinel arrives sooner, which downstream consumers could exploit.

**Cons:**
- Larger change: channel-shape refactor, plus handling for `input_downstream_ch` / `logging_downstream_ch` which carry workflow-level (not group-level) files.
- Doesn't address the cache-verification burst on `-resume` (that's a `HashBuilder`-side concurrency issue independent of the DAG shape).
- More test surface (empty-group, only-empty-output, etc.).

## Recommendation

Option A is the small fast fix and covers more ground (also helps on `-resume`). Option B is structurally cleaner and worth doing as a follow-up; with B in place, the `maxForks` part of A becomes unnecessary, but the `aws.client.maxConnections` cap is still useful for resume-time protection.

dp-rice:
This is the config fix I used to debug the backlog run. It solved two related problems (#801):

1. With a large DOWNSTREAM batch, `WRITE_SENTINEL_DOWNSTREAM` crashed because it tried to hit S3 too many times simultaneously.
2. When I tried to resume the run, the cache verification also hit S3 limits and crashed.

Both of these steps run from the head node and can make too many simultaneous S3 calls.

Here, I'm proposing two config tweaks:

1. Reduce `aws.client.maxConnections` from 1000 to 50.
2. Cap `maxForks` for the sentinel processes at 10 to prevent too many concurrent jobs hitting S3 from the head node at the same time.

With the backlog run, these changes fixed the problems and did not appear to slow things down appreciably, and Claude thinks it probably wouldn't (though I wouldn't put much stock in that).

I also had Claude look through the git history to see if `maxConnections = 1000` had been introduced to solve a particular problem and it looks like it's been there since very early development.

P.S. I ran `nextflow lint` and only found preexisting errors.

---

[Claude from here]

Closes #801.

On ~100-group DOWNSTREAM batches, `WRITE_SENTINEL_DOWNSTREAM` was crashing: the simultaneous per-group sentinel fan-out polled S3 hard enough to exceed S3's per-prefix request-rate threshold, triggering a 503 cascade that aborted the workflow. The same cascade also reproduces during `-resume` cache verification (HashBuilder polling thousands of cached work-dir paths at startup).

The fix reverts `aws.client.maxConnections` from a longstanding upward override (1000) back to nf-amazon's default (50). The `1000` traces to early INDEX-workflow scale-up work in Jan 2024 with no documented justification. A direct S3 burst test (multi-process, single key) sustained 12k HEAD req/s with zero 503s — well above the documented 5500/s per-prefix limit — so the default 50 is conservatively safe. A new `sentinel` process label on `WRITE_SENTINEL_RUN` and `WRITE_SENTINEL_DOWNSTREAM` carries `maxForks = 10` as a sentinel-layer cap — belt-and-suspenders, and locks in the protection at the sentinel layer so future SDK-pool tuning doesn't silently undo it.

**Changes:**

- `configs/profiles.config`: `aws.client.maxConnections` 1000 → 50 (nf-amazon default). Caps head-node S3 concurrency. Per-task S3 ops inside Batch worker containers are unaffected.
- `modules/local/writeSentinel{Run,Downstream}/main.nf`: add `label "sentinel"`. Both are head-node `exec:` processes that poll many S3 paths via `SentinelUtils.waitForFiles`.
- `configs/resources.config`: add `withLabel: sentinel { maxForks = 10 }`.
- `CHANGELOG.md`: append a Bugfix bullet to the existing `v3.2.2.0-dev` section. No `pyproject.toml` bump — point-level fix on already-bumped dev.

No automated test — the 503 cascade only manifests under real S3 load. The 100-group failing batch from #801 completes cleanly under this config.


dp-rice (commented):

## Correctness

**No issues found.** The 100× drop in `aws.client.maxConnections` is the load-bearing fix and the `maxForks = 10` label cap is a coherent belt-and-suspenders; `SentinelUtils.waitForFiles` uses a serial `findAll`, so 10 concurrent sentinel processes each polling serially yields ~10 in-flight HEAD requests, comfortably under both the SDK pool size and S3's per-prefix threshold.

## Testing

**No issues found.** The 503/SDK-retry/operator-semaphore cascade only reproduces under real S3 load, so the existing `tests/modules/local/writeSentinel*/main.nf.test` snapshots remain the appropriate test surface; they don't load `configs/resources.config` and so are unaffected by the new label, and the PR description records the manual verification on the failing 100-group batch.

## Documentation

- **D1** (**nit**): The inline comment in `configs/profiles.config` explains the *cascade* well but doesn't note that `10` is empirical (chosen from the #801 verification, not derived from a documented S3 limit). That detail is in the PR description but won't survive in `git blame` once the PR is squash-merged — adding a half-sentence to the comment (\"chosen empirically from the failure reproducer in #801\") would make future tuning safer.

## Quality

- **Q1** (**nit**): `label \"sentinel\"` on `WRITE_SENTINEL_RUN` is semantically correct but operationally a no-op — that process only fires once per workflow, so `maxForks = 10` never binds. Worth keeping (it documents the head-node-poller class and pre-positions the cap if a second per-run sentinel is ever added), but the comment in `configs/resources.config` reads as if the cap is active on both processes; clarifying that the runtime constraint is specifically on `WRITE_SENTINEL_DOWNSTREAM`'s per-group fan-out would match reality.

## Risks

- **R1** (**suggestion**): The PR description scopes the throughput impact to \"the head node\" and notes per-task S3 ops inside Batch containers are unaffected (Fusion), which is correct — but the head-node SDK pool is also what services `publishDir` copies and other Nextflow-driven S3 operations, not just the two sentinel processes. On workflows where many tasks finish near-simultaneously and publish outputs, those copies will now serialize through 10 connections instead of 1000. Manual verification on the 100-group batch implies this is acceptable in practice, but a one-line addition to the PR description acknowledging publish-time S3 traffic as another consumer of the capped pool would give reviewers with operational context the full picture.

configs/profiles.config
```diff
@@ -7,7 +7,14 @@ manifest {
...
 docker.enabled = true
 wave.enabled = true
-aws.client.maxConnections = 1000
+// Cap the head-node S3 SDK connection pool. Without this cap, two head-process
+// S3 bursts can exceed S3's per-prefix request-rate threshold: HashBuilder's
+// startup cache verification on `-resume` against thousands of cached work-dir
+// paths, and the per-group WRITE_SENTINEL_DOWNSTREAM fan-out at end-of-workflow.
+// Either burst can trigger the 503 -> SDK mark/reset retry -> operator-fork-
+// semaphore exhaustion cascade. The value 10 was chosen empirically from the
+// failing workflow run in issue #801, not derived from a documented S3 limit.
```
dp-rice: I don't know how you feel about this sort of long explanatory comment. I could go either way.

willbradshaw: I dislike them and would like to make Claude stop it

configs/resources.config
```diff
@@ -53,6 +53,19 @@ process {
...
+    // Head-node sentinel processes (WRITE_SENTINEL_RUN, WRITE_SENTINEL_DOWNSTREAM)
+    // run with `executor 'local'` and validate published outputs by polling many
+    // S3 paths via SentinelUtils.waitForFiles. The cap binds in practice on the
+    // per-group WRITE_SENTINEL_DOWNSTREAM fan-out (WRITE_SENTINEL_RUN only fires
+    // once per workflow), preventing the burst of HEAD requests from exceeding
+    // S3's per-prefix request-rate threshold. Belt-and-suspenders alongside the
+    // SDK-level `aws.client.maxConnections` cap in configs/profiles.config; this
+    // also pins the protection at the sentinel layer so future tuning of the
+    // SDK pool does not silently undo it. See issue #801.
+    withLabel: sentinel {
```
dp-rice, line 57: I went back and forth on whether this is the right label and decided that it's fine. The property we want to guard against is: "Lots of parallel calls to S3 from the head node" but as far as I can tell this is the only type of process that does this and the only one that we're likely to write.

modules/local/writeSentinelRun/main.nf
```diff
@@ -4,6 +4,7 @@
 // Shared regex/poll/timestamp helpers live in lib/SentinelUtils.groovy.
 process WRITE_SENTINEL_RUN {
     executor 'local'
+    label "sentinel"
```
dp-rice, line 7: Since this process doesn't actually fan out (it just runs once per delivery), the `maxFork` cap doesn't actually do anything. We could drop the label here if you want.

configs/profiles.config
```diff
@@ -7,7 +7,11 @@ manifest {
 // Currently universal settings
 docker.enabled = true
 wave.enabled = true
-aws.client.maxConnections = 1000
+// Cap head-node S3 SDK connection pool below S3's per-prefix request-rate
+// threshold. Head-process bursts (HashBuilder cache verification on -resume;
+// WRITE_SENTINEL_DOWNSTREAM fan-out) can otherwise trigger 503 cascades.
+// Value chosen empirically from #801, not derived from a documented S3 limit.
+aws.client.maxConnections = 10
```
willbradshaw: Has any work gone into seeing what the ceiling on this is? I'm nervous about cutting this globally by 2 OOM without testing. Could we put it in the `withLabel: sentinel` statement?

willbradshaw:
@dp-rice one question, everything else lgtm

configs/profiles.config
dp-rice: > Could we put it in the withLabel: sentinel statement?

Unfortunately not. It's a global config. It also would not help with the `-resume` problem.

I only have indirect evidence on the ceiling:

- I am *pretty* sure that the there is only one concurrent S3 call from the head node per sentinel process.
- The 100-group batch failed, suggesting that 100 connections is too many.
- We regularly have ~30 groups (including some of the backlog batches) and these never hit this bug, (either the sentinel or the cache verification version), suggesting that 30 connections is safe.
- 50 appears to be the nextflow default (according to Claude).

I propose setting it to 50.

willbradshaw: Sounds good, let's do it and see if we run into any issues

dp-rice: I also spent a few minutes with Claude trying to elicit the 503 errors without running a giant delivery through the pipeline and couldn't.

configs/profiles.config
dp-rice: reduced to one line each

willbradshaw (approved):

PR #819, opened 2026-05-22
Refresh index inputs: Kraken2 DB, phage 38018 exclude, broaden host overrides

willbradshaw:
Refresh the inputs to the next INDEX rebuild. No schema or output-layout changes; the next-built index will differ in content (different Kraken2 DB, one additional hard-excluded taxon, broader propagation of the existing host-infection overrides).

**Changes:**
- `configs/index.config`: Kraken2 standard DB → `k2_standard_20260226.tar.gz` (was `20251015`) — refresh to the current upstream genome-idx build.
- `configs/index.config`, `configs/index-for-run-test.config`, `tests/configs/index.config`: add taxid `38018` ("Bacteriophage sp.") to `viral_taxids_exclude_hard` — a non-Caudoviricetes phage catch-all that we want hard-excluded from the reference virus DB.
- `ref/host-infection-overrides.json`: extend the four existing overrides (WNV, LCMV, Puumala, Banzi) from `["human", "vertebrate"]` to `["human", "primate", "mammal", "vertebrate"]` so the MATCH override also propagates through `infection_status_primate` and `infection_status_mammal`, not just the species-root pair.


willbradshaw:
Regarding the primate/mammal updates: it's true we don't currently use these, and possibly we should just cut them from the index entirely at some point. However, until we do that, we should treat them consistently to other host taxa; it doesn't make sense for a virus to be marked as human- and vertebrate-infecting but not primate-infecting, which was the case before this PR.

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # v3.2.2.0-dev
 
+- Refresh index inputs: bump Kraken2 standard DB to `k2_standard_20260226.tar.gz` (from `20251015`); add taxid `38018` ("Bacteriophage sp.") to `viral_taxids_exclude_hard` in all three INDEX configs (`configs/index.config`, `configs/index-for-run-test.config`, `tests/configs/index.config`); expand the four existing entries in `ref/host-infection-overrides.json` (WNV, LCMV, Puumala, Banzi) from `["human", "vertebrate"]` to `["human", "primate", "mammal", "vertebrate"]` so the override propagates through intermediate host groups, not just the species-level and root-level groups. Together these change the annotated virus DB and Kraken2 DB on next index rebuild; no schema or output-layout changes.
```
chrisdoering8197, line 3: nit: Since the move from `["human", "vertebrate"]` to `["human", "primate", "mammal", "vertebrate"]` for the hard included species occurred internal to this version do you want to describe it as a change or instead leave out the intermediate state?

chrisdoering8197 (approved):
LGTM! Just one minor comment, take it or leave it.

CHANGELOG.md
willbradshaw: Good point. I'll leave it in for now but will clean it up before merging to main.

willbradshaw:
@chrisdoering8197 can you try approving again? I needed to up your permissions so that the approval goes through.

chrisdoering8197 (approved):

chrisdoering8197:
@willbradshaw just re-approved. Sorry for jumping the gun while it was still a draft!

PR #831, opened 2026-06-12
CI: remove unnecessary disk-cleanup step from setup-nf-test

jo-delia-faraguna:
Removes the `Maximize build space` disk-cleanup step from the `setup-nf-test` composite action. The step was added in #556 (Dec 2025) to fix a `no space on device` failure in the index test, but measurement shows the CI runners now have ample headroom and the step is pure overhead.

**Measured disk usage** (instrumented `df -h` before cleanup, after cleanup, and after tests across all nf-test jobs):

| | `ubuntu-latest` (index/run/downstream) | `ubuntu-16` large runner (modules/subworkflows) |
|---|---|---|
| Total disk | 145 G | 581 G |
| **Free before cleanup** | **89 G** (39% used) | **528 G** (10% used) |
| Free after cleanup | 112 G (24% used) | 550 G (6% used) |
| Reclaimed by cleanup | ~22 G | ~22 G |
| Free after tests (with cleanup) | 101–108 G | 537–538 G |
| **Cleanup step duration** | **~4m54s** | **~4m58s** |

Peak test footprint is small (after-cleanup → after-test used): index +7 G, run +11 G, downstream +3 G, modules/subworkflows +13 G. Even the heaviest job (`run`) would finish with **~78 G free without any cleanup**; the large runner never drops below ~515 G free.

**Conclusion:** the step costs ~5 min per job across 5 nf-test jobs on every PR to reclaim ~22 GB that nothing uses. Removing it is safe given the headroom. If a `no space` regression ever recurs (e.g. a runner-images change), re-adding is a one-line revert.

**Changes:**
- Remove the `Maximize build space` step from `.github/actions/setup-nf-test/action.yml`.
- Changelog entry under `v3.2.2.0-dev`.

(The branch history includes a temporary `df -h` instrumentation commit used to gather the numbers above; it is reverted in the final diff, which only removes the cleanup step.)

### DIFF
--- a/.github/actions/setup-nf-test/action.yml
+++ b/.github/actions/setup-nf-test/action.yml
@@ -20,6 +20,8 @@ runs:
     - name: Maximize build space
       shell: bash
       run: |
+        echo "=== Disk usage BEFORE maximizing build space ==="
+        df -h
         sudo rm -rf /usr/share/dotnet
         sudo rm -rf /usr/local/lib/android
         sudo rm -rf /opt/ghc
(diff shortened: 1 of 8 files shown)


willbradshaw (approved):
I'm nervous about this, but we can try it

PR #840, opened 2026-06-15
Fix CI doc file detection

evanfields:
Add `predicate-quantifier: every` to the `dorny/paths-filter` step in `.github/workflows/check-changelog.yml` so the `!**/*.md` / `!docs/**` negation patterns are actually applied; under the default `some` quantifier every file matched `**`, so the detector never recognized a docs-only change.


jo-delia-faraguna (approved):
LGTM, thanks.

PR #844, opened 2026-06-18
Make COUNT_READS_PER_CLADE scale with data, not taxonomy size

jdf:
In `COUNT_READS_PER_CLADE`, clade counts were computed by a full depth-first traversal of the *entire* (~2.5M-node) NCBI taxonomy (twice), plus a third full-tree pass to write output, so cost was pinned to the taxonomy size rather than to how much data the sample actually contained.

This rewrites the counting to propagate each directly-assigned taxid's counts **upward** through its ancestor chain. The work — and the peak memory — now scales with the number of taxids the sample actually hit (`O(observed taxids × tree depth)`) instead of the size of the taxonomy. Outputs are byte-for-byte identical to the previous implementation.

**Changes:**
- **Upward propagation instead of full-tree DFS** (`accumulate_clade_counts`): for each directly-assigned taxid, walk its parent chain to the root and add its counts to every ancestor, building both the total and dedup clade counters in a single pass. Only nodes on a path from an observed taxid to a root are ever touched, so the millions of zero-count nodes that previously inflated memory (the OOM driver) are never materialized.
- **Flat `child→parent` map** (`build_parent_map`) replaces the parent→children `defaultdict(set)` tree. Ancestry is reconstructed on demand by following parent pointers — no precomputed/serialized ancestor index needed. Retains the duplicate-child validation.
- **Lazy, per-lineage cycle detection** in `ancestors()` replaces the eager recursive whole-taxonomy `detect_cycle` scan. Cycles are detected on the lineages the sample actually touches. (Behavioral note for reviewers: cycles in untouched parts of a self-built index artifact are no longer scanned for; flagged as a conscious trade-off.)
- **Sparse-tree output**: rows are emitted via a sorted pre-order DFS over only the touched nodes, which reproduces the previous row ordering exactly. Verified byte-for-byte against `test-data/toy-data/count-reads-per-clade/expected-output.tsv`.
- **Off-tree reads** (assigned to taxids absent from the taxonomy) are still dropped — matching the documented module behavior — but now emit a stderr warning with a count instead of being silently discarded.
- **Tests**: rewrote the module unit tests to the new API with equivalent coverage (duplicate-child, cycle, off-tree drop, the clade-count math, and the integration cases).

**Validation:** unit tests (17/17), `ruff`, and `mypy` all pass locally; output verified byte-identical on the toy fixture. The nf-test toolchain isn't available locally — the `COUNT_READS_PER_CLADE` nf-test in CI is the authoritative check on the unchanged-output claim.

**Versioning:** Point-level change (performance only; no results, schema, or index/pipeline-compatibility change), which sits below the Results-level bump already in `3.2.2.0-dev`, so the version is unchanged and only a CHANGELOG entry was added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # v3.2.2.0-dev
 
+- Speed up `COUNT_READS_PER_CLADE` and eliminate its peak-memory OOM kills by propagating each directly-assigned taxid's counts upward through its ancestor chain, instead of doing a full depth-first traversal of the entire (~2.5M-node) NCBI taxonomy; work now scales with the number of taxids the sample actually hit. Outputs are byte-for-byte identical to the previous implementation (no schema or output-format change). Off-tree reads (assigned to taxids absent from the taxonomy) are still dropped, but now emit a stderr warning with a count instead of being silently discarded.
 - Add `set -euo pipefail` to `BLASTN` module.
 - Add `pigz` to the `python` container so Python processes can use parallel (de)compression.
 - CI: Fix the CHANGELOG CI check wrongly failing docs-only PRs that don't update CHANGELOG.md.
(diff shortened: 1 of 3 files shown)


jdf:
Claude-generated code reviewed by me before marking ready per policy.

jdf:
@jeffkaufman FYI

Issue #841, opened 2026-06-16
Bracken output is empty due to Kraken2 standard database taxonomy change breaking domain-level classification


joshkome:
## Description
The pipeline produces empty Bracken output when using Kraken2 standard databases from after December 2024. This is caused by a change in the taxonomy hierarchy between databases that removes the `D` rank for Eukaryota. In Kraken databases since 2025, Eukaryota is now `R2`. 

Bracken is called with `-l D` from a `classification_level` parameter set in `subworkflows/local/profile/main.nf`, based on the assumption that Eukaryota sits at rank `D` and that Homo sapiens is the only eukaryote in the Standard database. Since the Kraken database URL in `configs/index.config` is from 10-15-2025, no reads are assigned rank `D` and Bracken skips. 

## Recommended: 
- Pin an older Kraken2 database URL in `configs/index.config` or update default classification level assigned in `subworkflows/local/profile/main.nf` to `R2`.
- Make the Bracken classification level configurable via `params.bracken_level` in `configs/run.config` and make version compatibility clear

PR #846, opened 2026-06-23
Replace bracken for domain count estimates

evanfields:
Kraken2 standard databases built after December 2024 reorganized the taxonomy so that Eukaryota no longer sits at rank code `D` (it's now `R2`). Because Bracken was invoked with `-l D`, on these newer databases no reads land at the requested level and Bracken silently skips, producing empty `{sample}_bracken.tsv.gz` outputs. This PR drops the Bracken dependency entirely and replaces it with a small Kraken2-report summary that estimates per-domain abundance directly, keyed on stable NCBI taxids rather than the fragile rank codes — so it's immune to the taxonomy-version shift that broke the old path.

We weren't really doing much with Bracken: just getting domain level estimates, which doesn't much use Bracken's clever Bayesian logic. And empirically we relied on the domains we care about being at a matching rank code in Kraken's database.

The new `KRAKEN_DOMAIN_SUMMARY` module identifies the four domains (Bacteria, Archaea, Eukaryota, Viruses) by taxid and reports their read counts in the same filename and column layout as before, for drop-in compatibility. `added_reads` redistributes reads classified *above* the domain rows back down the tree in two levels: reads stuck at "cellular organisms" (taxid 131567) are prorated among the three cellular domains only, and remaining root-level reads across all four — both proportional to Kraken clade counts. This two-level heuristic respects the taxonomy and avoids over-assigning the rare Viruses domain; unclassified reads stay excluded. The same table is also published under an honest name, `experimental/{sample}_kraken_domains.tsv.gz`, as a migration target away from the legacy `bracken` filename (experimental outputs carry no stability guarantees).

**Changes:**
- Remove the `BRACKEN` module, its container (`bracken_plus_utils.yml`), and the `classification_level` plumbing; add `KRAKEN_DOMAIN_SUMMARY` with `kraken_domain_summary.py` and its pytest suite.
- Wire the new module into the `profile`/`taxonomy` subworkflows and `run` workflow; update toy test data and nf-tests accordingly.
- Update docs (`run.md`, `output.md`, `downstream.md`, `config.md`) and the `bracken.schema.json` to reflect the Kraken-derived source and the new experimental output.

## Validation against historical Bracken output

To confirm the Kraken-derived table reproduces what (non-broken) Bracken produced, I
validated it against real historical production data from runs built on a
pre-2024-taxonomy Kraken2 database — the regime where Bracken still emitted correct,
non-empty domain estimates, so each run carries both the Kraken report (the heuristic's
input) and the Bracken output (ground truth).

Across the comparison the heuristic's Kraken clade counts match Bracken's exactly, so the
only difference is how each method redistributes the small above-domain residual. The
heuristic reproduces Bracken's estimates closely for every domain, and on the rare,
business-critical **Viruses** domain it tracks Bracken with median relative error ~0% and
no systematic bias.

Representative spread of samples (per-sample non-ribosomal profiling read set; names
obfuscated). **K** = Kraken-assigned clade reads (the identical input to both methods),
**Bk** = Bracken estimate, **H** = new heuristic estimate:

| Sample | Bacteria K | Bacteria Bk | Bacteria H | Eukaryota K | Eukaryota Bk | Eukaryota H | Viruses K | Viruses Bk | Viruses H |
|---|---|---|---|---|---|---|---|---|---|
| Sample 1 | 15679 | 15763 | 15779 | 230 | 239 | 231 | 58 | 58 | 58 |
| Sample 2 | 75948 | 76251 | 76255 | 887 | 901 | 889 | 7244 | 7254 | 7259 |
| Sample 3 | 75023 | 76211 | 76165 | 928 | 947 | 941 | 16503 | 16605 | 16652 |
| Sample 4 | 52715 | 53315 | 53371 | 826 | 917 | 836 | 18367 | 18371 | 18393 |
| Sample 5 | 74141 | 74350 | 74333 | 1179 | 1191 | 1181 | 35102 | 35104 | 35130 |
| Sample 6 | 66717 | 67490 | 67491 | 1203 | 1233 | 1215 | 44418 | 44554 | 44565 |
| Sample 7 | 38827 | 38908 | 38967 | 5891 | 5985 | 5912 | 39350 | 39351 | 39365 |
| Sample 8 | 45772 | 46077 | 46013 | 588 | 600 | 590 | 65373 | 65379 | 65452 |

Addresses #841 / Linear COMP-2514


modules/local/krakenDomainSummary/resources/usr/bin/kraken_domain_summary.py
evanfields, line 1: This file is extremely verbose. Look for ways to simplify, keeping in mind this file exists just for one Nextflow process with tightly scoped inputs and outputs.

modules/local/krakenDomainSummary/resources/usr/bin/kraken_domain_summary.py
```diff
@@ -0,0 +1,281 @@
...
+
+
+@dataclass(frozen=True)
+class KrakenRow:
+    """The fields we use from one Kraken2 report row."""
+
+    reads_clade: int
+    rank: str
+    taxid: int
+    name: str
```
evanfields, line 86: Do we ever use name or rank? If not, can we just simplify the file parsing to be a simple dict[int, int] of `{taxid: count}`?

modules/local/krakenDomainSummary/resources/usr/bin/kraken_domain_summary.py
```diff
@@ -0,0 +1,281 @@
...
+
+
+####################
+# HELPER FUNCTIONS #
+####################
+
+
+def open_by_suffix(filename: str | Path, mode: str = "r") -> IO[str]:
```
evanfields: Why do we need to handle multiple suffixes? Don't we know exactly the file type we're calling with in the single place in the pipeline we invoke this script?

modules/local/krakenDomainSummary/resources/usr/bin/kraken_domain_summary.py
```diff
@@ -0,0 +1,281 @@
...
+    rows: list[KrakenRow] = []
+    with open_by_suffix(report_path) as report_file:
+        for line in report_file:
+            fields = line.rstrip("\n").split("\t")
+            # Skip blank lines and a header row (which starts with "pc_reads_total").
+            if len(fields) < KRAKEN_FIELD_COUNT or fields[0] == "pc_reads_total":
+                continue
+            rows.append(
+                KrakenRow(int(fields[1]), fields[5], int(fields[6]), fields[7].strip())
+            )
```
evanfields: We only use a few taxids, right, don't need to store everything?

modules/local/krakenDomainSummary/resources/usr/bin/kraken_domain_summary.py
```diff
@@ -0,0 +1,281 @@
...
+    return rows
+
+
+def prorate(residual: int, weights: list[int]) -> list[int]:
```
evanfields: I wonder if this can be inlined, especially if we trust that kraken output is not going to give us counts < 0

modules/local/krakenDomainSummary/resources/usr/bin/kraken_domain_summary.py
```diff
@@ -0,0 +1,281 @@
...
+
+    added = {row.taxid: 0 for row in domain_rows}
+
+    def add_prorated(target_rows: list[KrakenRow], residual: int) -> None:
+        """Prorate `residual` across `target_rows` by clade count and bank it."""
+        weights = [row.reads_clade for row in target_rows]
+        for row, share in zip(target_rows, prorate(residual, weights), strict=True):
+            added[row.taxid] += share
+
+    # Level 1: residual inside "cellular organisms" -> cellular domains only.
```
evanfields: Look for simplifications in this block

modules/local/krakenDomainSummary/resources/usr/bin/kraken_domain_summary.py
```diff
@@ -0,0 +1,281 @@
...
+    """
+    with open_by_suffix(output_path, "w") as output_file:
+        if not rows:
+            return
+        output_file.write("\t".join(OUTPUT_FIELDS) + "\n")
+        for row in rows:
+            output_file.write("\t".join(row) + "\n")
+
+
+def create_domain_summary(report_path: str | Path, output_path: str | Path) -> None:
```
evanfields: Since we only ever call this in main, we can just inline?

CHANGELOG.md
evanfields, line 1: Try to shorten the changelog entry here. The details are good for something like the PR description, but a normal reader who wants to see what changed doesn't need this level of detail.

main.nf
```diff
@@ -90,7 +90,10 @@ output {
         tags nextflow_file_class: "publish", "nextflow.io/temporary": "false"
     }
     experimental_run {
-        path "experimental"
+        // Publish the RUN domain-abundance table (emitted as profile_ch.bracken,
+        // a (sample, file) tuple) under an honest name without a copy process, by
+        // renaming on publish with the `>>` operator.
```
evanfields: This comment is confusing to a future reader who doesn't have our collaborative context, or remember us researching nextflow file renaming conventions. "Honest name"? Be shorter and more direct.

workflows/run.nf
```diff
@@ -50,6 +50,10 @@ workflow RUN {
         reads_trimmed_viral = viral_ch.kmer_trimmed
         qc_results_run = qc_results_ch
         other_results_run = other_results_ch
-        experimental_run = channel.empty()
+        // Bridge: additionally publish the domain-abundance table under an honest
+        // name in experimental/ (renamed on publish, no copy — see the experimental_run
+        // output block in main.nf) as the migration target away from the legacy
+        // "bracken" filename. Identical content to results/{sample}_bracken.tsv.gz.
+        experimental_run = profile_ch.bracken
```
evanfields, line 55: The comment here might be a tad confusing, "bridge"?

tests/modules/local/krakenDomainSummary/main.nf.test
```diff
@@ -0,0 +1,42 @@
...
+                // (taxid 131567) and at the root, so it exercises the full
+                // taxid-keyed two-level proration in-container.
+                '''
+                input[0] = Channel.of("test")
+                    | combine(Channel.of("${projectDir}/test-data/toy-data/kraken-domain-summary/kraken.report"))
+                '''
+            }
+        }
+        then {
+            // Should run without failures
```
evanfields: This seems like a no-value-add comment

tests/modules/local/krakenDomainSummary/main.nf.test
```diff
@@ -0,0 +1,42 @@
...
+
+    test("Should derive a two-level domain summary, robust to non-D rank codes") {
+        tag "expect_success"
+        when {
+            params {}
+            process {
+                // Toy report uses post-2024 taxonomy rank codes (R1/R2, no "D")
+                // and has reads above the domains both at "cellular organisms"
+                // (taxid 131567) and at the root, so it exercises the full
+                // taxid-keyed two-level proration in-container.
```
evanfields: good comment

tests/modules/local/krakenDomainSummary/main.nf.test
```diff
@@ -0,0 +1,42 @@
...
+            def rows = path(process.out.output[0][1]).linesGzip
+            // The "cellular organisms" residual goes only to the cellular domains;
+            // Viruses receives just its share of the small root-level residual
+            // (20), not the cellular residual it would get under naive proration.
+            assert rows == [
+                "name\ttaxid\trank\tkraken2_assigned_reads\tadded_reads\tnew_est_reads\tfraction_total_reads",
+                "Bacteria\t2\tR2\t400\t80\t480\t0.33803",
+                "Archaea\t2157\tR2\t400\t80\t480\t0.33803",
+                "Eukaryota\t2759\tR2\t200\t40\t240\t0.16901",
+                "Viruses\t10239\tR1\t200\t20\t220\t0.15493",
```
evanfields: I wonder if we need to test correctness at all in this `main.nf.test`. Thoughts? In general can we use the faster and more readable pytest tests for correctness, here just test wiring?

schemas/bracken.schema.json
```diff
@@ -52,7 +52,7 @@
     {
       "name": "added_reads",
       "title": "Added Reads",
-      "description": "Number of reads redistributed to this taxon by Bracken.",
+      "description": "Number of reads classified above the domain rows in the Kraken2 report and redistributed to this domain. Reads at 'cellular organisms' (taxid 131567) are prorated among the cellular domains (Bacteria, Archaea, Eukaryota) only; the remaining root-level reads are prorated across all four domains. Proration is proportional to Kraken2 domain clade counts, so the rare Viruses domain is not over-assigned cellular reads. Unclassified reads are not redistributed.",
```
evanfields: I'm unsure if we need this level of detail here, thinking probably not as the heuristic details are an implementation detail which we might change without changing the schema?

modules/local/krakenDomainSummary/resources/usr/bin/test_kraken_domain_summary.py
```diff
@@ -0,0 +1,218 @@
...
+    report = tmp_path / "kraken.report"
+    output = tmp_path / "domain.tsv"
+    write_report(report, report_rows)
+
+    kraken_domain_summary.create_domain_summary(report, output)
+
+    assert output.read_text() == ""
+
+
+def test_create_domain_summary_reads_gzipped_report(tmp_path: Path) -> None:
```
evanfields: See comment in kraken_domain_summary.py about handling both gzip and regular input

modules/local/krakenDomainSummary/resources/usr/bin/test_kraken_domain_summary.py
```diff
@@ -0,0 +1,218 @@
...
+    assert read_output(output) == [
+        "name\ttaxid\trank\tkraken2_assigned_reads\tadded_reads\tnew_est_reads\tfraction_total_reads",
+        "Bacteria\t2\tD\t600\t60\t660\t0.66000",
+        "Archaea\t2157\tD\t200\t20\t220\t0.22000",
+        "Eukaryota\t2759\tD\t100\t10\t110\t0.11000",
+        "Viruses\t10239\tD\t10\t0\t10\t0.01000",
+    ]
+
+
+def test_create_domain_summary_splits_residual_across_both_levels(
```
evanfields: Seems like this and the previous two tests could be combined into one correctness test?
For happy path can we not just have two test cases, one with and one without some reads assigned to "cellular organisms", and jointly they cover all well formed input?

test-data/toy-data/kraken-domain-summary/kraken.report.gz
willbradshaw, line 1: We should try to avoid adding binary files to the repo if we can avoid it; can the script not handle plaintext Kraken reports? If not, we should add that feature (can copy verbatim from another Python module, lots of them have it)

main.nf
```diff
@@ -90,7 +90,9 @@ output {
         tags nextflow_file_class: "publish", "nextflow.io/temporary": "false"
     }
     experimental_run {
-        path "experimental"
+        // Publish the domain-abundance table under a name that reflects its
+        // Kraken2 source, renamed in place with `>>` (no copy process).
+        path { sample, file -> file >> "experimental/${sample}_kraken_domains.tsv.gz" }
```
willbradshaw, line 95: Whoa what is this construct? Does it pass linting?

workflows/run.nf
```diff
@@ -50,6 +50,10 @@ workflow RUN {
         reads_trimmed_viral = viral_ch.kmer_trimmed
         qc_results_run = qc_results_ch
         other_results_run = other_results_ch
-        experimental_run = channel.empty()
+        // Also publish the domain-abundance table in experimental/ under a name
```
willbradshaw: This is a very long comment for a very simple action

modules/local/krakenDomainSummary/resources/usr/bin/kraken_domain_summary.py
```diff
@@ -0,0 +1,219 @@
...
+
+def parse_arguments() -> argparse.Namespace:
+    """
+    Parse command-line arguments.
+
+    Returns:
+        Parsed command-line arguments.
+    """
+    parser = argparse.ArgumentParser(description=DESC)
+    parser.add_argument("report", help="Gzipped Kraken2 report.")
```
willbradshaw: Suggest giving this the ability to handle plaintext

subworkflows/local/taxonomy/main.nf
```diff
@@ -34,10 +33,9 @@ workflow TAXONOMY {
         kraken_headers = "pc_reads_total,n_reads_clade,n_reads_direct,n_minimizers_total,n_minimizers_distinct,rank,taxid,name"
         kraken_head_ch = HEAD_KRAKEN_REPORTS(kraken_ch.report, kraken_headers, "kraken_report")
         kraken_label_ch = LABEL_KRAKEN_REPORTS(kraken_head_ch.output, "sample", "kraken_report")
-        // Run Bracken and munge reports
-        bracken_ch = BRACKEN(kraken_ch.report, kraken_db_ch, params_map.classification_level, params_map.bracken_threshold) // NB: Not streamed
-        bracken_rehead_ch = REHEAD_BRACKEN(bracken_ch, "taxonomy_id,taxonomy_lvl,kraken_assigned_reads", "taxid,rank,kraken2_assigned_reads")
-        bracken_label_ch = LABEL_BRACKEN(bracken_rehead_ch.output, "sample", "bracken")
+        // Derive Bracken-shaped domain abundance reports directly from Kraken
+        bracken_ch = KRAKEN_DOMAIN_SUMMARY(kraken_ch.report)
+        bracken_label_ch = LABEL_BRACKEN(bracken_ch.output, "sample", "bracken")
```
willbradshaw: In the spirit of reducing data passes, we could drop this and fold its functionality into KRAKEN_DOMAIN_SUMMARY.

modules/local/krakenDomainSummary/resources/usr/bin/kraken_domain_summary.py
```diff
@@ -0,0 +1,219 @@
...
+
+    clade = {taxid: row.clade for taxid, row in rows.items()}
+    added = dict.fromkeys(domains, 0)
+
+    def prorate(taxids: list[int], residual: int) -> None:
+        """Split `residual` across `taxids` by clade count, banking into `added`."""
+        total = sum(clade[taxid] for taxid in taxids)
+        for taxid in taxids:
+            if total:
+                added[taxid] += residual * clade[taxid] // total
```
willbradshaw: Won't a naive floor-divide here occasionally drop counts? Not a big deal for abundant taxa, but could be meaningful for small ones.

willbradshaw (changes requested):
Looks good, minor changes only. The main one is that I would like to avoid adding new binary files to the repo if I can, so the script should handle plaintext input. (This is good practice anyway.)

PR #861, opened 2026-06-28
Update Nucleaze version to 1.5 alpha

evanfields:
We previously pinned an old commit of Nucleaze, which had a risk of some transient OOM or stack overflow errors. A recent Nucleaze commit fixes those, so update Nucleaze.

**NB**: the Nucleaze kmer index format has changed, so we'll need to update the index we're planning for MGS Workflow 3.2.2.


willbradshaw (approved):

evanfields:
Evan-guided Claude-produced report on whether the new Nucleaze commit is safe for us:

# Production suitability memo — nucleaze stack/OOM fix (merged as `4090fe3`)

**Question.** Is the stack/OOM-fix work now on `main` (tip
`4090fe39757d0011abdc73ba500537ba2e45cf7a`, "`4090fe3`") safe to deploy into the
production pipeline that previously pinned `333912e` (v1.4.2)? The bar for this
upgrade is narrow: it must **fix the production OOM / stack-overflow crashes** and
must **not introduce new bugs or performance regressions**. It does *not* have to
fix every pre-existing rough edge.

**Verdict: GO.** The change fixes both crash modes (one structurally, one provably),
produces **byte-identical output** to the current production binary on prod-scale
data, is **faster** when matching-bound (never slower), and requires **no change to
how the pipeline invokes nucleaze**. Residual risks are pre-existing (not
regressions) and do not affect the reported short-read incident. Details and caveats
below.

Evidence: source audit, the Rust test suite, and A/B runs of freshly-built `333912e`
vs the candidate binaries on a prod-like comparison harness (BBDuk vs nucleaze,
synthetic interleaved paired datasets). All timing and stack-overflow-reproduction
figures in this memo were measured on a 14-core M4 MacBook (a laptop), **not**
production hardware — which matters: the overflow needs prod-scale concurrency to
reach the real 12 MB stack (§2), and the throughput win is bottleneck-dependent (§4).
The candidate was evaluated at the pre-merge branch tip `046e124`; that tree was merged to `main` as `4090fe3` and is verified
**byte-identical** to `046e124` (`git diff 046e124 4090fe3` is empty), so every
result below transfers unchanged. No nucleaze source was modified to produce this
memo (the working tree was briefly checked out to older commits only to build
comparison binaries, then restored).

---

## 1. What is being shipped

`333912e` is an ancestor of the prior `main` (`2208f36`), which is an ancestor of the
new tip `4090fe3`. So the upgrade bundles **two** deltas, and both reach production
for the first time:

| Delta | Content |
|---|---|
| `333912e` → `2208f36` (prior `main`) | bloom/approx feature work: `KmerStore` trait refactor, `MiniBloom` blocked bloom filter + `--fpr` mode, indexing speedups. Was already on `main`, but **new to production**. |
| `2208f36` → `4090fe3` (new `main`) | The crash fix (`98d7116` "Bound in-flight chunks and de-nest per-read matching") plus rustfmt/comment commits, then a final squashed commit `4090fe3` folding a test-only fix and a bloom-only cleanup. |

The crash fix proper is `98d7116`. Everything after it is cosmetic / test-only /
bloom-only:

- rustfmt and comment-wording commits — no behavior change.
- a fix to the `test_metadata_kmer_functional` integration test (test file only; see §5).
- removal of an unused in-band `k` sentinel from **bloom** index construction (touches
  only the `--fpr` path).

**Consequence:** the exact-mode (production) *runtime* behavior of `4090fe3` is
identical to the verified `98d7116`. None of the follow-ups touch the exact-mode
matching or I/O path.

---

## 2. Does it fix the OOM / stack overflow? — Yes

Both production symptoms share one root cause in `333912e`: **no backpressure on
chunk creation**. The reader fills ~5 MB arenas and dispatches them with
fire-and-forget `rayon::spawn` into Rayon's *unbounded* global injector; the only
bounded channel sits downstream of it. When the writer falls behind, the backlog
grows without limit (→ OOM), and that same backlog is the breadth that lets a
join-blocked Rayon worker stack other chunks' nested `par_iter().collect()` on its
own frame (→ stack overflow). The OOM is provable directly from the source; the
nested-stack mechanism was confirmed against the pinned `rayon-core 1.13.0`
(`spawn` injects globally; a `join`-blocked worker runs found injected jobs on its
current stack).

The candidate fixes both, and I verified the implementation in `src/core.rs`:

1. **Stack overflow — removed structurally (de-nesting).** The per-chunk match loop
   is now sequential — `local_offsets.iter().map(...).collect()` (`core.rs:700-707`),
   not `par_iter`. A spawned chunk task therefore contains no `join`/`collect`, so a
   worker running it never blocks-and-steals another chunk onto its stack. Matching
   stack depth is now **O(1)** regardless of concurrency or backlog. Parallelism is
   retained at the chunk level (many spawned tasks across the pool).

2. **OOM — bounded by a creation-to-write permit.** A `PermitPool` (a pre-filled
   crossbeam channel used as a counting semaphore, `core.rs:42-65`) is sized
   `current_num_threads() + 24` (`core.rs:652`). A permit is acquired **before the
   arena is filled** (`core.rs:747`, re-acquired at `:772`/`:809`) and is carried
   inside the `SequenceChunk` as an **RAII guard** (`_permit`, `core.rs:92`) that is
   dropped only when the chunk is dropped — i.e. **after it is written**. Total
   live arenas (`reader-current + in-flight + output channel + reorder heap`) ≤
   permit count, so **peak memory is flat in input size**.

The permit is released on *write* (via the guard) rather than on match-completion or
channel-receipt, which is what also bounds the `--order` reorder heap — the buffer
that channel capacity alone cannot bound. And since the permit count (≈ tens) is far
below `MAX_BUFFERED_CHUNKS = 1000` (`core.rs:1010`), the "too many out-of-order
chunks" hard-abort can no longer be reached. Ordered mode cannot deadlock: chunk ids
are assigned in creation order (`core.rs:692`), so the in-order chunk the writer
waits for was created before anything now parked in the heap and is already somewhere
in the pipeline holding its permit.

**Empirical confirmation of the stack fix.** A fault-injection reproduction (the
real 12 MB stack does not overflow on the 14-core M4 MacBook used here — it needs
prod-scale vCPU counts — so the Rayon stack was lowered to amplify the *same* nested-stealing
mechanism, identically on both binaries) shows the overflow on the pre-fix code and
its elimination by the fix:

| Rayon stack, `threads=14`, same input | pre-fix (`2208f36`) | fix (`98d7116`) |
|---|---|---|
| 16 KB | **15/15 overflow** | 0/15 |
| 24–48 KB | overflow (intermittent, mirrors prod) | 0/N |
| `threads=1`, any size | never overflows (no work-stealing) | never |

The overflow appears only with concurrency — the signature of the work-stealing
path — and the fix never overflowed across 60+ runs.

**Honest caveat (does not change the verdict):** there is still no production
`RUST_BACKTRACE` confirming the *specific* prod stack crash is this exact frame
rather than another large frame. But the OOM is source-proven, the stack mechanism
is source-proven and reproduced, and the fix removes the mechanism and bounds memory
*regardless* of which frame the prod crash was — so it is a strict improvement with
no downside on that axis.

---

## 3. Regression analysis — no new bugs in the production path

### 3a. Exact-mode classification is byte-identical to current production

The exact-mode matcher is the production path (`333912e` is exact-mode only; the
pipeline does not pass `--fpr`). Two independent lines of evidence:

**Source.** `git diff 333912e..4090fe3 -- src/kmer_ops.rs` shows the matcher was only
*mechanically* refactored into the `KmerStore` trait. `process_read`'s rolling
k-mer / reverse-complement recurrence, the canonical `min(kmer, rc_kmer)`, the
`threshold` early-exit, the 1024-shard layout, the shard hash (`kmer ^ (kmer>>12) &
mask`), and `contains` are unchanged — only renamed (`bit_cap`→`k_cap`,
`ref_kmers`→`index`). No arithmetic changed.

**Direct A/B (the decisive test).** I built fresh release binaries of **`333912e`**
and the **candidate** and ran both on prod-like interleaved datasets in the
production mode (`--interinput --order --canonical`, k=31), then `diff`ed the outputs:

| Dataset | Reads | Old `333912e` | Candidate | Output diff |
|---|---|---|---|---|
| rare-contig (100 bp ref) | 2,000,000 | 180 matched | 180 matched | **byte-identical** (both mates) |
| rare-bacterium (1 Mb ref) | 10,000,000 | 50,354 matched | 50,354 matched | **byte-identical** (both mates) |

Byte-identical *including order* — stronger than a "sorted-equal" bar. On the larger
dataset the 50,354 matched reads = exactly the 25,177 hit pairs in the dataset's
synthetic generation ground truth, so both versions are also correct against truth,
not merely equal to each other.

This proves the **whole** prod delta (bloom/`KmerStore` refactor *and* the crash fix)
introduces zero classification or output change for exact mode.

### 3b. The bloom feature is inert for production

Bloom mode is gated entirely behind `--fpr > 0` (`core.rs:164`). With `--fpr` unset
(the production invocation), `fpr` defaults to 0 and control never enters the bloom
branch — it uses `HashShards` + `process_read` exactly as before. The bloom code, its
`unsafe` blocks, atomics, and the architecture-specific prefetch in `kmer_ops.rs`
cannot execute on the production path. So even though bloom is new and carries the
most novel/`unsafe` code in the delta, it is **opt-in and unreachable** unless the
pipeline starts passing `--fpr`.

### 3c. Determinism / stability

Two candidate `--order` runs on the same input produced **byte-identical** output
(reproducibility holds), consistent with the comparison harness's prior "stability:
10 runs identical" result for both old and new versions.

### 3d. BBDuk agreement (correctness vs the gold standard)

Archived results from the prod-like BBDuk-vs-nucleaze comparison harness:

- v1.4.2 (current prod era): BBDuk agreement **PASS** on all core + scaling scenarios.
- the post-refactor 1.5.0-alpha line: BBDuk agreement **PASS** on all core + scaling.

Combined with §3a (candidate ≡ `333912e`, byte-for-byte), the candidate agrees with
BBDuk transitively *and* by direct prior measurement. I did not re-run the full BBDuk
harness on the exact candidate commit because the byte-identical-to-prod result is a
stronger correctness statement than sorted-equal-to-BBDuk, and `333912e` is itself
already validated against BBDuk.

---

## 4. Performance — no regression (an improvement when matching-bound)

On the prod-scale heavy dataset (rare-bacterium, 3.2 GB, 5M read pairs), `--order`
mode, identical args. To attribute the gain I measured a **three-way** comparison
(median of 3 warm same-session runs), since the prod delta changed two things that
could affect throughput — the de-nesting *and* the bloom-era refactor:

| Binary | matching path | `--threads 4` | `--threads 14` |
|---|---|---|---|
| old `333912e` | static `process_read`, **nested** `par_iter` | 1.52 s | 0.756 s |
| `2208f36` (prior main) | `Arc<dyn Fn>`, **nested** `par_iter` | 1.46 s | 0.748 s |
| candidate (`4090fe3`) | `Arc<dyn Fn>`, **sequential** `iter` | **1.14 s** | 0.731 s |

Attribution at 4 threads (where matching is the bottleneck): old → candidate is
**~1.33×** (not the 1.5× a single noisy first run suggested). The split is:

- **Bloom-era refactor (`333912e`→`2208f36`): ≈ neutral** (1.52 → 1.46 s, within
  noise). The refactor actually *added* a per-read cost — matching now goes through an
  `Arc<dyn Fn>` trait object (non-inlinable) instead of the old direct
  `processor.process_read(seq)` — but other changes offset it, so net throughput is
  unchanged. Not a regression.
- **The crash fix's de-nesting (`2208f36`→candidate): the entire win** (1.46 → 1.14 s,
  ~1.28×). This *is* the work-stealing cost going away — both the per-read split/steal
  atomic bookkeeping of the inner `par_iter`, and the cache-locality loss from a
  worker matching scattered stolen sub-ranges instead of streaming one contiguous
  arena. The permit pool added in the same commit is neutral-to-tiny overhead.

**Caveat — the win is bottleneck-dependent.** At `--threads 14` all three converge
(~0.73–0.76 s): the single-threaded reader/parser of a 3.2 GB file becomes the
bottleneck, so matching is no longer limiting and the de-nesting gain shrinks into
noise. In production at high vCPU counts and/or with gzipped input (decompression on
the reader thread), expect the pipeline to be reader/IO-bound, where the matching
speedup mostly will not show up in wall-clock — there the value of this change is the
**crash fix**, not throughput. The throughput win appears when matching is the
bottleneck (moderate thread counts, larger references, higher match rates, or larger
`--minhits` that force more of each read to be scanned). The harness's broader trend
(new ≈ 0.13–0.14 s/M reads vs old ≈ 0.30–0.42 s/M) reflects the whole 1.4.2→1.5.0
line including big *index-build* speedups, not just matching.

The bottom line for the upgrade decision is unchanged: **no regression at any thread
count** (candidate ≥ old everywhere measured), and a real matching-stage win when
matching is the bottleneck.

Peak RSS was not separately measured in this pass; the memory *bound* is established
by construction (§2) — peak is flat in input size and capped at `permit_count ×
per-chunk bytes`.

---

## 5. Rust test suite

`cargo test --release` on the candidate: **26/26 integration tests pass**, 0 failures
(covers unpaired, paired-file, interleaved, all paired/interleaved output
permutations, canonical, minhits threshold, serialization round-trip, the metadata
sentinel, stdin, bloom basic/canonical/saveref, and arg-validation failures).

Note: `test_metadata_kmer_functional` **failed on `2208f36`** (a pre-existing test
bug: it checked only shard 0 for the sentinel, which with 1024 shards almost never
lands there, and its second case passed `--ref` so the rejected index silently
rebuilt). The follow-up fixes the *test*, not the product, and documents the
sentinel's collision-freedom for k ≤ 31. So the candidate is *greener* than the prior
main, not riskier.

---

## 6. Interface / invocation changes — none required

Every flag the pipeline uses today is unchanged in name, short form, and semantics:
`--in`, `--in2`, `--ref`, `--binref`, `--saveref`, `--outm`, `--outu`, `--outm2`,
`--outu2`, `--k`, `--minhits`, `--threads`, `--maxmem`, `--interinput`/`-i`,
`--order`/`-o`, `--canonical`/`-c`. **The existing production command line runs as-is.**

What changed, none of it breaking for the current invocation:

- **New optional flag `--fpr <float>`** (`main.rs:144-146`) enabling bloom mode.
  Default 0 = disabled = exact mode. Omit it and behavior is identical to prod.
- **`--in` now defaults to stdin** when omitted (`main.rs:100-102`, `core.rs:144-147`).
  The pipeline passes `--in` explicitly, so this is irrelevant to it; it only adds a
  capability.
- **`--ref` is no longer mandatory** *if* `--binref` is given; `validate_args`
  requires at least one of the two (`main.rs:216-228`). The pipeline passes `--ref`,
  so unaffected.
- **Version string** is now `1.5.0-alpha` (was `1.4.2`). If anything downstream
  parses `--version`, update that expectation.
- **Minor doc bug (not a blocker):** `--help` advertises a `--bloomcap 5M` option
  (`main.rs:69-71`) that is **not actually a parsed argument** (no field in `Args`;
  the validation for it is commented out at `main.rs:280-287`). Passing `--bloomcap`
  would be rejected by clap as unknown. Bloom-only and cosmetic; bloom sizing is
  derived automatically from reference size. Worth cleaning up but irrelevant to the
  exact-mode production path.

---

## 7. Residual risks and caveats (all pre-existing; none block the short-read upgrade)

These are carried over from `333912e`, not introduced by this change, and the
short-read production incident is unaffected:

1. **Long-read memory ceiling is bounded but can be large.** Chunks are still sized by
   **read count** (10,000 reads, `core.rs:678`), not bytes. For long, high-variance
   reads a single chunk's arena can be hundreds of MB, so peak ≈ `(num_threads+24) ×
   per-chunk-bytes` — potentially several GB on a high-vCPU box. This is strictly
   **better** than `333912e` (which was *unbounded* and grew with input), but it is
   not as tight as it should be. Byte-target / pair-aware chunking is the
   explicitly-deferred follow-up. The reported crashes were on the short-read path,
   where each chunk is ~5 MB and the bound is small (~tens of chunks × 5 MB), so this
   does not gate the upgrade — but watch long-read runs and prioritize byte-chunking
   if their memory headroom is tight.

2. **Permit count is a fixed `num_threads + 24`, not a memory-target dial.** Fine for
   short reads at any reasonable vCPU count (e.g. on a hypothetical 96-vCPU production
   box → ~120 × ~5.25 MB ≈ 0.6 GB).
   It is not configurable from the CLI; if a container is very memory-constrained
   relative to its vCPU count on long-read data, that is the lever to revisit.

3. **Fire-and-forget dispatch / swallowed send error.** The spawned chunk task still
   does `let _ = sender.send(chunk)` (`core.rs:718`) and tasks are not joined. If the
   writer exits early or a worker panics, a chunk could be dropped without surfacing
   an error. This is **unchanged from `333912e`** (no regression) and is a known
   follow-up cleanup. The reader thread *is* joined and its errors propagated
   (`core.rs:1061`).

4. **`stack_size(12 MB)` is retained** (`core.rs:127`), deliberately. De-nesting makes
   it unnecessary for matching, but lowering it needs an audit of the remaining flat
   Rayon sites (indexing, serialized-index load). Keeping it is harmless and the right
   call for a low-risk release.

---

## 8. Recommendation

**Deploy `4090fe3`.** It meets the stated bar:

- **Fixes the OOM** structurally (creation-to-write permit bounds every buffer,
  including the `--order` heap) and the **stack overflow** structurally (de-nested
  matching makes depth O(1)); the OOM is source-proven and the stack fix is both
  source-proven and reproduced.
- **No new bugs in the production path:** exact-mode output is byte-identical to
  `333912e` on 2M- and 10M-read prod-like datasets (both mates, `--order`), all 26
  Rust tests pass, bloom is inert without `--fpr`, and the previously-red metadata
  test is now green.
- **No performance regression:** ~1.33× faster on the heavy dataset when
  matching-bound (the de-nesting recovering the work-stealing cost), neutral when
  reader/IO-bound, never slower; flat, bounded memory. See §4 for the three-way
  attribution.
- **No invocation change:** the existing command line works unchanged.

Suggested follow-through (none blocking):

- Keep the production invocation exactly as today (do **not** add `--fpr`; exact mode
  is what is validated here).
- If convenient, capture one `RUST_BACKTRACE=1` dump from a *pre-upgrade* crash to
  formally close attribution — nice-to-have, not a gate.
- Track the follow-ups: byte-target + pair-aware chunking for long reads, a
  memory-target-derived permit count, joined workers with error propagation, and the
  `--bloomcap` help/flag cleanup.

PR #867, opened 2026-07-01
Fix NUCLEAZE deadlock when given an incompatible index

evanfields:
An incompatible --binref index makes nucleaze terminate during indexing, before it opens its input/output FIFOs. The background pigz helpers are then left blocked forever in open(), so the task never finishes — it stalls until killed by hand.

Capture nucleaze's exit status; on failure, kill the helpers before exiting so the task fails fast with nucleaze's error. The success path still waits on the helpers, preserving output gzip trailer flushing and decoder-error propagation.

Note that the Claude reviewer's D1 is intentional, for now: depending when this lands relative to the dev->main merge that introduces Nucleaze to main, we might not need a new CHANGELOG.md entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,7 +48,7 @@
     - INDEX: new `NUCLEAZE_INDEX` process builds `virus-genomes-masked.nucleaze.bin` once, alongside the existing bowtie2/minimap2 viral indexes.
     - RUN reads `nucleaze_k` from the index's `input/index-params.json` so the screen-time `k` always matches the index it screens against.
     - `NUCLEAZE` exposes `keep_match` / `keep_nomatch` flags (default true); the viral path passes `keep_nomatch: false` to skip compressing the discarded majority.
-    - `NUCLEAZE` pipes gzipped inputs through pigz FIFOs instead of letting needletail decompress them single-threaded internally; ~22 % wall reduction on the process (Illumina-100M samples, 8-core sandbox). Input-side pigz is capped at 2 threads since pigz cannot extract more parallelism from an ordinary single-stream gz. Module test covers the gz-input FIFO path.
+    - `NUCLEAZE` pipes gzipped inputs through pigz FIFOs instead of letting needletail decompress them single-threaded internally; ~22% wall reduction on the process (Illumina-100M samples, 8-core sandbox). Input-side pigz is capped at 2 threads since pigz cannot extract more parallelism from an ordinary single-stream gz. Module test covers the gz-input FIFO path.
     - Removes the unused `BBDUK_HITS_INTERLEAVE` process and its test.
     - Internal channel/param renames: `bbduk_match`/`bbduk_trimmed`/`min_kmer_hits`/`bbduk_suffix` → `kmer_match`/`kmer_trimmed`/`minhits`/`kmer_suffix`.
 - Add `triage-trivy` skill (`.claude/skills/triage-trivy/`) for per-CVE triage of `scan-containers` CI failures, structured to make `.trivyignore` the harder path. `CLAUDE.md`'s CI-status section points at it.
--- a/tests/modules/local/nucleaze/main.nf.test
+++ b/tests/modules/local/nucleaze/main.nf.test
@@ -272,12 +272,13 @@ nextflow_process {
         tag "expect_failure"
         tag "paired_end"
         tag "gz_input"
-        // Regression test for the FIFO deadlock: an incompatible --binref index
-        // makes nucleaze exit during indexing, before it opens the pigz FIFOs,
-        // so the pigz helpers block in open() forever unless the module tears
-        // them down. R1.fastq is passed as the "index" — it is not a serialized
-        // nucleaze index, so nucleaze fails during indexing and the process must
-        // fail promptly rather than stalling.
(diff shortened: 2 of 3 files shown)


evanfields (commented):

## Correctness

**No issues found.** The teardown logic in `modules/local/nucleaze/main.nf:82-103` captures both pipeline stages correctly via the `pipestatus=(0 0)` / `|| pipestatus=("${PIPESTATUS[@]}")` idiom (`PIPESTATUS` is still valid when the `||` right-hand side expands it), and all four nucleaze/tee status combinations resolve as intended under `pipefail`. On nucleaze failure it kills the FIFO-blocked pigz helpers — SIGTERM interrupts their blocking `open()`, so the `kill`/`wait` teardown cannot itself hang — and on success it drains all helpers first, preserving gzip-trailer flushing and decoder-error propagation. The failure branch always has a non-empty `PIDS` (at least one output FIFO is created since one of `keep_match`/`keep_nomatch` must be true), so the `set -u` array expansion is safe.

## Testing

- **T1** (**nit**): The new regression test (`tests/modules/local/nucleaze/main.nf.test:271`) asserts only `process.failed`. A missing fix manifests as a hang (caught only via the CI job timeout), and `process.failed` is also satisfied by any unrelated failure. Consider additionally asserting the failure is the indexing failure — e.g. on `process.exitStatus`, or that the captured stats/stderr references the bad index — so the test cannot silently pass for the wrong reason after a future change. The sibling gz-input success test (line 213) already isolates the gz path, so this is a robustness nit rather than a coverage gap.

## Documentation

- **D1** (**suggestion**): The only CHANGELOG.md change in this PR is a cosmetic `~22 %` → `~22%` edit to the pre-existing NUCLEAZE entry (line 51); nothing documents the deadlock fix itself. Because the NUCLEAZE feature is still unreleased under `# v3.2.2.0-dev`, folding this fix into it without a dedicated entry is defensible — but that intent isn't stated anywhere. Either add a sub-bullet under the NUCLEAZE changelog entry (e.g. "Fix a FIFO deadlock when `NUCLEAZE` is given an incompatible index") or note in the PR description that the fix folds into the unreleased NUCLEAZE feature, so reviewers understand why there is no dedicated changelog line. The inline comments in the module (lines 82-103) are otherwise thorough and correctly explain the *why*.

## Quality

**No issues found.** The implementation matches the existing script idioms (`set -euo pipefail`, the `PIDS` array, named-FIFO handling with `wait`-based error surfacing), and using a lowercase `pipestatus` variable to snapshot the builtin `PIPESTATUS` is a good choice. The multi-exit control flow is intricate but well-explained by the accompanying comments.

## Risks

**No issues found.** The single behavioral change — `NUCLEAZE` now fails fast with nucleaze's error instead of hanging until manually killed when given an incompatible index — is a strict improvement and is clearly described in the PR description. There are no interface, output-format, performance, resource, or dependency changes for a human reviewer to weigh.

tests/modules/local/nucleaze/main.nf.test
```diff
@@ -268,6 +268,57 @@ nextflow_process {
...
+                    minhits: "1",
+                    suffix: "viral"
+                ]
+            }
+            process {
+                '''
+                input[0] = GZIP_R1.out.map { s, f -> [s, f] }
+                    .join(GZIP_R2.out.map { s, f -> [s, f] })
+                    .map { s, f1, f2 -> [s, [f1, f2]] }
+                input[1] = "${projectDir}/test-data/tiny-index/reads/R1.fastq"
```
jo-delia-faraguna, line 314: **[clarification]** So you cause a failure by passing a read `.fastq` file in as the index? If so, maybe add a comment to make that clear.

CHANGELOG.md
```diff
@@ -48,7 +48,7 @@
     - INDEX: new `NUCLEAZE_INDEX` process builds `virus-genomes-masked.nucleaze.bin` once, alongside the existing bowtie2/minimap2 viral indexes.
     - RUN reads `nucleaze_k` from the index's `input/index-params.json` so the screen-time `k` always matches the index it screens against.
     - `NUCLEAZE` exposes `keep_match` / `keep_nomatch` flags (default true); the viral path passes `keep_nomatch: false` to skip compressing the discarded majority.
-    - `NUCLEAZE` pipes gzipped inputs through pigz FIFOs instead of letting needletail decompress them single-threaded internally; ~22 % wall reduction on the process (Illumina-100M samples, 8-core sandbox). Input-side pigz is capped at 2 threads since pigz cannot extract more parallelism from an ordinary single-stream gz. Module test covers the gz-input FIFO path.
+    - `NUCLEAZE` pipes gzipped inputs through pigz FIFOs instead of letting needletail decompress them single-threaded internally; ~22% wall reduction on the process (Illumina-100M samples, 8-core sandbox). Input-side pigz is capped at 2 threads since pigz cannot extract more parallelism from an ordinary single-stream gz. Module test covers the gz-input FIFO path.
```
jo-delia-faraguna, line 51: Don't think I don't know what you're doing here ;)

jo-delia-faraguna (approved):
LGTM, thanks for doing this.

tests/modules/local/nucleaze/main.nf.test
evanfields: Clarified and reworded the block comment above

PR #877, opened 2026-07-06
Disable Tower run-monitoring observer in test_run profile (fix benchmark CI hang)

dp-rice:
Standard-branch replacement for #876 (which was on a `release/…` branch). **Deferred: not part of 3.2.2.0** — since #874's CI went green, this is saved for after the dev→main merge.

**Root cause.** The benchmark/chain runs under `-profile test_run` with `TOWER_ACCESS_TOKEN` in the env (for Wave container auth + rate limits). That token alone auto-enables Nextflow's Seqera Platform (Tower) run-monitoring observer. On the pinned Nextflow **25.10.4**, `TowerClient.onFlowComplete()` does an HTTP PUT with **no read timeout**; on a stale connection it blocks the main thread forever, so the JVM never exits after the workflow completes and the benchmark job hangs to its 2h timeout. Upstream: **[nextflow-io/nextflow#6885](https://github.com/nextflow-io/nextflow/issues/6885)** (intermittent / timing-dependent; fixed in a later Nextflow).

**Fix.** `tower.enabled = false` in the `test_run` profile — the observer is never created, so `onFlowComplete()` can't hang. Wave still reads the token from the env independently.

**Scope.** `test_run` only (benchmarks / chained test). If the hang ever appears on a non-benchmark tokened run, the real fix is upgrading Nextflow past 25.10.4.

**Before merging (post-3.2.2.0):** rebase onto the reset `dev` and bump to the next `-dev` version (e.g. `3.2.2.1-dev`), moving the CHANGELOG entry under that heading. Until then this is a held draft and `check-version` will be red (non-release branch + non-`-dev` version) — expected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,7 +35,7 @@
 - Make `DOWNLOAD_VIRAL_GENOMES` retry the dehydrated `datasets download` step with exponential backoff, so a transient NCBI stream error there no longer aborts the task. (#828)
 - Fix the `Show version info` step in the manual-reset workflow, which failed on non-checked-out branch refs. (#805)
 - Add `set -euo pipefail` to the `BLASTN` module. (#838)
-- Disable the Seqera Platform (Tower) run-monitoring observer in the `test_run` profile (`tower.enabled = false`). It was auto-enabled by `TOWER_ACCESS_TOKEN` (present for Wave container auth), but on the pinned Nextflow `25.10.4` its `onFlowComplete()` PUT lacks an HTTP read timeout and intermittently hangs JVM exit after the workflow completes, timing out `benchmark-illumina-100M` CI ([nextflow-io/nextflow#6885](https://github.com/nextflow-io/nextflow/issues/6885)). Wave still authenticates from the environment token independently. (#PRNUM)
+- Disable the Seqera Platform (Tower) run-monitoring observer in the `test_run` profile (`tower.enabled = false`). It was auto-enabled by `TOWER_ACCESS_TOKEN` (present for Wave container auth), but on the pinned Nextflow `25.10.4` its `onFlowComplete()` PUT lacks an HTTP read timeout and intermittently hangs JVM exit after the workflow completes, timing out `benchmark-illumina-100M` CI ([nextflow-io/nextflow#6885](https://github.com/nextflow-io/nextflow/issues/6885)). Wave still authenticates from the environment token independently. (#876)
 
 ## Cleanup and best practice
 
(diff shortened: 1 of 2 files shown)


dp-rice:
⚠️ **Verification: this fix does not work as written.**

On-demand benchmark run [28813527879](https://github.com/securebio/nao-mgs-workflow/actions/runs/28813527879) on this branch's commit `c54ad5f4` (which contains `tower.enabled = false` in `test_run`):
- The run completed **success** in ~58 min with a clean exit — **but** the Tower observer was still active: the log prints `Monitor the execution with Seqera Platform` with two real watch URLs (`cloud.seqera.io/user/will1/watch/…`, one per RUN/DOWNSTREAM phase).
- So `tower.enabled = false` (set in the `test_run` profile) does **not** override the `TOWER_ACCESS_TOKEN` env-var auto-enable on Nextflow 25.10.4 — the observer ran anyway. The passing run was the flaky `onFlowComplete()` PUT happening to succeed, not the fix taking effect.

**Needs rework.** Options to investigate:
1. Whether `tower.enabled = false` set at **top level** (not inside a profile) is read early enough by `TowerFactory` — the profile-scoped value may be applied too late.
2. Provide the Wave token via `tower.accessToken` in config while leaving `TOWER_ACCESS_TOKEN` **unset** in the env (Wave may auth from config without the env var auto-enabling the run observer).
3. The robust fix: upgrade Nextflow past 25.10.4 (the [#6885](https://github.com/nextflow-io/nextflow/issues/6885) read-timeout fix), currently deferred via `.nextflowignore`.
4. Stopgap only: raise `timeout-minutes` (masks the flake, doesn't fix it).

Keeping as draft pending a working approach.

dp-rice:
Closing — this approach doesn't work. Verified via on-demand benchmark run [28813527879](https://github.com/securebio/nao-mgs-workflow/actions/runs/28813527879) on commit `c54ad5f4`: `tower.enabled = false` in the `test_run` profile does **not** disable the Seqera Platform observer on Nextflow 25.10.4 — the run still printed `Monitor the execution with Seqera Platform` with live watch URLs, so the `TOWER_ACCESS_TOKEN` env var auto-enables the observer regardless of the config flag. The run passed only because the flaky `onFlowComplete()` PUT happened to succeed, not because the hang was removed.

A working fix will need a different approach (top-level `tower.enabled=false` read earlier by `TowerFactory`, providing the Wave token via config instead of the env var, or the Nextflow upgrade that adds the read timeout — #6885). Leaving this to the sandbox-agent investigation.

Issue #873, opened 2026-07-01
NUCLEAZE: fail fast on binary/index version mismatch instead of hanging (follow-up to #867)


dp-rice:
## Summary

`NUCLEAZE` can still hang indefinitely (silent, zero output) when the `nucleaze` **binary version** does not match the version that built the index's `virus-genomes-masked.nucleaze.bin`. #867 fixed the *clean-error* case but not the *hang* case. RUN should detect a binary/index version mismatch and fail fast with a clear message.

## Background

#867 ("Fix NUCLEAZE deadlock when given an incompatible index") added a guard in `modules/local/nucleaze/main.nf`: run nucleaze, capture its exit status, and if non-zero, kill the blocked pigz FIFO helpers so the task fails fast instead of stalling.

That guard only fires **after nucleaze exits**:

```bash
nucleaze --in ... --binref ${index} --k ... 2>&1 | tee ${stats} || pipestatus=("${PIPESTATUS[@]}")
nucleaze_status=${pipestatus[0]}
if [[ "${nucleaze_status}" -ne 0 ]]; then
    kill "${PIDS[@]}" ...   # tear down blocked pigz helpers
    exit "${nucleaze_status}"
fi
```

## Observed failure (not covered by #867)

Running dev's pipeline with the **wrong** nucleaze binary (v1.4.2 from `rust-tools:main`) against an index also built by v1.4.2, driven by dev's v1.5.0-alpha-oriented module:

- nucleaze produced **zero** stdout/stderr (not even a usage error) and **never exited** — it hung during index loading.
- Because nucleaze never returned, the wrapper blocked forever on the invocation line and never reached the exit-status check, so #867's cleanup could not run.
- Net effect: 19 tasks "RUNNING" for ~2.5h with no progress, until killed by hand.

In short: #867 handles a nucleaze that *dies*; it does not handle a nucleaze that *hangs* on an interface/index it doesn't understand (the version-mismatch case).

## Proposed fix

Add a pre-flight version check in RUN's nucleaze path so a mismatch fails fast with a legible error rather than hanging. Options (any/combination):

1. Record the nucleaze version used to build the index (e.g. in `input/index-params.json`) and, at RUN time, compare it against the running `nucleaze --version`; error clearly if they differ (or are incompatible).
2. As defense-in-depth, wrap the nucleaze invocation with a timeout / watchdog so an indefinite stall becomes a failed task with a clear message instead of a silent multi-hour hang.

[...]

PR #881, opened 2026-07-09
Docs: warn that rust_tools_version defaults to :main for dev builds/runs

### LINKED ISSUE #872: Docs: warn that RUST_TOOLS_VERSION defaults to :main when building indexes / running against dev
## Summary

When building an index or running the pipeline against the **dev** branch *before it is merged to main*, `rust_tools_version` silently defaults to `"main"`, so the run uses main's (older) Rust tools — including an older `nucleaze` — instead of dev's. This is a silent footgun that is effectively undocumented in the places a user would look.

## What happened

While benchmarking the `dev` (3.2.2.0) pipeline + a new index against `main` before merge:

- The INDEX workflow was run from a correct `dev` checkout, but `RUST_TOOLS_VERSION` was not set, so `configs/containers.config`'s default (`System.getenv("RUST_TOOLS_VERSION") ?: "main"`) resolved to `:main`. The index's `virus-genomes-masked.nucleaze.bin` was therefore built with **nucleaze v1.4.2** (main's pin), not **v1.5.0-alpha** (dev's pin). `index-params.json` recorded `rust_tools_version: "main"`.
- The subsequent RUN (also defaulting to `:main`) then screened reads with v1.4.2 driven by dev's v1.5.0-alpha-oriented `NUCLEAZE` module → the task **hung silently for 2.5h** (0 bytes of output) until killed by hand.

## Why it wasn't catchable ahead of time

- `docs/installation.md` §7 (Run index/reference workflow) — the actual index-build instructions — has **no** mention of `rust_tools_version`. Same for `docs/usage.md`, `docs/run.md`, `docs/index.md`, `docs/troubleshooting.md`.
- The only mention is `docs/developer.md`, under *testing your own local Rust changes* (framed around `--rust_tools_version dev-$(whoami)`), so it doesn't obviously apply to "build an index / run the pipeline against dev."
- The `?: "main"` default produces no warning when a dev run silently uses main's tools.
- The orchestrator's pipeline↔index compatibility check is skipped for branch versions, so nothing flagged the mismatch.

## Proposed fix (docs)

1. In `docs/installation.md` §7 (index build) and `docs/usage.md` (running the pipeline), add a note: when building an index or running against **dev before it's merged to main**, set `RUST_TOOLS_VERSION=dev` (or pass `--rust_tools_version dev`), because the default `:main` container lags dev's Rust tools (e.g. nucleaze). Note both knobs (env var and `--rust_tools_version`) and that they must be set for **both** the index build and the run.
2. Cross-reference this from the nucleaze/CHANGELOG context and, if helpful, from the `benchmark-index` skill's review notes.

## Related

- The silent-hang failure mode is tracked separately (fail-fast on nucleaze binary vs index version mismatch); a clear runtime error is the stronger long-term guard, but the docs gap is worth closing regardless.

evanfields:
The Rust-tools container version resolves to `System.getenv("RUST_TOOLS_VERSION") ?: "main"`, so building an index or running the pipeline from an unmerged `dev` checkout silently uses main's *older* Rust tools rather than dev's. The only existing mention of the knob is in `docs/developer.md` (framed around local Rust development), so a user following the standard index-build / run instructions has no signal that they need to set it. This closes that docs gap; the complementary runtime fail-fast guard is tracked separately (#873 / COMP-2734).

**Changes:**
- `docs/installation.md` §7 (index build): add a `[!NOTE]` that the Rust-tools version defaults to `main`, and to pass `--rust_tools_version dev` (or `export RUST_TOOLS_VERSION=dev`) when building from an unmerged `dev` checkout — reusing the same version at run time.
- `docs/usage.md` §3 (running the pipeline): matching run-side `[!NOTE]`, noting the version must match the one used to build the index at `--ref_dir`.
- `.claude/skills/benchmark-index/SKILL.md`: add a Step 5 reviewer reminder to sanity-check `rust_tools_version` in `params_changes.tsv` — a `main` value on a dev-built index means it accidentally used main's older Rust tools.

Docs-only change: all three files are Markdown, so `check-changelog.yml`'s docs-only filter exempts it and no CHANGELOG entry / version bump is required (verified via pr-preflight).

Closes #872.

### DIFF
--- a/.claude/skills/benchmark-index/SKILL.md
+++ b/.claude/skills/benchmark-index/SKILL.md
@@ -131,6 +131,7 @@ Two reminders that are not obvious from the template alone:
 
 - Trust the script's annotations. If a row has `covered_by_hard_exclude = 2169574`, the script verified the lineage; do not manually re-classify. The "Categorization buckets" glossary below explains what each script-emitted label means.
 - Err on the side of inclusion in §Recommendations. Every plausible candidate change should appear, even at `low` confidence. In particular: any stale reference in §1, every uncovered promotion in §4, and every override policy gap in §4 should each appear as a candidate change.
+- Sanity-check `rust_tools_version` in `params_changes.tsv`. It selects the Rust-tools container used to build the index and defaults to `main`, so a value of `main` on an index built from a `dev` checkout means the index was built with main's (older) Rust tools by accident — flag it in §5, and note that RUN must later use a matching version.
 
 ### Step 6 - Hand off
 
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -110,6 +110,9 @@ nextflow run REPO_DIR \
 
 Replace `<BASE_DIR>` with the directory (likely on S3) where you want to store your index files, and `<BATCH_QUEUE_NAME>` with your AWS Batch job queue name. (You shouldn't need to modify anything else at this stage. However, if you'd like to, you can [learn more about what each parameter does here](./config.md).)
 
+> [!NOTE]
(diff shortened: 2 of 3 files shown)


dp-rice (approved):
Thanks!

PR #884, opened 2026-07-12
ci(rust-tools): publish stable-tagged container image

evanfields:
CI that builds and pushes the Rust-tools container to public ECR only ever produced `:main` and `:dev` tags — never `:stable`. The nao-mgs-orchestrator E2E matrix runs a `stable` leg that pulls `rust-tools:stable` (via `--rust_tools_version stable`); because that tag never existed, the `rust_tools`-labelled `PROCESS_VSEARCH_CLUSTER_OUTPUT` step couldn't start and the stable leg failed. This closes that gap.

**Changes:**
- `.github/workflows/rust-tools.yml`: add `stable` to the push-trigger branches, and derive the ECR tag from `github.ref_name` (via an `env:` var, per the repo's no-`${{ }}`-in-`run:` rule) instead of a hardcoded main/dev ternary. Because the trigger is restricted to `{dev, main, stable}`, `github.ref_name` is exactly the tag to publish. Build/scan/paths-filter jobs are unchanged.
- `docs/developer.md`: note that `stable` also triggers a rebuild and that images are published under a branch-matching tag.
- CHANGELOG entry under the existing `v3.2.2.1-dev` heading (point-level, CI-only; no pipeline/index/schema change).

Note: merging this does **not** retroactively create `rust-tools:stable` — the image is published only on the next qualifying push/reset to `stable`. The tag was seeded manually (built from `stable`'s own source, which differs from `main`'s: nucleaze v1.4.2 vs v1.5.0-alpha, and no pigz) to unblock the E2E stable leg now. The pipeline default in `configs/containers.config` remains `main`; only the orchestrator's stable leg sets `--rust_tools_version stable`.

### DIFF
--- a/.github/workflows/rust-tools.yml
+++ b/.github/workflows/rust-tools.yml
@@ -2,7 +2,7 @@ name: Rust Tools CI
 
 on:
   push:
-    branches: [dev, main]
+    branches: [dev, main, stable]
   pull_request:
 
 permissions:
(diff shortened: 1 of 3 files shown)


jo-delia-faraguna (approved):
LGTM. Is the default rust tools version of `main` [here](https://github.com/securebio/nao-mgs-workflow/blob/main/configs/containers.config#L10) still reasonable for pipeline executions?

evanfields:
> Is the default rust tools version of `main` [here](https://github.com/securebio/nao-mgs-workflow/blob/main/configs/containers.config#L10) still reasonable for pipeline executions?

I think so. Particular executions can pass the version they need, but (a) this is how it was before, so minimal change and (b) we should increasingly treat main as, well, the main branch.

